PR #7, opened 2023-02-23
Add Neuron exporter

JingyaHuang:
This PR add exporter for neuron/neuronx compiler to ease the tracing of models in Transformers.

- [x] Exporter PyTorch model -> Neuron compiled TorchScript model
- [x] Tests
- [x] Move utils

---

Snippet for a quick test:

```python
import copy
from pathlib import Path

from transformers import AutoModelForSequenceClassification

from optimum.neuron.exporter import export, validate_model_outputs
from optimum.neuron.exporter.model_configs import DistilBertNeuronConfig

model_id = "distilbert-base-uncased-finetuned-sst-2-english"
model = AutoModelForSequenceClassification.from_pretrained(model_id)
reference_model = copy.deepcopy(model)
neuron_config = DistilBertNeuronConfig(config=model.config, task="sequence-classification")
output_path = Path("model.pt")

input_shapes = {"batch_size": 2, "sequence_length": 18}
export(
    model=model,
    config=neuron_config,
    output=output_path,
    input_shapes=input_shapes,
    auto_cast="matmult",
    auto_cast_type="fp16",
)

# Validation
run = 10
for _ in range(run):
    validate_model_outputs(
        config=neuron_config,
        reference_model=reference_model,
        neuron_model_path=output_path,
        neuron_named_outputs=["logits"],
        input_shapes=input_shapes,
    )
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ EXTRAS_REQUIRE = {
     "tests": TESTS_REQUIRE,
     "quality": QUALITY_REQUIRES,
     "neuron": ["neuron-cc[tensorflow]", "torch-neuron", "protobuf==3.20.2", "torchvision"],
+    "neuronx": ["neuronx-cc==2.*", "torch-neuronx", "torchvision"],
 }
 
 setup(


JingyaHuang:
For `xlm` and `flaubert`, I observed higher than usual(`1e-5`) diff to PyTorch on CPU even without any casting(FP32):

* xlm
```
E   AssertionError: xlm, sequence-classification -> The maximum absolute difference between the output of the reference model and the Neuron exported model is not within the set tolerance 0.0001:
E   - logits: max diff = 0.09034809470176697
```
* flaubert
```
E   AssertionError: flaubert, sequence-classification -> The maximum absolute difference between the output of the reference model and the Neuron exported model is not within the set tolerance 0.0001:
E   - logits: max diff = 0.04413249343633652
```

JingyaHuang:
I remove the exporter test from the common CI as it needs to be run on CIs using AWS instance anyway. Will open a CI for neuron/neuronx exporters.

philschmid:
> I remove the exporter test from the common CI as it needs to be run on CIs using AWS instance anyway. Will open a CI for neuron/neuronx exporters.

We could add tests which just test if exports works. Since you can compile models on non Inferentia instances, but the results are off.

JingyaHuang:
hey @michaelbenayoun, can you re-review this PR as I separate neuron and neuronx cases and added some neuron specific args today.

optimum/neuron/__init__.py
```diff
@@ -12,12 +12,32 @@
...
-from .utils.training_utils import patch_transformers_for_neuron_sdk
+from transformers.utils import _LazyModule
 
 
-if not os.environ.get("DISABLE_TRANSFORMERS_PATCHING", False):
-    patch_transformers_for_neuron_sdk()
+_import_structure = {
+    "hf_argparser": ["TrainiumHfArgumentParser"],
+    "trainers": ["TrainiumTrainer", "Seq2SeqTrainiumTrainer"],
+    "utils": ["is_neuron_available", "is_neuronx_available", "patch_transformers_for_neuron_sdk"],
```
michaelbenayoun, line 23: Here we call `patch_transformers_for_neuron_sdk`, we dont import it.
But I think it does not do anything anyways.

optimum/neuron/exporter/convert.py
```diff
@@ -0,0 +1,279 @@
...
+            "The maximum absolute difference between the output of the reference model and the Neuron "
+            f"exported model is not within the set tolerance {atol}:\n{msg}"
+        )
+
+
+def export(
+    model: "PreTrainedModel",
+    config: "NeuronConfig",
+    output: Path,
+    input_shapes: Optional[Dict[str, Tuple[int, ...]]] = None,
```
michaelbenayoun, line 148: Why is this needed?
The shape should already be in the `config`.

optimum/neuron/exporter/convert.py
```diff
@@ -0,0 +1,279 @@
...
+        raise RuntimeError(
+            "Cannot export the model because the neuron(x) compiler is not installed. See https://awsdocs-neuron.readthedocs-hosted.com/en/latest/frameworks/torch/torch-setup.html."
+        )
+
+
+def export_neuronx(
+    model: "PreTrainedModel",
+    config: "NeuronConfig",
+    output: Path,
+    input_shapes: Optional[Dict[str, Tuple[int, ...]]] = None,
```
michaelbenayoun, line 165: Same comment

setup.py
```diff
@@ -14,7 +14,7 @@
 
 INSTALL_REQUIRES = [
     "transformers >= 4.26.0",
-    "optimum",
+    "optimum @ git+https://github.com/huggingface/optimum.git",
```
michaelbenayoun, line 17: Hmmmm, should we require that?

JingyaHuang: I will revert it. I previously put it this way as the CI will fail without our lazy import fix. I think that we shall install optimum from source for some CIs, but I also agree that we shall not require users to install optimum from source for using it, so maybe in TEST_REQUIRES but not INSTALL_REQUIRES.

optimum/neuron/exporter/convert.py
JingyaHuang: Yep, may bad. Will use the config.

optimum/neuron/__init__.py
JingyaHuang: Oh right. Put it outside the import code.

.github/workflows/test_common.yml
```diff
@@ -47,5 +47,5 @@ jobs:
       run: |
         # Setting HUGGINGFACE_CO_STAGING to true for only one job of the matrix as the staging tests cannot run in parallel.
         export HUGGINGFACE_CO_STAGING=${{ matrix.python-version == '3.8' && matrix.os == 'ubuntu-20.04' }}
-        pytest tests/test_*.py
+        pytest tests/test_examples.py tests/test_utils.py
```
michaelbenayoun, line 50: No I think we should test anything there, hence "test common".
Since export is a feature on its own, let's make a directory `tests/export` and put `test_export.py` there (and all the futur tests related to export there as well)

optimum/neuron/exporter/__main__.py
michaelbenayoun, line 1: Why?

optimum/neuron/utils/argument_utils.py
```diff
@@ -86,3 +86,28 @@ def expected_validation_function(arg):
                 "Tranium, set the environment variable OPTIMUM_DISABLE_STRICT_MODE to 1."
             )
             raise ValueError(f"{error_msg}\n{raise_error_msg}")
+
+
+def convert_neuronx_compiler_args_to_neuron(auto_cast="none", auto_cast_type=None, disable_fast_relayout=False):
+    """
+    Build `compiler_args` for neuron compiler.
```
michaelbenayoun, line 93:
```suggestion
    Builds `compiler_args` for neuron compiler.
```

optimum/neuron/utils/argument_utils.py
```diff
@@ -86,3 +86,28 @@ def expected_validation_function(arg):
                 "Tranium, set the environment variable OPTIMUM_DISABLE_STRICT_MODE to 1."
             )
             raise ValueError(f"{error_msg}\n{raise_error_msg}")
+
+
+def convert_neuronx_compiler_args_to_neuron(auto_cast="none", auto_cast_type=None, disable_fast_relayout=False):
```
michaelbenayoun, line 91: Let's add type annotations here because it can be ambiguous: `auto_cast` can be `None`, but also `"none"` for instance.

optimum/neuron/utils/argument_utils.py
```diff
@@ -86,3 +86,28 @@ def expected_validation_function(arg):
...
+    elif auto_cast in ["none", "all"]:
+        compiler_args.extend(["--fast-math", auto_cast])
+    elif auto_cast == "matmult":
+        if auto_cast_type is None:
+            compiler_args.extend(["--fast-math", "fp32-cast-matmult"])
+        else:
+            compiler_args.extend(["--fast-math", f"fp32-cast-matmult-{auto_cast_type}"])
+    else:
+        raise ValueError(
+            f"The auto_cast {auto_cast} is not valid. Please use one of the following: none, all or matmult."
```
michaelbenayoun, line 107:
```suggestion
            f"The auto_cast value {auto_cast} is not valid. Please use one of the following: none, all or matmult."
```

setup.py
```diff
@@ -35,6 +35,7 @@
 EXTRAS_REQUIRE = {
     "tests": TESTS_REQUIRE,
     "quality": QUALITY_REQUIRES,
+    "neuron": ["neuron-cc[tensorflow]", "torch-neuron", "protobuf==3.20.2", "torchvision"],
```
michaelbenayoun, line 38: Why not make it a base requirement?

JingyaHuang: Some might need installing neuronx for INF2 instead of neuron for INF1.

optimum/neuron/exporter/__main__.py
JingyaHuang: Place holder for exporter CLI.

optimum/neuron/utils/argument_utils.py
JingyaHuang: Added.

JingyaHuang:
@michaelbenayoun Can we merge?

michaelbenayoun (commented):
Left a question, otherwise LGTM!

optimum/neuron/utils/argument_utils.py
```diff
@@ -86,3 +86,32 @@ def expected_validation_function(arg):
                 "Tranium, set the environment variable OPTIMUM_DISABLE_STRICT_MODE to 1."
             )
             raise ValueError(f"{error_msg}\n{raise_error_msg}")
+
+
+def convert_neuronx_compiler_args_to_neuron(
+    auto_cast: Optional[str] = "none",
+    auto_cast_type: Optional[str] = None,
+    disable_fast_relayout: Optional[bool] = False,
```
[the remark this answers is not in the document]
JingyaHuang: Situations for these args equal to None are considered, I think it's okay. Any concern?

philschmid:
> It is aligned with optimum.exporters. We also need to find a good way there but it's a bit problematic. The tasks here are more related to the auto class names.

I do not see `optimum.exporters` used in the code. So why can't we do it the most user friendly way? And make it easy for users to understand. And not do it differently from what `transformers`, the `hub`, `tasks` pages, and the community does. 
The package is not publicly shared so we don't need to make similar mistakes. 
Our majority target user group are `transformers` users and not `optimum` user. Meaning we should keep everything as close as possible. And as a nonexpert, used to search for `text-classification` on the hub for models or read about it on hf.co/tasks. I would expect as "task" `text-classification` to export. 

also cc @jeffboudier

michaelbenayoun:
It will be part of `optimum.exporters.neuron` (will do in a PR soon if it can work). 
Also this heavily relies on the `TasksManager` which uses those names. 

We already thought of that before, the issue comes more from Transformers which defined many names for the same thing. 
The initial thing done was to mimic the auto class names (`AutoModelForSequenceClassification`) because that's what the tasks manager does: it allows to create auto classes from a string. 

We cannot change this here, and we discussed about that in root optimum already with no clear answer. An argument could be made that we do not export pipelines (except for stable diffusion) but models so matching the classes names make more sense.
In any case the task argument should not be needed for most models on the Hub as we can infer them.

philschmid:
> It will be part of optimum.exporters.neuron (will do in a PR soon if it can work).
> Also this heavily relies on the TasksManager which uses those names.

What do you mean by this? 

> We cannot change this here, and we discussed about that in root optimum already with no clear answer. An argument could be made that we do not export pipelines (except for stable diffusion) but models so matching the classes names make more sense.

I am not sure why cannot make the right decision and changes here? The package stands for itself? IMO our main priority should be to make it as natural and user-friendly as possible. This doesn't mean insisting on something which was done in the past. Especially, since there is a strong signal that the known and used term for this "task" as the attribute suggest `text-classification` is. 

> In any case the task argument should not be needed for most models on the Hub as we can infer them.

Thats good!

michaelbenayoun:
This export will use the `TasksManager`, just as every exporter,
As you can see, the mapping is defined here: https://github.com/huggingface/optimum/blob/main/optimum/exporters/tasks.py#L115

So we cannot change. I do not think this is a big deal, all our users seem to get along well with that for other exporters backend.

philschmid:
> So we cannot change. I do not think this is a big deal, all our users seem to get along well with that for other exporters backend.

Do we have data on this? Do we know how often the exporters are used by nonexpert users? 
I am really not a big fan of naming "tasks" different from what common sense is. That is bad IMO. 

for me, it makes absolutely makes no sense to have a workflow like this
```python
...
neuron_config = DistilBertNeuronConfig(config=model.config, task="sequence-classification")
.... 
export(...)
...
clf = pipeline("text-classification", ...)
...
^thats feels like bad UX to me. 

I know with `NeuronModelFor` classes we can use the `from_transformers` key word and handle the abstraction but still. The above snippet straight up feels wrong and we shouldn't not make bad design choices like that.

michaelbenayoun:
I do not have data on how many people use it, also what is a nonexpert user?

Your common sense is because you are more used to pipelines, someone using models would most likely call feel very okay with  "sequence-classification" since we have `BertForSequenceClassification`,  and basically all text classification models are called `XXXForSequenceClassification`.

Calling "bad design" simply a naming for which I gave you the rational behind is a bit strong, especially since we are adapting from Transformers which uses both, and on the level of exporters, sticking to the models (which exporters actually exports) seems more logical than sticking to pipelines.

PR #16, opened 2023-03-16
Move to exporters

michaelbenayoun:
Moves `optimum.neuron.exporter` to `optimum.exporters.neuron`.


JingyaHuang:
You shall add an `__init__.py` file under exporters folder, otherwise neuron won't be identified.

Despite that LGTM. Tested on INF2, it will work well once `__init__.py` is added.

JingyaHuang (approved):
LGTM! THX!

PR #25, opened 2023-03-22
Add CI for inference tests on INF1

JingyaHuang:
Add inferentia 1 CI for testing :
- [x] `tests/exporter/test_*`


.github/workflows/test_inf1.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: |
+          python3.7 -m venv aws_neuron_venv_pytorch 
+          source aws_neuron_venv_pytorch/bin/activate
+          python -m pip install -U pip
+          python -m pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
+          python -m pip install .[neuron,tests]
+      - name: Run tests
+        run: |
+          source aws_neuron_venv_pytorch/bin/activate
+          python -m pytest tests/exporters/test_*
```
michaelbenayoun, line 71: I think once the compilation cache PR is merged, this can become 
```suggestion
          pytest -m is_inferientia_test tests
```

tests/exporters/exporters_utils.py
```diff
@@ -32,14 +32,34 @@
...
+        ],
+    },
+    "roberta": {
+        "hf-internal-testing/tiny-random-RobertaModel": [
+            "masked-lm",
+            "sequence-classification",
+            "token-classification",
+            "question-answering",
+        ],
+    },
```
michaelbenayoun, line 58: Why?
If the other tasks are not supported, we should make sure we do not register them as suppoted!

JingyaHuang: They are supported, but their tests on the default task are flaky(they succeed when I launch them individually but sometimes fail when I launch all tests, I need to investigate further why they failed randomly). I will remove them from the registration for now until I figure out why.

.github/workflows/test_inf1.yml
JingyaHuang: Will add the marker.

tests/exporters/exporters_utils.py
michaelbenayoun: Hmmm, that's weird. Maybe some state is shared among them.

JingyaHuang: This will be corrected by #33 !

JingyaHuang: Nope, it's still there... Will do further check.

PR #52, opened 2023-04-29
Update setup for neuronx

JingyaHuang:


setup.py
```diff
@@ -44,7 +44,13 @@
         "protobuf",
         "torchvision",
     ],
-    "neuronx": ["neuronx-cc==2.*", "torch-neuronx", "torchvision"],
+    "neuronx": [        
+        "wheel",
+        "neuronx-cc==2.*", 
+        "torch-neuronx",
```
philschmid, line 50: not sure if thats a good idea since if you run pip install you cannot install them since they are not on `pypi`

JingyaHuang: Yeah, you are right. Better we set up aws's repo as an extra path to pip. I just added the `PIP_EXTRA_INDEX_URL` env variable in setup. WDYT?

philschmid: Does `PIP_EXTRA_INDEX_URL` work when publishing to pypi and then installing it?

JingyaHuang: Good question, I just tested with `pip install .[neuronx]`. I can upload to testpypi for a double check.

philschmid: lets test and make sure it uses the correct index. Since AWS published dummy packages which you can still install from the normal pypi

JingyaHuang: @philschmid Just tested, seems working decently:
image
You can find the test package here: https://test.pypi.org/project/optimum-neuron/

philschmid: Thats awesome!

setup.py
```diff
@@ -1,8 +1,10 @@
+import os
 import re
 
 from setuptools import find_namespace_packages, setup
 
 
+os.system("export PIP_EXTRA_INDEX_URL=https://pip.repos.neuron.amazonaws.com")
```
michaelbenayoun, line 7:
```suggestion
os.environ["PIP_EXTRA_INDEX_URL"] = "https://pip.repos.neuron.amazonaws.com"
```

michaelbenayoun: What happens if it was already set by someone?
Can we append?

If not the next best thing we can do is cache the previous value of the variable, set it to what we need, and restore it back to its original value.

JingyaHuang: I just found out that the setuptools have native support for extra index url. Tested with PyPI, works well: https://test.pypi.org/manage/project/optimum-neuron/release/0.0.4.dev1/

JingyaHuang:
Pinged numpy and protobuf version, test package here: https://test.pypi.org/manage/project/optimum-neuron/release/0.0.4.dev2/
Now with `pip install optimum-neuron[neuronx]==0.0.4`(after release of course), you have all set.

PR #55, opened 2023-05-02
[exporter] Save compile configurations

### LINKED ISSUE #53: Add batch size and sequence length compile options to config
When compiling a model with `optimum-cli export neuron` it would be useful to store the compiled model batch size and sequence length values so that the inference server can use this information when creating the tokenizer for inference requests

JingyaHuang:
fix #53 
* Save static shapes during compilation into config.
* Save compilation arguments during compilation into config.
* Enhancement on the export arguments.

### DIFF
--- a/optimum/commands/export/neuron.py
+++ b/optimum/commands/export/neuron.py
@@ -65,12 +65,12 @@ def parse_args_neuron(parser: "ArgumentParser"):
     )
     optional_group.add_argument(
         "--auto_cast_type",
-        default=None,
-        help='The data type to cast FP32 operations to when auto-cast mode is enabled. Can be `"bf16"`, `"fp16"` or `"tf32"`.',
+        default="bf16",
+        help='The data type to cast FP32 operations to when auto-cast mode is enabled. Can be `"bf16"`, `"fp16"`, `"mixed"` or `"tf32"`.',
     )
     optional_group.add_argument(
         "--disable_fast_relayout",
-        default=False,
+        action="store_true",
         help="Whether to disable fast relayout optimization which improves performance by using the matrix multiplier for tensor transpose.",
     )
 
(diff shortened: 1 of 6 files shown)


optimum/commands/export/neuron.py
```diff
@@ -65,8 +65,8 @@ def parse_args_neuron(parser: "ArgumentParser"):
     )
     optional_group.add_argument(
         "--auto_cast_type",
-        default=None,
-        help='The data type to cast FP32 operations to when auto-cast mode is enabled. Can be `"bf16"`, `"fp16"` or `"tf32"`.',
+        default="bf16",
```
michaelbenayoun, line 68: Maybe add a choices argument to make sure it is one of the 4 values.

optimum/commands/export/neuronx.py
```diff
@@ -65,7 +65,7 @@ def parse_args_neuronx(parser: "ArgumentParser"):
     )
     optional_group.add_argument(
         "--auto_cast_type",
-        default=None,
+        default="bf16",
```
michaelbenayoun, line 68: Same comment

optimum/commands/export/neuron.py
JingyaHuang: Done!

optimum/commands/export/neuron.py
```diff
@@ -60,12 +60,16 @@ def parse_args_neuron(parser: "ArgumentParser"):
     )
     optional_group.add_argument(
         "--auto_cast",
+        type=str,
         default=None,
+        choices=["None", "matmul", "all"],
```
michaelbenayoun, line 65:
```suggestion
        choices=["none", "matmul", "all"],
```

JingyaHuang: Corrected with changes in __main__ [here](https://github.com/huggingface/optimum-neuron/pull/55/commits/ff62fdef264728d1cddc339077e53863a0689035)

PR #65, opened 2023-05-15
[Exporter] Remove default values & Ping neuron minor version

JingyaHuang:
- [ ] Not using default shapes for export (mandatory axes are now necessary to the exporter)
- [ ] Ping minor versions for neuron SDK 2.10

### DIFF
--- a/optimum/commands/export/neuron.py
+++ b/optimum/commands/export/neuron.py
@@ -83,19 +83,16 @@ def parse_args_neuron(parser: "ArgumentParser"):
     input_group.add_argument(
         "--batch_size",
         type=int,
-        default=1,
         help=f"Batch size {doc_input}",
     )
     input_group.add_argument(
         "--sequence_length",
         type=int,
-        default=16,
         help=f"Sequence length {doc_input}",
     )
     input_group.add_argument(
         "--num_choices",
         type=int,
-        default=4,
         help=f"Only for the multiple-choice task. Num choices {doc_input}",
     )
 
--- a/optimum/commands/export/neuronx.py
+++ b/optimum/commands/export/neuronx.py
@@ -78,19 +78,16 @@ def parse_args_neuronx(parser: "ArgumentParser"):
     input_group.add_argument(
         "--batch_size",
         type=int,
-        default=1,
         help=f"Batch size {doc_input}",
     )
     input_group.add_argument(
         "--sequence_length",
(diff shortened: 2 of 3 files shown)


optimum/commands/export/neuron.py
```diff
@@ -88,19 +88,16 @@ def parse_args_neuron(parser: "ArgumentParser"):
     input_group.add_argument(
         "--batch_size",
         type=int,
-        default=1,
```
michaelbenayoun, line 91: I think, `batch_size` might be the only shape we accept a default for, since it's for inference. But do not have a strong opinion on this.

optimum/commands/export/neuronx.py
```diff
@@ -83,19 +83,16 @@ def parse_args_neuronx(parser: "ArgumentParser"):
     input_group.add_argument(
         "--batch_size",
         type=int,
-        default=1,
```
michaelbenayoun, line 86: Same comment

optimum/exporters/neuron/base.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         self,
         config: "PretrainedConfig",
         task: str,
-        batch_size: int = 1,
+        batch_size: int = None,
```
michaelbenayoun, line 101:
```suggestion
        batch_size: Optional[int] = None,
```

optimum/commands/export/neuron.py
philschmid: Sounds good! Also compared to sequence length this should error when you are not providing the right batch size

philschmid (approved):

michaelbenayoun (approved):

docs/source/package_reference/configuration.mdx
```diff
@@ -33,3 +33,34 @@ Since many architectures share similar properties for their Neuron configuration
...
    They specify which input generators should be used for the dummy inputs, but remain model-agnostic.
 3. Model-specific classes like the `BertNeuronConfig` mentioned above. These are the ones actually used to export models.
+
+
+## Supported architectures
+
+
+| Architecture           | Task                                 |
+|------------------------|--------------------------------------|
+| ALBERT                 | All                                  |
```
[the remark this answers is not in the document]
JingyaHuang: Ok, will do!

PR #89, opened 2023-06-02
Deactivate test for docs

philschmid:


.github/workflows/test_on_trainium.yml
```diff
@@ -2,11 +2,13 @@ name: Optimum Neuron - Tranium-dependent Tests
...
-      - main
+    branches: [ main ]
+    paths:
+      - "optimum/**.py"
   pull_request:
-    branches:
-      - main
+    branches: [ main ]
+    paths:
+      - "optimum/**.py"
```
michaelbenayoun, line 11: Would it trigger the test if we change the workflow file?
Can we do like a filter saying "not docs" instead of saying what should trigger the CI?

philschmid: no but we can include the path as well

PR #104, opened 2023-06-21
Tensor Parallelism training with `neuronx_distributed`

michaelbenayoun:
As per title.

### DIFF
--- a/optimum/neuron/accelerate/optimizer.py
+++ b/optimum/neuron/accelerate/optimizer.py
@@ -28,7 +28,7 @@ if is_torch_xla_available():
     accelerate.optimizer.xm = xm
 
 if is_neuronx_distributed_available():
-    from neuronx_distributed import parallel_state
+    from neuronx_distributed.parallel_layers import parallel_state
 
 
 class NeuronAcceleratedOptimizer(AcceleratedOptimizer):
--- a/optimum/neuron/accelerate/optimizer.py
+++ b/optimum/neuron/accelerate/optimizer.py
@@ -45,6 +45,10 @@ class NeuronAcceleratedOptimizer(AcceleratedOptimizer):
                 self.optimizer.step(closure)
             elif self.accelerator_state.distributed_type is NeuronDistributedType.TENSOR_PARALLELISM:
                 xm.optimizer_step(self.optimizer, groups=parallel_state.get_data_parallel_group(as_list=True))
+                # xm.reduce_gradients(self.optimizer, groups=parallel_state.get_data_parallel_group(as_list=True))
+                # parameters = [p for group in self.optimizer.param_groups for p in group["params"]]
+                # parallel_layers.clip_grad_norm(parameters, 1.0, norm_type=2)
+                # self.optimizer.step()
             elif self.scaler is not None:
                 scale_before = self.scaler.get_scale()
                 self.scaler.step(self.optimizer, closure)
--- a/optimum/neuron/accelerate/optimizer.py
+++ b/optimum/neuron/accelerate/optimizer.py
@@ -17,7 +17,7 @@
 from accelerate.optimizer import AcceleratedOptimizer
 from accelerate.utils import DistributedType
 
-from ..utils import is_torch_xla_available
+from ..utils import is_neuronx_distributed_available, is_torch_xla_available
 from .utils.dataclasses import NeuronDistributedType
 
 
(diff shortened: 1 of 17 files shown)


optimum/neuron/accelerate/optimizer.py
```diff
@@ -40,6 +43,12 @@ def step(self, closure=None):
                 xm.optimizer_step(self.optimizer, optimizer_args=optimizer_args)
             elif self.accelerator_state.distributed_type is NeuronDistributedType.XLA_FSDP:
                 self.optimizer.step(closure)
+            elif self.accelerator_state.distributed_type is NeuronDistributedType.TENSOR_PARALLELISM:
+                xm.optimizer_step(self.optimizer, groups=parallel_state.get_data_parallel_group(as_list=True))
+                # xm.reduce_gradients(self.optimizer, groups=parallel_state.get_data_parallel_group(as_list=True))
+                # parameters = [p for group in self.optimizer.param_groups for p in group["params"]]
+                # parallel_layers.clip_grad_norm(parameters, 1.0, norm_type=2)
+                # self.optimizer.step()
```
JingyaHuang, line 51: Do you intend to leave those comments?

optimum/neuron/distributed/base.py
```diff
@@ -78,7 +83,9 @@ def saved_model_in_temporary_directory(cls, model: "PreTrainedModel"):
             tmpdir.cleanup()
 
     @abstractclassmethod
-    def parallelize(cls, model: "PreTrainedModel") -> "PreTrainedModel":
+    def parallelize(
+        cls, model: "PreTrainedModel", return_original_to_transformed_mapping: bool = False
+    ) -> Union["PreTrainedModel", Union["PreTrainedModel", Dict[int, "torch.nn.Parameter"]]]:
```
JingyaHuang, line 88:
```suggestion
    ) -> Union["PreTrainedModel", Dict[int, "torch.nn.Parameter"]]:
```
?

optimum/neuron/utils/cache_utils.py
```diff
@@ -208,6 +211,12 @@ def get_neuron_cache_path() -> Optional[Path]:
         else:
             path = Path("/var/tmp")
 
+        # TODO: is that correct?
+        # neuronxcc_version = get_neuronxcc_version()
+        # if version.parse(neuronxcc_version) < version.parse(_NEW_CACHE_NAMING_CONVENTION_NEURONXCC_VERSION):
```
JingyaHuang, line 216: If neuronxcc_version < compiled version, it's not going to work. The inversed is compatible according to Annapurna but I seems encountered issue with backward compatibility either.

JingyaHuang (approved):
That's huge, congrats on the PR, and thanks for working on it.

Left some minor nits, feel free to merge it whenever you like to.

optimum/neuron/accelerate/optimizer.py
michaelbenayoun: Not yet, I will later but for now it might come in handy.

optimum/neuron/distributed/base.py
michaelbenayoun: It was an error thanks!

optimum/neuron/accelerate/optimizer.py
michaelbenayoun: I tried with it and it still failed. Also we do not do any gradient accumulation in the tests so I am not sure?

optimum/neuron/distributed/utils.py
```diff
@@ -23,19 +23,58 @@
...
+        embedding_layer.embedding_dim,
+        dtype=embedding_layer.weight.dtype,
+    )
+
+    tp_rank = get_tensor_model_parallel_rank()
+    row_size, _ = parallel_embedding_layer.weight.shape
+
+    is_tied = False
+    if lm_head_layer is not None:
+        is_tied = id(embedding_layer.weight.data) == id(lm_head_layer.weight.data)
```
aws-rhsoln, line 47: If this is xla_tensor, then the data pointer may not be same. This would mean the weight has to be a CPU tensor first

optimum/neuron/accelerate/optimizer.py
aws-rhsoln: grad clipping is unrelated to grad_accumulation. What is the grad_clip value set by default for the non-tp case?

PR #117, opened 2023-06-28
Neuron model for causal lm partial support

dacorvo:
This is a very preliminary integration of a `NeuronModelForCausalLM` generic class to wrap neuronx optimized models.

For now, only `gpt2` model type is supported.

The following features are available:

- instantiate (export + compile) a neuron model from a `transformers` model_id  (`from_pretrained`),
- save the exported model locally (`save_pretrained`),
- instantiate (export + reload compilation artifacts) a neuron model from a local directory (`from_pretrained`),
- generate outputs (`generate`).

What is missing:
- between two generations, a specific `reset_generation` method must be called: I plan to remove that in a subsequent pull-request,
- we should check the compiler version when instantiating from a local path,
- the export should be available through the CLI,
- the new model class should have a corresponding `text-generation` pipeline.

What could be added if this is a use case:
- use cache (is it even possible since the compilation happens behind the scene in transformers-neuronx ?),
- support recompilation with different parameters when instantiating from a local path,

### DIFF
--- a/optimum/exporters/neuron/config.py
+++ b/optimum/exporters/neuron/config.py
@@ -16,17 +16,14 @@
 Common Neuron configuration classes that handle most of the features for building model specific
 configurations.
 """
-import importlib
 
-from ...exporters.base import ExportConfig
-from ...neuron.utils import is_transformers_neuronx_available
 from ...utils import (
     DummyBboxInputGenerator,
     DummyTextInputGenerator,
     DummyVisionInputGenerator,
     logging,
 )
-from .base import NeuronConfig
+from .base import NeuronConfig, NeuronDecoderConfig
 
 
 logger = logging.get_logger(__name__)
@@ -58,37 +55,9 @@ class TextAndVisionNeuronConfig(NeuronConfig):
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyVisionInputGenerator, DummyBboxInputGenerator)
 
 
-class DecoderNeuronConfig(ExportConfig):
+class TextNeuronDecoderConfig(NeuronDecoderConfig):
     """
-    Handles decoder architectures.
+    Handles text decoder architectures.
     """
 
-    NEURONX_ARGS = {}
-    NEURONX_MODULE = None
-    NEURONX_CLASS = None
-
-    def __init__(self, task):
-        if not is_transformers_neuronx_available():
-            raise ModuleNotFoundError("The transformers-neuronx package is required.")
-        module = importlib.import_module(self.NEURONX_MODULE)
-        self._neuronx_class = getattr(module, self.NEURONX_CLASS, None)
-        if self._neuronx_class is None:
-            raise ImportError(f"{self.NEURONX_CLASS} not found in {self.NEURONX_MODULE}. Please check versions.")
-
-    def split_kwargs(self, **kwargs):
-        """Split between kwargs that need to be passed when loading the transformers model
-        and those that need to be passed to the neuron optimizer.
-        """
-        model_kwargs = kwargs
-        neuron_kwargs = {}
-        for arg, default in self.NEURONX_ARGS.items():
-            if arg in model_kwargs:
-                neuron_kwargs[arg] = model_kwargs[arg]
-                model_kwargs.pop(arg)
-            else:
-                neuron_kwargs[arg] = default
-        return model_kwargs, neuron_kwargs
-
-    @property
-    def neuronx_class(self):
-        return self._neuronx_class
+    pass
--- a/optimum/exporters/neuron/config.py
+++ b/optimum/exporters/neuron/config.py
@@ -16,7 +16,10 @@
 Common Neuron configuration classes that handle most of the features for building model specific
 configurations.
 """
+import importlib
 
+from ...exporters.base import ExportConfig
+from ...neuron.utils import is_transformers_neuronx_available
 from ...utils import (
     DummyBboxInputGenerator,
     DummyTextInputGenerator,
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
     """
 
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyVisionInputGenerator, DummyBboxInputGenerator)
+
+
+class DecoderNeuronConfig(ExportConfig):
+    """
+    Handles decoder architectures.
+    """
+
+    NEURONX_ARGS = {"batch_size": 1, "n_positions": 128, "tp_degree": 2, "amp": "f32"}
+    NEURONX_MODULE = None
+    NEURONX_CLASS = None
+
+    def __init__(self, task, **kwargs):
+        if not is_transformers_neuronx_available():
+            raise ModuleNotFoundError("The transformers-neuronx package is required.")
+        module = importlib.import_module(self.NEURONX_MODULE)
+        self._neuronx_class = getattr(module, self.NEURONX_CLASS, None)
+        if self._neuronx_class is None:
+            raise ImportError(f"{self.NEURONX_CLASS} not found in {self.NEURONX_MODULE}. Please check versions.")
+
+    def split_kwargs(self, **kwargs):
+        """Split between kwargs that need to be passed when loading the transformers model
+        and those that need to be passed to the neuron optimizer.
+        """
+        model_kwargs = kwargs
+        neuron_kwargs = {}
+        for arg, default in self.NEURONX_ARGS.items():
+            if arg in model_kwargs:
+                neuron_kwargs[arg] = model_kwargs[arg]
+                model_kwargs.pop(arg)
+            else:
+                neuron_kwargs[arg] = default
+        return model_kwargs, neuron_kwargs
+
+    @property
+    def neuronx_class(self):
+        return self._neuronx_class
--- a/optimum/exporters/neuron/config.py
+++ b/optimum/exporters/neuron/config.py
@@ -67,7 +67,7 @@ class DecoderNeuronConfig(ExportConfig):
     NEURONX_MODULE = None
     NEURONX_CLASS = None
 
-    def __init__(self, task, **kwargs):
+    def __init__(self, task):
         if not is_transformers_neuronx_available():
             raise ModuleNotFoundError("The transformers-neuronx package is required.")
         module = importlib.import_module(self.NEURONX_MODULE)
--- a/optimum/exporters/neuron/config.py
+++ b/optimum/exporters/neuron/config.py
@@ -63,7 +63,7 @@ class DecoderNeuronConfig(ExportConfig):
     Handles decoder architectures.
     """
 
-    NEURONX_ARGS = {"batch_size": 1, "n_positions": 128, "tp_degree": 2, "amp": "f32"}
+    NEURONX_ARGS = {}
     NEURONX_MODULE = None
     NEURONX_CLASS = None
 
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -557,14 +557,8 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
     def __init__(self, model, config, model_path, generation_config):
         super().__init__(model, config, model_path, generation_config)
         self.cur_len = 0
-        neuron_config = getattr(config, "neuron", None)
-        if neuron_config is None:
-            raise ValueError(
-                "The specified directory does not contain a neuron model. "
-                "Please convert your model to neuron format by passing export=True."
-            )
-        self.batch_size = neuron_config["neuron_kwargs"]["batch_size"]
-        self.max_length = neuron_config["neuron_kwargs"]["n_positions"]
+        self.batch_size = model.config.batch_size
+        self.max_length = model.config.n_positions
 
     def reset_generation(self):
         self.cur_len = 0
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -594,7 +594,7 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
         output_hidden_states: bool = False,
         output_attentions: bool = False,
         attention_mask: torch.Tensor = None,
-        return_dict: bool = False,
+        return_dict: bool = True,
     ):
         if output_hidden_states or output_attentions or attention_mask is not None:
             warnings.warn(
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -557,6 +557,14 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
     def __init__(self, model, config, model_path, generation_config):
         super().__init__(model, config, model_path, generation_config)
         self.cur_len = 0
+        neuron_config = getattr(config, "neuron", None)
+        if neuron_config is None:
+            raise ValueError(
+                "The specified directory does not contain a neuron model. "
+                "Please convert your model to neuron format by passing export=True."
+            )
+        self.batch_size = neuron_config["neuron_kwargs"]["batch_size"]
+        self.max_length = neuron_config["neuron_kwargs"]["n_positions"]
 
     def reset_generation(self):
         self.cur_len = 0
@@ -593,9 +601,18 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
         return (out_logits,)
 
     def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
-        # Sanity check
+        # Sanity checks
         if past_key_values is not None:
             raise ValueError("This model does not support dynamic key, value cache.")
+        batch_size, sequence_length = input_ids.shape
+        if batch_size != self.batch_size:
+            raise ValueError(
+                f"The specified batch_size ({batch_size}) does not match the model static batch size ({self.batch_size})"
+            )
+        if sequence_length > self.max_length:
+            raise ValueError(
+                f"The current sequence length ({sequence_length}) exceeds the model static sequence length ({self.max_length})"
+            )
         # convert attention_mask to start_ids
         attention_mask = None
         start_ids = None
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -16,7 +16,7 @@
 
 import logging
 import warnings
-from typing import Optional
+from typing import TYPE_CHECKING, Dict, Optional, Union
 
 import torch
 from transformers import (
@@ -42,6 +42,13 @@ from transformers.utils import ModelOutput
 
 from .modeling_base import NeuronBaseModel
 from .modeling_decoder import NeuronDecoderModel
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
+    from tempfile import TemporaryDirectory
+
+    from transformers import GenerationConfig, PretrainedConfig
 
 
 logger = logging.getLogger(__name__)
@@ -554,11 +561,19 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
     auto_model_class = AutoModelForCausalLM
     main_input_name = "input_ids"
 
-    def __init__(self, model, config, model_path, generation_config):
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        config: "PretrainedConfig",
+        model_path: Union[str, "Path", "TemporaryDirectory"],
+        generation_config: Optional["GenerationConfig"] = None,
+    ):
         super().__init__(model, config, model_path, generation_config)
         self.cur_len = 0
         self.batch_size = model.config.batch_size
         self.max_length = model.config.n_positions
+        # The generate method from GenerationMixin expects the device attribute to be set
+        self.device = torch.device("cpu")
 
     def reset_generation(self):
         self.cur_len = 0
@@ -573,13 +588,13 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
     )
     def forward(
         self,
-        input_ids,
-        cache_ids,
-        start_ids=None,
-        output_hidden_states=False,
-        output_attentions=False,
-        attention_mask=None,
-        return_dict=False,
+        input_ids: torch.Tensor,
+        cache_ids: torch.Tensor,
+        start_ids: torch.Tensor = None,
+        output_hidden_states: bool = False,
+        output_attentions: bool = False,
+        attention_mask: torch.Tensor = None,
+        return_dict: bool = False,
     ):
         if output_hidden_states or output_attentions or attention_mask is not None:
             warnings.warn(
@@ -594,9 +609,9 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
             return ModelOutput([("logits", out_logits)])
         return (out_logits,)
 
-    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, **kwargs):
+    def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Dict[str, torch.Tensor]:
         # Sanity checks
-        if past_key_values is not None:
+        if kwargs.get("past_key_values", None) is not None:
             raise ValueError("This model does not support dynamic key, value cache.")
         batch_size, sequence_length = input_ids.shape
         if batch_size != self.batch_size:
@@ -635,6 +650,6 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
 
         return model_inputs
 
-    def can_generate(self):
-        """Returns True to validate the check that the model using `GenerationMixin.generate()` can indeed generate."""
+    def can_generate(self) -> bool:
+        """Returns True to validate the check made in `GenerationMixin.generate()`."""
         return True
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -15,11 +15,13 @@
 """NeuronModelForXXX classes for inference on neuron devices using the same API as Transformers."""
 
 import logging
+import warnings
 from typing import Optional
 
 import torch
 from transformers import (
     AutoModel,
+    AutoModelForCausalLM,
     AutoModelForMaskedLM,
     AutoModelForMultipleChoice,
     AutoModelForQuestionAnswering,
@@ -27,6 +29,7 @@ from transformers import (
     AutoModelForTokenClassification,
 )
 from transformers.file_utils import add_start_docstrings, add_start_docstrings_to_model_forward
+from transformers.generation import GenerationMixin
 from transformers.modeling_outputs import (
     BaseModelOutputWithPooling,
     MaskedLMOutput,
@@ -35,8 +38,10 @@ from transformers.modeling_outputs import (
     SequenceClassifierOutput,
     TokenClassifierOutput,
 )
+from transformers.utils import ModelOutput
 
 from .modeling_base import NeuronBaseModel
+from .modeling_decoder import NeuronDecoderModel
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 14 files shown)


optimum/neuron/modeling_decoder.py
```diff
@@ -114,31 +121,80 @@ def _from_transformers(
...
+            "neuron_kwargs": neuron_kwargs
+        }
+
+        return cls._from_pretrained(checkpoint_dir, config)
+
+    @classmethod
+    def _get_neuron_paths(cls, model_dir):
+        if isinstance(model_dir, TemporaryDirectory):
+            # Convert the model directory to a path
+            model_path = model_dir.name
```
michaelbenayoun, line 139:
```suggestion
            model_path = model_dir.name
```

optimum/neuron/modeling_decoder.py
```diff
@@ -114,31 +121,80 @@ def _from_transformers(
...
+        # Update the config
+        config.neuron = {
+            "task": task,
+            "neuron_kwargs": neuron_kwargs
+        }
+
+        return cls._from_pretrained(checkpoint_dir, config)
+
+    @classmethod
+    def _get_neuron_paths(cls, model_dir):
```
michaelbenayoun, line 136:
```suggestion
    def _get_neuron_paths(cls, model_dir: Union[str, Path, TemporaryDirectory]) -> Tuple[str, str, str]:
```

optimum/neuron/modeling_decoder.py
```diff
@@ -114,31 +121,80 @@ def _from_transformers(
...
+    ) -> "NeuronModelDecoder":
+        # Verify we are actually trying to load a neuron model
+        neuron_config = getattr(config, "neuron", None)
+        if neuron_config is None:
+            raise ValueError("The specified directory does not contain a neuron model. "
+                             "Please convert your model to neuron format by passing export=True.")
 
-        # Instantiate the optimized neuronx model from the transformers checkpoint
-        neuronx_model = exporter.neuronx_class.from_pretrained(save_dir_path, **neuron_kwargs)
+        # Evaluate the configuration passed during export
```
michaelbenayoun, line 165:
```suggestion
```

optimum/neuron/modeling_decoder.py
```diff
@@ -114,31 +121,80 @@ def _from_transformers(
...
+        neuron_kwargs = neuron_config["neuron_kwargs"]
 
-        # Compile the model
+        # Instantiate the exporter
+        exporter = get_exporter(config, task)
+
+        # Evaluate the path to the checkpoint and compiled directories
+        model_path, checkpoint_path, compiled_path = cls._get_neuron_paths(model_id)
+
+        # Instantiate the optimized neuronx model from the original model checkpoint
```
michaelbenayoun, line 175:
```suggestion
```

optimum/neuron/modeling_decoder.py
```diff
@@ -148,16 +204,37 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
             save_directory (`Union[str, Path]`):
                 Directory where to save the model file.
         """
-        raise NotImplementedError()
+        # Evaluate the path to the source checkpoint and compiled directories
```
michaelbenayoun, line 207:
```suggestion
```

optimum/neuron/modeling_decoder.py
```diff
@@ -148,16 +204,37 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
...
+        _, src_chkpt_path, src_compiled_path = self._get_neuron_paths(self.model_path)
+
+        # Evaluate the path to the destination checkpoint and compiled directories
```
michaelbenayoun, line 210:
```suggestion
```

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
     """
 
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyVisionInputGenerator, DummyBboxInputGenerator)
+
+
+class DecoderNeuronConfig(ExportConfig):
```
michaelbenayoun, line 61: It should inherint from `NeuronConfig` imo.

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+    """
+    Handles decoder architectures.
+    """
+
+    NEURONX_ARGS = {"batch_size": 1, "n_positions": 128, "tp_degree": 2, "amp": "f32"}
+    NEURONX_MODULE = None
+    NEURONX_CLASS = None
+
+    def __init__(self, task, **kwargs):
```
michaelbenayoun, line 70:
```suggestion
    def __init__(self, task: str):
```

Why allow for keyword arguments here?

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+        if not is_transformers_neuronx_available():
+            raise ModuleNotFoundError("The transformers-neuronx package is required.")
```
michaelbenayoun, line 72:
```suggestion
            raise ModuleNotFoundError("The transformers-neuronx package is required. You can install it by running the following command: pip install transformers-neuronx")
```

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+        module = importlib.import_module(self.NEURONX_MODULE)
+        self._neuronx_class = getattr(module, self.NEURONX_CLASS, None)
+        if self._neuronx_class is None:
+            raise ImportError(f"{self.NEURONX_CLASS} not found in {self.NEURONX_MODULE}. Please check versions.")
+
+    def split_kwargs(self, **kwargs):
+        """Split between kwargs that need to be passed when loading the transformers model
+        and those that need to be passed to the neuron optimizer.
```
michaelbenayoun, line 80:
```suggestion
        """
        Splits between kwargs that need to be passed when loading the transformers model
        and those that need to be passed to the neuron optimizer.
```

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
 
     DUMMY_INPUT_GENERATOR_CLASSES = (DummyTextInputGenerator, DummyVisionInputGenerator, DummyBboxInputGenerator)
+
+
+class DecoderNeuronConfig(ExportConfig):
+    """
+    Handles decoder architectures.
+    """
+
+    NEURONX_ARGS = {"batch_size": 1, "n_positions": 128, "tp_degree": 2, "amp": "f32"}
```
michaelbenayoun, line 66: Why are those values hardcoded here?

optimum/exporters/neuron/config.py
JingyaHuang, line 61: Would it make sense to inherit from `NeuronConfig` as I might have put / will put some type check of configs.

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+    NEURONX_MODULE = None
+    NEURONX_CLASS = None
```
michaelbenayoun, line 68: Maybe we can refactor both like:

```python
NEURONX_CLASS`: str = "gpt2.model.GPT2ForSampling"
```

I do not think that adding the `transformers_neuronx` here is neeeded since it's implied and we can add it once when we load the actual class.

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+        module = importlib.import_module(self.NEURONX_MODULE)
+        self._neuronx_class = getattr(module, self.NEURONX_CLASS, None)
+        if self._neuronx_class is None:
+            raise ImportError(f"{self.NEURONX_CLASS} not found in {self.NEURONX_MODULE}. Please check versions.")
+
+    def split_kwargs(self, **kwargs):
+        """Split between kwargs that need to be passed when loading the transformers model
+        and those that need to be passed to the neuron optimizer.
+        """
+        model_kwargs = kwargs
```
michaelbenayoun, line 82:
```suggestion
        model_kwargs = copy.deepcopy(kwargs)
```

optimum/neuron/modeling.py
```diff
@@ -506,3 +511,136 @@ def forward(
...
+        self.cur_len += input_ids.shape[-1]
+        model_inputs = {
+            "input_ids": input_ids,
+            "cache_ids": cache_ids,
+            "start_ids": start_ids,
+        }
+
+        return model_inputs
+
+    def can_generate(self):
```
michaelbenayoun, line 644: property or attribute?
```suggestion
    def can_generate(self) -> bool:
```

optimum/neuron/modeling.py
```diff
@@ -506,3 +511,136 @@ def forward(
...
+    )
+    def forward(
+        self,
+        input_ids,
+        cache_ids,
+        start_ids=None,
+        output_hidden_states=False,
+        output_attentions=False,
+        attention_mask=None,
+        return_dict=False,
```
michaelbenayoun, line 588: nit: maybe add type annotation.

optimum/neuron/modeling_decoder.py
michaelbenayoun, line 1: Maybe we should move all the modeling classes and files for inference in a module.
Like `optimum.neuron.inference` or something.

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+            trust_remote_code=trust_remote_code,
+            **model_kwargs,
+        )
+
+        # Save the model checkpoint in a temporary directory
+        checkpoint_dir = TemporaryDirectory()
+        save_pretrained_split(model, checkpoint_dir.name)
+
+        # Update the config
+        config.neuron = {"task": task, "neuron_kwargs": neuron_kwargs}
```
michaelbenayoun, line 130: Not in favor of that because we want to keep the model config similar to what is used in Transformers.

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_exporter(config, task):
+    return TasksManager.get_exporter_config_constructor(model_type=config.model_type, exporter="neuron", task=task)()
+
+
+class NeuronDecoderModel(OptimizedModel):
```
michaelbenayoun, line 49: Why dont we inherint from `NeuronBaseModel`?

test_llm_optimum.py
michaelbenayoun, line 1: What is this file?
Should it be removed?

michaelbenayoun (commented):
Left some comments, but overall looks great!

optimum/exporters/neuron/config.py
dacorvo: I considered this, but NeuronConfig is too specific to the models that were previously integrated: none of its methods nor class constants are actually required by the export code for causal LM models.
For the causl LM models, all we need is:
- some way to identify the corresponding transformers_neuronx model class,
- some way to identify the kwargs that need to be passed when calling from_pretrained on that class.

optimum/exporters/neuron/config.py
dacorvo: These are the default values for these parameters.

Note that for now every single model class in transformers_neuronx has a slightly different set of parameters, but these ones seem to be common. If this is confirmed that these parameters are really common to all models, then I could have declared them directly as named parameters in the _from_transformers method.

optimum/exporters/neuron/config.py
dacorvo: I agree that transformers_neuronx could be omitted and inserted just when the class is imported.

I cannot merge the two constants, but I will need to split them back at import, because using `importlib` I cannot import a single symbol without loading the module that contains it entirely. In other words, I cannot do that:
```
module = importlib.import_module(`transformers_neuronx.gpt2.model.GPT2ForSampling)
```
It will raise the following exception:
```
ModuleNotFoundError: No module named 'transformers_neuronx.gpt2.model.GPT2ForSampling'; 'transformers_neuronx.gpt2.model' is not a package
```
This is why I need first to load the module (I load it with importlib), then access the class as an attribute.

optimum/exporters/neuron/config.py
dacorvo: I think this is a left-over. I will remove it

optimum/neuron/modeling.py
dacorvo: This is actually a method of the PretrainedModel class that is called by the GenerationMixin.generate method.
Since we don't inherit from PretrainedModel we need to provide it explicitly.
This is a copy-paste of what is done in `onnxruntime` GenerationMixin.
I can definitely add the type annotation though.

optimum/neuron/modeling_decoder.py
dacorvo: Again, this is because NeuronBaseModel is too specific: I looked carefully and there isn't a single method or attribute I could reuse here.

optimum/neuron/modeling_decoder.py
dacorvo: I need to keep track of the task (to get the correct exporter), then I also need the named parameters that were used to configure the export.

I saw in NeuronBaseModel that the config is already extended with a lot of items, all named `neuron_XX`.
Do you mean you don't want to store anything neuron-specific in the config ? i certainly could serialize the neuron_config in its own file, like the generation_config, but since @JingyaHuang already had put a lot of stuff there I thought it was ok.

test_llm_optimum.py
dacorvo: Yes, it is in its own "revertme" commit.

dacorvo: I left it just for the time of the review because I don't know if it could be used as a base for a text-generation example (if any is required).

optimum/exporters/neuron/config.py
michaelbenayoun: Then I would move this to the `neuron/base.py` file since it can be considered as a base class in that sense.

optimum/exporters/neuron/config.py
michaelbenayoun: For me it could change, for instance `tp_degree` will need to be bigger for very big models and so on.

optimum/exporters/neuron/config.py
michaelbenayoun: Yes, I think it's okay to split them:

```python
module_name, attribute_name = self.NEURONX_CLASS.rsplit(".", maxsplit=1)
```

optimum/neuron/modeling_decoder.py
michaelbenayoun: I did not know. I think it's better to have everything in another file yes.
Do not want to extend the work on this but yes, in general we try not to change the original config.

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+        checkpoint_dir = TemporaryDirectory()
+        save_pretrained_split(model, checkpoint_dir.name)
+
+        # Update the config
+        config.neuron = {"task": task, "neuron_kwargs": neuron_kwargs}
+
+        return cls._from_pretrained(checkpoint_dir, config)
+
+    @classmethod
+    def _get_neuron_paths(cls, model_dir: Union[str, Path, TemporaryDirectory]) -> Tuple[str, str, str]:
```
JingyaHuang, line 135: What if I want to load from the hub, is it something feasible?

https://github.com/huggingface/optimum-neuron/blob/980dc5cfaa15b23ac3691b78f640a7182e3fd4df/optimum/neuron/modeling_base.py#L141-L143

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # Evaluate the configuration passed during export
+        task = neuron_config["task"]
+        neuron_kwargs = neuron_config["neuron_kwargs"]
+
+        exporter = get_exporter(config, task)
+
+        model_path, checkpoint_path, compiled_path = cls._get_neuron_paths(model_id)
+
+        neuronx_model = exporter.neuronx_class.from_pretrained(checkpoint_path, **neuron_kwargs)
```
JingyaHuang, line 170: For optimum modeling, we usually put the compilation under `from_transformers`, and `from_pretrained` will either load a pre-compiled model or compile it by calling `from_transformers`.

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _save_pretrained(self, save_directory: Union[str, Path]):
+        _, src_chkpt_path, src_compiled_path = self._get_neuron_paths(self.model_path)
+        _, dst_chkpt_path, dst_compiled_path = self._get_neuron_paths(save_directory)
+
+        shutil.copytree(src_chkpt_path, dst_chkpt_path)
+
+        if src_compiled_path is None:
+            # The compiled model has never been serialized: do it now
+            self.model._save_compiled_artifacts(dst_compiled_path)
```
JingyaHuang, line 200: Why don't we serialize the compiled model anyway under `from_transformers` after compilation? Then here we have fewer cases to consider.

optimum/neuron/modeling.py
```diff
@@ -506,3 +511,136 @@ def forward(
...
+        if output_hidden_states or output_attentions or attention_mask is not None:
+            warnings.warn(
+                "Warning: These arguments are not used by forward(): \
+                (output_hidden_states, output_attentions, attention_mask)"
+            )
+        # Evaluate the output logits, storing the current key and values at the indices specified by cache_ids
+        out_logits = self.model.forward(input_ids, cache_ids, start_ids)
+        out_logits = out_logits[:, None, :]
+        # Since we are using a static cache, we don't need to return past keys and values
+        if return_dict:
```
JingyaHuang, line 599: AFAIK, in optimum, we always `return_dict`, maybe just keep the convention.

JingyaHuang (commented):
@dacorvo That's a very awesome Pr, thanks for the work. Congrats!

Left some nits, feel free to ping me for discussion or another review.

optimum/exporters/neuron/config.py
JingyaHuang: @dacorvo @michaelbenayoun if we finally decide to move this class to base.py, call it rather `NeuronDecoderConfig` .

optimum/exporters/neuron/config.py
JingyaHuang: If let's say gpt2 and llama not sharing the same fixed non-customizable parameters, may we put it under `GPT2NeuronConfig` directly?

optimum/neuron/modeling_decoder.py
JingyaHuang: I previously wanted to have a separate config for neuron-specific things, but to move fast I put them under pretrained config. @philschmid 

I now actually think it's not bad to have them in one place.

optimum/exporters/neuron/config.py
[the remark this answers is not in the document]
dacorvo: As explained it is not really possible, but i will move the class as requested.

optimum/exporters/neuron/config.py
dacorvo: Okay, this deserves a better explanation (I was very confused my self when i went through the _transformers-neuronx_ "export").

Firstable, one needs to understand that unlike other models, the export requires a preliminary "modeling/transform" step that happens before the tracing/compilation step.

In the current _transformers_neuronx_ implementation, this step must be performed every time from the source model, even when reloading a model that has already been compiled (this is why I need to copy the "checkpoint" model along the compiled model).

Unfortunately, this step requires the exact same parameters to be passed when the model is first compiled, then reloaded, and believe it or not, the information is not serialized. This is why I need to:
- list the parameters that need to be passed,
- serialize them myself in neuron_config,
- provide my own default values to avoid having one parameter set with a default value by the _transformers_neuronx_ code (note that I actually could have read back the values from the neuronx model `config` member: the one that is not serialized).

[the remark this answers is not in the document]
dacorvo: This is configurable.

dacorvo: > If let's say gpt2 and llama not sharing the same fixed non-customizable parameters, may we put it under `GPT2NeuronConfig` directly?

As explained in my previous answer, the parameters CAN be customized, but I need to provide my own default values. But yes, I could have put them under `GPT2NeuronConfig` directly.

Alternatively, as I already mentioned in another answer, I could at least declare batch_size, tp_degree and amp directly in the _from_transformers signature. These are really common to all models.

I thought n_positions was too, but I just double-checked and it is more used as a way to override the original config.n_positions.

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+
+    def __init__(self, task, **kwargs):
+        if not is_transformers_neuronx_available():
+            raise ModuleNotFoundError("The transformers-neuronx package is required.")
+        module = importlib.import_module(self.NEURONX_MODULE)
+        self._neuronx_class = getattr(module, self.NEURONX_CLASS, None)
+        if self._neuronx_class is None:
+            raise ImportError(f"{self.NEURONX_CLASS} not found in {self.NEURONX_MODULE}. Please check versions.")
+
+    def split_kwargs(self, **kwargs):
```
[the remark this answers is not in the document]
dacorvo: Actually none of these parameters are mandatory, but if we don't provide values they will be set to default values.

As I said before, maybe I could make a difference between the three main parameters and the other ones.

optimum/exporters/neuron/config.py
```diff
@@ -53,3 +56,39 @@ class TextAndVisionNeuronConfig(NeuronConfig):
...
+        and those that need to be passed to the neuron optimizer.
+        """
+        model_kwargs = kwargs
+        neuron_kwargs = {}
+        for arg, default in self.NEURONX_ARGS.items():
+            if arg in model_kwargs:
+                neuron_kwargs[arg] = model_kwargs[arg]
+                model_kwargs.pop(arg)
+            else:
+                neuron_kwargs[arg] = default
```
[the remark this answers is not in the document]
dacorvo: Actually, they will never be any unused `kwargs`: the remaining `model_kwargs` will all be passed to `get_model_from_task`, and it will raise an error if one of them is not supported.
I think that the `modeling_base.py` export code ignores the model `kwargs` when calling `get_model_from_task`, but I cannot do that because some of them are really important like `low_cpu_memory`.

optimum/neuron/modeling.py
```diff
@@ -506,3 +511,136 @@ def forward(
...
+TEXT_GENERATION_EXAMPLE = r"""
+    Example of text generation:
+
+    ```python
+    >>> from transformers import {processor_class}
+    >>> from optimum.neuron import {model_class}
+    >>> import torch
+
+    >>> tokenizer = {processor_class}.from_pretrained("{checkpoint}")
+    >>> model = {model_class}.from_pretrained("{checkpoint}", export=True)
```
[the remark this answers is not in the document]
dacorvo: I think I did at line 572, but maybe I did it wrong ...

optimum/neuron/modeling.py
dacorvo: OK, so do I need to raise an error or add a warning and ignore if return_dict=False ?

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+            use_auth_token=use_auth_token,
+            local_files_only=local_files_only,
+            force_download=force_download,
+            trust_remote_code=trust_remote_code,
+            **model_kwargs,
+        )
+
+        # Save the model checkpoint in a temporary directory
+        checkpoint_dir = TemporaryDirectory()
+        save_pretrained_split(model, checkpoint_dir.name)
```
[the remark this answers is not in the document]
dacorvo: Here, this is not the compiled file yet, but just the original model (a.k.a checkpoint), that is serialized with each weight in a different file under a `pytorch.bin` directory (yes, the directory name is confuzing).

The compiled model (a.k.a compilation artifacts) is a pickle file containing all compiled binaries (one for each decoder layer). The file is named `neuron-program.pkl`.

optimum/neuron/modeling_decoder.py
dacorvo: We need to decide for one solution or the other. I don't have any strong opinion myself, but if I put the neuron config in a separate file it will be just for these models for now.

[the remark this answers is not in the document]
dacorvo: Yes I can.

optimum/neuron/modeling_decoder.py
dacorvo: I don't know, because my knowledge of the interactions with the hub is still very limited. This is something that would come in subsequent pull-requests (I might need guidance).

optimum/neuron/modeling_decoder.py
dacorvo: This is what i did at first, but then I realized I need to apply the "modeling/transform" step also when loading a model from a local path. This is why it happens only later.

optimum/neuron/modeling_decoder.py
dacorvo: Because in from_transformers we don't have the path to where it should be saved. But I agree that everybody will do a save_pretrained after a call to from_pretrained(export=True).

test_llm_optimum.py
```diff
@@ -0,0 +1,96 @@
+import argparse
```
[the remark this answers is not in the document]
dacorvo: I added that in a commit with a revertme in the title: it will be removed later and was just here as an illustration of how the instantiation happens.

I now realize that you are more used to review files all at once than reviewing commits one by one, so this went under your radar and michael's.

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/exporters/neuron/config.py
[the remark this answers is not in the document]
dacorvo: Done

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/exporters/neuron/config.py
dacorvo: Done

optimum/neuron/modeling_decoder.py
```diff
@@ -0,0 +1,208 @@
...
+    ):
+        if not is_transformers_neuronx_available() or not isinstance(model, NeuronxPretrainedModel):
+            raise ValueError("The source model must be a transformers_neuronx.PreTrainedModel.")
+
+        super().__init__(model, config)
+        self.model_path = model_path
+        if generation_config is None:
+            generation_config = GenerationConfig.from_model_config(config)
+        self.generation_config = generation_config
+        self.device = torch.device("cpu")
```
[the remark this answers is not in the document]
dacorvo: I added a comment.

optimum/neuron/modeling_decoder.py
dacorvo: I'll stick with adding a neuron config inside the main config (@JingyaHuang did the same for othe rmodels).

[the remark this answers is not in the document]
dacorvo: Done

optimum/neuron/modeling.py
dacorvo: Done

michaelbenayoun (approved):
LGTM!
Thanks for the great work @dacorvo 

Can be merged once @JingyaHuang approved.

optimum/neuron/modeling_decoder.py
JingyaHuang: Never mind, we can get this in first and add the hub support in a separate PR. 
Basically it is just done through the `HfApi()` of `huggingface_hub`, as how it was done for other optimum modeling: https://github.com/huggingface/optimum-neuron/blob/980dc5cfaa15b23ac3691b78f640a7182e3fd4df/optimum/neuron/modeling_base.py#L141-L148

Feel free to ping me when you start to tackle this.

optimum/neuron/modeling.py
JingyaHuang: @dacorvo Personally don't feel it necessary, what about just remove `return_dict` from input arguments?

optimum/neuron/utils/version_utils.py
```diff
@@ -42,3 +46,23 @@ def get_neuroncc_version() -> str:
         raise ValueError("Neuron Compiler python package is not installed.")
     _neuroncc_version = neuroncc.__version__
     return _neuroncc_version
+
+
+def check_compiler_compatibility(compiler_type: str, compiler_version: str):
```
JingyaHuang, line 51: 👍

JingyaHuang (approved):
LGTM as well!

Amazing job, let's get it merged 🚀 !

JingyaHuang:
Oh btw, let's add a guiding doc for decoder model as well (maybe in a separate PR): 
* https://huggingface.co/docs/optimum-neuron/guides/models
* https://github.com/huggingface/optimum-neuron/blob/main/docs/source/guides/models.mdx

(And also precise the limit like currently only works for gpt2 in the supported architecture: https://huggingface.co/docs/optimum-neuron/package_reference/configuration)

optimum/neuron/modeling.py
dacorvo: I tried to remove it, but I got an error and remembered why I had to put it there in the first place: this is because generate will pass it no matter what.

dacorvo:
> Oh btw, let's add a guiding doc for decoder model as well (maybe in a separate PR):
> 
> * https://huggingface.co/docs/optimum-neuron/guides/models
> * https://github.com/huggingface/optimum-neuron/blob/main/docs/source/guides/models.mdx
> 
> (And also precise the limit like currently only works for gpt2 in the supported architecture: https://huggingface.co/docs/optimum-neuron/package_reference/configuration)

Ok, that will be my next pull-request, now that I finally figured out that I needed to actually edit the mdx files to get something generated 🤭.

optimum/neuron/modeling.py
JingyaHuang: Alright, I see, for onnxruntime it seems that someone added `**kwargs` to let it pass which I feel not the optimal solution either... but why not, let's either adopt that solution or set the default value to `True` at least?

dacorvo: Done

PR #118, opened 2023-06-29
Fixes #114 : Add `neuron_compiler` and `neuron_compiler_version` to c…

### LINKED ISSUE #114: Add neuronx-cc or neuron-cc version to config.json
We noticed that compiled models might not be compatibel across neuron-cc or neuronx-cc versions and raise runtime errors, e.g. 
```bash
File /usr/local/lib/python3.8/dist-packages/torch/jit/_serialization.py:162, in load(f, map_location, _extra_files)
    160 cu = torch._C.CompilationUnit()
    161 if isinstance(f, str) or isinstance(f, pathlib.Path):
--> 162     cpp_module = torch._C.import_ir_module(cu, str(f), map_location, _extra_files)
    163 else:
    164     cpp_module = torch._C.import_ir_module_from_buffer(
    165         cu, f.read(), map_location, _extra_files
    166     )

RuntimeError: Could not load the model status=2 message=Invalid
```
Those provide not the best UX. We should add the compiled neuron version into the config.json as `neuron_version` and assert those when using the NeuronModel class.

dangdana:
…onfig.json

### DIFF
--- a/optimum/neuron/modeling_base.py
+++ b/optimum/neuron/modeling_base.py
@@ -156,8 +156,7 @@ class NeuronBaseModel(OptimizedModel):
             else:
                 file_name = neuron_files[0].name
 
-        # check compiler compatability of saved model vs system installation
-        # Vagrantly assumes that the user won't try to load an xcc on a cc machine (and vise/versa)
+        # Check compiler compatibility(compiler type and version) of the saved model vs. system.
         if hasattr(config, "neuron_compiler") and hasattr(config, "neuron_compiler_version"):
             min_version = get_neuronxcc_version() if config.neuron_compiler == "neuronx-cc" else get_neuroncc_version()
             if version.parse(config.neuron_compiler_version) > version.parse(min_version):
--- a/optimum/neuron/modeling_base.py
+++ b/optimum/neuron/modeling_base.py
@@ -160,24 +160,24 @@ class NeuronBaseModel(OptimizedModel):
         # Check compiler compatibility(compiler type and version) of the saved model vs. system.
         if hasattr(config, "neuron_compiler"):
             if config.neuron_compiler == "neuron-cc":
-                if not is_neuron_available():
-                    raise RuntimeError(
-                        f"Pretrained model was compiled for {config.neuron_compiler}, but {config.neuron_compiler} is not installed."
-                    )
-                min_version = get_neuroncc_version()
+                compiler_available_fn = is_neuron_available
+                installed_compiler_version_fn = get_neuroncc_version
             elif config.neuron_compiler == "neuronx-cc":
-                if not is_neuronx_available():
-                    raise RuntimeError(
-                        f"Pretrained model was compiled for {config.neuron_compiler}, but {config.neuron_compiler} is not installed."
-                    )
-                min_version = get_neuronxcc_version()
+                compiler_available_fn = is_neuronx_available
+                installed_compiler_version_fn = get_neuronxcc_version
             else:
                 raise RuntimeError(f"Pretrained model compiler type {config.neuron_compiler} not recognized.")
-            if version.parse(config.neuron_compiler_version) > version.parse(min_version):
+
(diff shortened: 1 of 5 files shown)


philschmid:
Thank you for opening the PR! We ll look at it. 
This should close then #114

optimum/neuron/modeling_base.py
```diff
@@ -154,6 +156,16 @@ def _from_pretrained(
             else:
                 file_name = neuron_files[0].name
 
+        # check compiler compatability of saved model vs system installation
+        # Vagrantly assumes that the user won't try to load an xcc on a cc machine (and vise/versa)
```
JingyaHuang, line 160:
```suggestion
        # Check compiler compatibility(compiler type and version) of the saved model vs. system.
```

optimum/neuron/modeling_base.py
```diff
@@ -154,6 +156,16 @@ def _from_pretrained(
...
+        if hasattr(config, "neuron_compiler") and hasattr(config, "neuron_compiler_version"):
```
JingyaHuang, line 161: Here I would suggest checking the compiler type(neuronx-cc / neuron-cc) and the version separately.
Like
* First check if the compiler type matches, if not raise.
* Then check if the version is good.

optimum/neuron/modeling_base.py
```diff
@@ -154,6 +156,16 @@ def _from_pretrained(
...
+            min_version = get_neuronxcc_version() if config.neuron_compiler == "neuronx-cc" else get_neuroncc_version()
+            if version.parse(config.neuron_compiler_version) > version.parse(min_version):
```
JingyaHuang, line 163: Here I am curious about the backward compatibility of the neuron compiler. Would a model compiled with an older version of neuron(x)-cc compiler be compatible to run under a newer environment?  @mmcclean-aws

tests/test_modeling.py
```diff
@@ -174,6 +174,13 @@ def test_save_model(self):
             self.assertTrue(NEURON_FILE_NAME in folder_contents)
             self.assertTrue(CONFIG_NAME in folder_contents)
 
+    def test_save_load_model(self):
+        # a config.json can change when converted and saved, so check loadability is maintained
+        with tempfile.TemporaryDirectory() as tmpdirname:
```
JingyaHuang, line 179: Here I am not sure fully understood the test.

Are you worried that there might be a problem with re-loading the config if we save it once again? (for exemple in the case of this PR, users previously saved their models can run `model.save_pretrained()` to update their config with compiler version info)

If so, I think it is good to think of it but not really mandatory to add this test, as `NeuronBaseModel` is designed to be quite tolerant when having extra or lacking of not mandatory information in the config file.

JingyaHuang (commented):
Hey @dangdana, thanks a lot for contributing, the PR looks awesome 💯  !

Left a few nits to discuss.

JingyaHuang:
And to fix the styling, you can run 
```bash
make style
```

tests/test_modeling.py
dangdana: I used a test like this to ensure that the new config.json fields were being correctly processed when the NeuronBaseModel was loaded.  It did catch a bug where I had ">=" on the version check, when I should have just had ">".  

You're right about the tolerances of NeuronBaseModel.  I think this test served its purpose and is unlikely to catch anything in the future.  I'll remove it.

optimum/neuron/modeling_base.py
dangdana: NEFFs are backwards compatible, but official confirmation would be good from @mcclean-aws .

mmcclean-aws: Yes, the Neuron runtime ensures backwards compatibility of NEFFs

JingyaHuang (approved):
LGTM, thanks again for the PR! Let's get it merged! @dangdana

PR #123, opened 2023-07-06
Improve stored config for inference

JingyaHuang:
The way for storing neuron-related configuration done in #117 is better than what I put in place, this PR is inspired to better organize neuron config.

### DIFF
--- a/optimum/neuron/utils/argument_utils.py
+++ b/optimum/neuron/utils/argument_utils.py
@@ -151,24 +151,24 @@ def store_compilation_config(
         update_func = config.__setitem__
     else:
         update_func = config.__setattr__
-    neuron_configs = {}
+    config_args = {}
 
     # Add neuron version to the config, so it can be checked at load time
-    neuron_configs["compiler_type"] = compiler_type
-    neuron_configs["compiler_version"] = compiler_version
+    config_args["compiler_type"] = compiler_type
+    config_args["compiler_version"] = compiler_version
 
     # Add input shapes during compilation to the config
     for axis, shape in input_shapes.items():
         axis = f"static_{axis}"
-        neuron_configs[axis] = shape
+        config_args[axis] = shape
 
-    neuron_configs["dynamic_batch_size"] = dynamic_batch_size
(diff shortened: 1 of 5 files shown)


optimum/neuron/utils/argument_utils.py
```diff
@@ -143,29 +143,32 @@ def store_compilation_config(
...
-    update_func("neuron_compiler", neuron_compiler)
-    update_func("neuron_compiler_version", neuron_compiler_version)
+    neuron_configs["compiler_type"] = compiler_type
+    neuron_configs["compiler_version"] = compiler_version
 
     # Add input shapes during compilation to the config
     for axe, shape in input_shapes.items():
-        axe = f"neuron_{axe}"
-        update_func(axe, shape)
+        axe = f"static_{axe}"
```
dacorvo, line 162: This is really a detail, but `axe` is french, isn't it ? what you really mean is `axis`, not 🪓 right ?

optimum/neuron/utils/argument_utils.py
```diff
@@ -143,29 +143,32 @@ def store_compilation_config(
...
-    neuron_compiler_version: str,
+    compiler_type: str,
+    compiler_version: str,
     **kwargs,
 ):
     if isinstance(config, OrderedDict):
         update_func = config.__setitem__
     else:
         update_func = config.__setattr__
+    neuron_configs = {}
```
dacorvo, line 154: You seem to call it `neuron_config` (singular) in other places.

dacorvo (commented):
LGTM. Just two really minor comments.

optimum/neuron/utils/argument_utils.py
JingyaHuang: That's true... Will change it!

dacorvo (approved):

optimum/neuron/utils/argument_utils.py
JingyaHuang: As the `neuron_configs` here is not the same thing as `neuron_config`(instance of `NeuronConfig`), I should probably not name them the same. I might just go with `config_args`.

PR #125, opened 2023-07-06
Tensor Parallelism fix

michaelbenayoun:
- [x] Tensor parallelism works
- [x] Lazy loading system making TP usable with minimal code change

Current support for TP:

- [x] LLama, **full support**: parallel embeddings, parallel self attention (the `gate_proj` can be handled better via merging), parallel mlp.
- [x] GPTNeo
- [x] BERT
- [x] RoBERTa


examples/language-modeling/run_clm.py
```diff
@@ -390,17 +391,19 @@ def main():
...
-        model = AutoModelForCausalLM.from_pretrained(
-            model_args.model_name_or_path,
-            from_tf=bool(".ckpt" in model_args.model_name_or_path),
-            config=config,
-            cache_dir=model_args.cache_dir,
-            revision=model_args.model_revision,
-            use_auth_token=True if model_args.use_auth_token else None,
-            torch_dtype=torch_dtype,
-        )
+        with lazy_load_for_parallelism(tensor_parallel_size=training_args.tensor_parallel_size):
```
dacorvo, line 394: Instead of using this context before any `AutoModelForCausalLM` when you want to use TP, invoke this context, why didn't you use a neuron/TP specialized child class of `AutoModelForCausalLM` and override the methods that need to be modified ?

optimum/neuron/trainers.py
```diff
@@ -217,6 +247,36 @@ def trigger_on_step_middle_for_neuron_cache_callback(self, model: "PreTrainedMod
...
+    @staticmethod
+    def get_optimizer_cls_and_kwargs(args: TrainingArguments) -> Tuple[Any, Any]:
+        optimizer_cls, optimizer_kwargs = transformers_get_optimizer_cls_and_kwargs(args)
+        if check_if_transformers_greater("4.30.0") and args.tp_plugin.should_parallelize:
+            optimizer_cls = Parallelizer.make_optimizer_constructor_lazy_for_tp(optimizer_cls)
+        return optimizer_cls, optimizer_kwargs
+
+    @patch_within_function(("transformers.Trainer.get_optimizer_cls_and_kwargs", get_optimizer_cls_and_kwargs))
+    def create_optimizer(self):
+        return super().create_optimizer()
```
dacorvo, line 278: This is a general comment: I am really not comfortable with using monkey patching in any deployed code: in my experience it is brittle and very hard to maintain (it is OK for tests though).

examples/language-modeling/run_clm.py
```diff
@@ -390,17 +391,19 @@ def main():
...
+            model = AutoModelForCausalLM.from_pretrained(
```
dacorvo, line 395: Also, couldn't you take advantage of the `low_cpu_mem_usage` parameter ?

dacorvo (commented):
First, congratulations for making it work: this is a very nice contribution.

Now, that's a lot of changes, on top of your previous big pull-request, so apologies for not being able to review in details the updates to the legacy transformers algorithm. I made a few comments along the way.

Finally, I have two questions:

1 - looking at other optimum backends, including the neuron inference backend, It seems that the general pattern/paradigm is to create mirror classes of the transformers classes implementing the same APIs: here you seem to be heading in a different direction, keeping the same classes but modifying their behaviour. Is this an implementation constraint or a design choice ?

2 - you make an heavy use of monkey patching. Again, is it a constraint or a design choice ? If it is a design choice, on what grounds did you take it ?

michaelbenayoun:
> 1 - looking at other optimum backends, including the neuron inference backend, It seems that the general pattern/paradigm is to create mirror classes of the transformers classes implementing the same APIs: here you seem to be heading in a different direction, keeping the same classes but modifying their behaviour. Is this an implementation constraint or a design choice ?

The reason I do this instead of having seperate classes is because then I would need to do the same thing for all the `AutoModel` classes where what I want to change is not really dependent on them. Also we want to keep the example code change as minimal as possible.

> 2 - you make an heavy use of monkey patching. Again, is it a constraint or a design choice ? If it is a design choice, on what grounds did you take it ?

A bit of both. Basically, we want to be able to change the original `Trainer` behaviour without copy pasting big methods just to change one line of code. While this is not neat, this is the best approach we found.

optimum/neuron/trainers.py
michaelbenayoun: I would agree in general, but here we really heavily on `transformers` and `accelerate` classes / methods which can sometimes be huge. 

For instance, the `__init__` method of the `Accelerator` object is 219 lines of code. Basically idea is to monkey-patch in a controlled way.

examples/language-modeling/run_clm.py
michaelbenayoun: `low_cpu_mem_usage` loads ~1x model size in RAM. We actually cannot afford that at all here.

Let's the example is run on 32 cores, if the model is 7B parameters big, it's 14 GB in ram for one worker for nothing since it's going to be parallelized later on.

Instead, we skip loading altogether and postpone it to when we parallelize the model.

michaelbenayoun:
Merging after discussing offline.

PR #130, opened 2023-07-10
Optimize generate sampling loop

dacorvo:
The transformers top-k sampling code is not optimized, as it applies the softmax and multinomial sampling on the full `[batch_size, vocab_size]` tensor, whereas it could apply it on the `[batch_size, top-k]` subset.
This cannot easily be optimized in transformers without breaking the API, but we can provide an optimized neuron sampling loop.

This makes the generation using the transformers API as fast as when using the AWS custom sampling loop, yet still allowing to use transformers-only features (temperature, detailed outputs ...).

The results below are for GPT2 on a `inf2.8xlarge` instance.

|                           | AWS sample | transformers sample | neuron sample |
|---------------------------|------------|---------------------|---------------|
| batch_size 2 128 tokens   | 0.5 s      | 0.8 s               | 0.5 s         |
| batch_size 16 128 tokens  | 0.8 s      | 2.5 s               | 0.8 s         |
| batch_size 2 1000 tokens  | 4.3 s      | 6.8 s               | 4.6 s         |
| batch_size 16 1000 tokens | 10.8 s     | 26.8 s              | 11 s          |

### DIFF
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -662,7 +662,7 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
         return True
 
     class FastTopKLogitsWarper(LogitsWarper):
-        r"""Return [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores"""
+        r"""Returns [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores"""
 
         def __init__(self, top_k: int, filter_value: float):
             self.top_k = top_k
@@ -692,16 +692,12 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
         **model_kwargs,
     ) -> Union[SampleDecoderOnlyOutput, torch.LongTensor]:
         r"""
-        Generates sequences of token ids for models with a language modeling head using **multinomial sampling** and
+        This is a simplified version of the transformers `GenerationMixin.sample()` method that is optimized for neuron inference.
+
+        It generates sequences of token ids for models with a language modeling head using **multinomial sampling** and
         can be used for text-decoder, text-to-text, speech-to-text, and vision-to-text models.
 
-        <Tip warning={true}>
-
-        In most cases, you do not need to call [`~generation.GenerationMixin.sample`] directly. Use generate() instead.
-        For an overview of generation strategies and code examples, check the [following
-        guide](../generation_strategies).
-
-        </Tip>
+        Please refer to https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationMixin.sample.
 
         Parameters:
             input_ids (`torch.LongTensor` of shape `(batch_size, sequence_length)`):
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+                used to tell if the generation loop should stop.
+            logits_warper (`LogitsProcessorList`, *optional*):
+                An instance of [`LogitsProcessorList`]. List of instances of class derived from [`LogitsWarper`] used
+                to warp the prediction score distribution of the language modeling head applied before multinomial
+                sampling at each generation step.
+            pad_token_id (`int`, *optional*):
+                The id of the *padding* token.
+            eos_token_id (`Union[int, List[int]]`, *optional*):
+                The id of the *end-of-sequence* token. Optionally, use a list to set multiple *end-of-sequence* tokens.
+            output_attentions (`bool`, *optional*, defaults to `False`):
```
JingyaHuang, line 723: Any particular reason for not including `max_length` in the docstring?

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+        pad_token_id: Optional[int] = None,
+        eos_token_id: Optional[Union[int, List[int]]] = None,
+        output_attentions: Optional[bool] = None,
+        output_hidden_states: Optional[bool] = None,
+        output_scores: Optional[bool] = None,
+        return_dict_in_generate: Optional[bool] = None,
+        synced_gpus: bool = False,
+        streamer: Optional["BaseStreamer"] = None,
+        **model_kwargs,
+    ) -> Union[SampleDecoderOnlyOutput, torch.LongTensor]:
```
JingyaHuang, line 693: Might it be reusable by encoder-decoder architecture? If so I think it would be better to create a `NeuronGenerationMixin` and IMO it should also boost the performance of other module in Optimum, if so let's put it in optimum main repo @fxmarty

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+            [`~generation.SampleDecoderOnlyOutput`], [`~generation.SampleEncoderDecoderOutput`] or `torch.LongTensor`:
+            A `torch.LongTensor` containing the generated tokens (default behaviour) or a
+            [`~generation.SampleDecoderOnlyOutput`] if `model.config.is_encoder_decoder=False` and
+            `return_dict_in_generate=True` or a [`~generation.SampleEncoderDecoderOutput`] if
+            `model.config.is_encoder_decoder=True`.
+
+        Examples:
+
+        ```python
+        >>> from transformers import (
```
JingyaHuang, line 752: The snippet is a little redundant here.

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
     def can_generate(self) -> bool:
         """Returns True to validate the check made in `GenerationMixin.generate()`."""
         return True
+
+    class FastTopKLogitsWarper(LogitsWarper):
+        r"""Return [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores"""
```
michaelbenayoun, line 665:
```suggestion
        r"""Returns [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores"""
```

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+        ...     logits_processor=logits_processor,
+        ...     logits_warper=logits_warper,
+        ...     stopping_criteria=stopping_criteria,
+        ... )
+
+        >>> tokenizer.batch_decode(outputs, skip_special_tokens=True)
+        ['Today is a beautiful day, and we must do everything possible to make it a day of celebration.']
+        ```"""
+        # We don't support all parameters
+        if synced_gpus:
```
JingyaHuang, line 802: If we override the function, we can consider excluding those useless parameters and associated code?

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+        >>> outputs = model.sample(
+        ...     input_ids,
+        ...     logits_processor=logits_processor,
+        ...     logits_warper=logits_warper,
+        ...     stopping_criteria=stopping_criteria,
+        ... )
+
+        >>> tokenizer.batch_decode(outputs, skip_special_tokens=True)
+        ['Today is a beautiful day, and we must do everything possible to make it a day of celebration.']
+        ```"""
```
michaelbenayoun, line 800: Not sure if we need to copy / paste the docstring. Also it does not respect our format which does not use the `*optional*` part.

JingyaHuang (commented):
Looks great! Left some points to discuss a little bit.

optimum/neuron/modeling.py
```diff
@@ -653,3 +660,286 @@ def prepare_inputs_for_generation(self, input_ids: torch.Tensor, **kwargs) -> Di
...
+
+            next_token_logits = outputs.logits[:, -1, :]
+
+            # pre-process distribution
+            next_token_scores = logits_processor(input_ids, next_token_logits)
+            if fast_topk:
+                # Get [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores
+                next_token_scores, next_token_indices = logits_warper(input_ids, next_token_scores)
+            else:
+                next_token_scores = logits_warper(input_ids, next_token_scores)
```
michaelbenayoun, line 872: That is the specific part we added? 
If not, let's comment to show where we differ from the original implementation.

michaelbenayoun (approved):
LGTM!
Really nice improvement in terms of performance @dacorvo

optimum/neuron/modeling.py
dacorvo: This comes straight from the transformers code-base. I think they forgot it ...

optimum/neuron/modeling.py
dacorvo: There is already a `NeuronGenerationMixin`, but for training, and with a different implementation (it completely overloads `generate()` for a start). 

Regarding the applicability of that optimization to other backends, I did some tests on another ec2 instance that uses intel CPUs, and the results are not so convincing (the inf2 instances use AMD CPUs).

optimum/neuron/modeling.py
dacorvo: Again, this is the docstring from transformers. I will remove most of the details and maybe just link to the original method documentation.

optimum/neuron/modeling.py
dacorvo: OK, I will just put a reference to the transformers doc then.

optimum/neuron/modeling.py
dacorvo: I left it because it will be passed by generate no matter what. This is a general issue with the methods we overload. My policy so far: if the parameter is always there, declare it and tell you will ignore it. The other option is just to let it  sink in the kwargs that are ignored, but I am a bit reluctant to do that.

optimum/neuron/modeling.py
dacorvo: OK, I will do that. 

You can track the specific changes in the commits though: I first copied the method in one commit and then modified it in a second commit.
I understand that all commits in a pull-request are squashed, but in that particular configuration I find it convenient to keep separate commits.

JingyaHuang (approved):
LGTM, thanks for the PR.

optimum/neuron/modeling.py
michaelbenayoun: It is more to be able to know fast what is neuron-specific vs what comes from transformers.

PR #133, opened 2023-07-13
Add support for pushing and fetching Neuron compiled text-generation models from the hub

dacorvo:
Pre-compiled text-generation neuron models can now be pushed to the hub and retrieved later, allowing to perform an inference without recompiling the model.

### DIFF
--- a/optimum/neuron/modeling_decoder.py
+++ b/optimum/neuron/modeling_decoder.py
@@ -22,6 +22,7 @@ from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING, Optional, Tuple, Union
 
 import torch
+from huggingface_hub import HfApi, HfFolder
 from transformers import GenerationConfig
 
 from ..exporters.neuron.model_configs import *  # noqa: F403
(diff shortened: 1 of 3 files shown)


optimum/neuron/modeling_decoder.py
```diff
@@ -224,3 +236,45 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
...
+        if isinstance(use_auth_token, str):
+            huggingface_token = use_auth_token
+        elif use_auth_token:
+            huggingface_token = HfFolder.get_token()
+        else:
+            raise ValueError("You need to provide `use_auth_token` to be able to push to the hub")
+        api = HfApi(endpoint=endpoint)
+
+        user = api.whoami(huggingface_token)
+        self.git_config_username_and_email(git_email=user["email"], git_user=user["fullname"])
```
michaelbenayoun, line 257: Why?

optimum/neuron/modeling_decoder.py
```diff
@@ -224,3 +236,45 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
...
+        )
+        for path, subdirs, files in os.walk(save_directory):
+            for name in files:
+                local_file_path = os.path.join(path, name)
+                hub_file_path = os.path.relpath(local_file_path, save_directory)
+                # FIXME: when huggingface_hub fixes the return of upload_file
+                try:
+                    api.upload_file(
+                        token=huggingface_token,
+                        repo_id=f"{repository_id}",
```
michaelbenayoun, line 273:
```suggestion
                        repo_id=repository_id,
```

optimum/neuron/modeling_decoder.py
```diff
@@ -224,3 +236,45 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
...
+                        path_or_fileobj=os.path.join(os.getcwd(), local_file_path),
+                        path_in_repo=hub_file_path,
+                    )
+                except KeyError:
+                    pass
+                except NameError:
+                    pass
```
michaelbenayoun, line 280: Why?

michaelbenayoun (approved):
LGTM, great feature @dacorvo !

optimum/neuron/modeling_decoder.py
dacorvo: This is a copy/paste from optimum. I assume it allows to identify the author if it is not the same as the one identified by the token (same thing ads when you add a commit from someone else in your branch/pull-request).

optimum/neuron/modeling_decoder.py
dacorvo: Also a copy-paste, but yes I will change it.

optimum/neuron/modeling_decoder.py
dacorvo: Also a copy/paste. I have np idea what the fix is and it it was eventually pushed. I'll check with the original committer in optimum.

dacorvo: I removed it. We will see if it breaks under some configurations, but I don't think so.

PR #137, opened 2023-07-14
Fix recompilation in greedy sampling

bocchris-aws:
Before moving tensors to the CPU to be processed by the `logit_processor` in greedy search, we dynamically slice it to `seq_length`. Since with each iteration of generating the next tokens this variable increases, recompilation is triggered. This PR changes the order and moves the tensor to CPU before slicing it to prevent recompilation.


bocchris-aws:
Closing as accidentally other changes got included.

PR #164, opened 2023-08-08
minor doc fix

oOraph:
fix bash comments on multiline commands


JingyaHuang (approved):
LGTM, thanks a lot for the fix @oOraph !

PR #167, opened 2023-08-10
[Stable Diffusion] Fix the image size value inferring

JingyaHuang:
In the dummy input generator, the default value of `image_size` would be the `sample_size`, which would lead to shapes using for compilation unmatch with the `height` and `width` given by users. 

This PR is suppose to fix how the `height` and `width` should be infer.

### DIFF
--- a/optimum/exporters/neuron/__main__.py
+++ b/optimum/exporters/neuron/__main__.py
@@ -244,8 +244,7 @@ def main_export(
             )
 
             logger.info(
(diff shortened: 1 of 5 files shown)


JingyaHuang:
Also gently pinging @oOraph and @juliensimon, if you are using `optimum-neuron` 0.0.9, if the `height`(`width`) divide by the `vae_scale_factor` factor doesn't  equal to the `sample_size` in the checkpoint's config file, the inference might fail due to inconsistent latent shape without the PR.

PR #173, opened 2023-08-17
Support multiple prompts for generation example

dacorvo:
This allows multiple prompts separated by '|' to be passed on the command line, and therefore removes the need for the batch_size parameter.


JingyaHuang (approved):
LGTM!

PR #178, opened 2023-08-18
Enable exporter on non INF instances

JingyaHuang:
Add the possibility to disable the validation in `optimum-cli` for better export experience on CPU only instances.

* Use the flag `--disable-validation` when export with no neuron devices
```
optimum-cli export neuron --model yiyanghkust/finbert-tone --disable-validation --sequence_length 128 --batch_size 1 test_issue/
```

### DIFF
--- a/optimum/commands/export/neuron.py
+++ b/optimum/commands/export/neuron.py
@@ -61,7 +61,7 @@ def parse_args_neuron(parser: "ArgumentParser"):
     optional_group.add_argument(
         "--disable-validation",
         action="store_true",
-        help="Whether to disable the validation of inference on neuron device compared to the outputs of original PyTorch model on CPU.",
+        help="Whether to disable the validation of the exported model on neuron device compared to the outputs of original PyTorch model on CPU.",
     )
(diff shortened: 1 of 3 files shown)


optimum/commands/export/neuron.py
```diff
@@ -58,6 +58,11 @@ def parse_args_neuron(parser: "ArgumentParser"):
         action="store_true",
         help="Allow to use custom code for the modeling hosted in the model repository. This option should only be set for repositories you trust and in which you have read the code, as it will execute on your local machine arbitrary code present in the model repository.",
     )
+    optional_group.add_argument(
+        "--disable-validation",
+        action="store_true",
+        help="Whether to disable the validation of inference on neuron device compared to the outputs of original PyTorch model on CPU.",
```
michaelbenayoun, line 64:
```suggestion
        help="Whether to disable the validation of the exported model on neuron device compared to the outputs of original PyTorch model on CPU.",
```

michaelbenayoun (approved):

PR #182, opened 2023-08-21
Improve default compilation arguments for stable diffusion

JingyaHuang:
* With experiment,  casting `matmul` operations from `FP32` to `BF16` can get a reasonable trade-off between inference performance and models' accuracy. Currently default does no cast which is slower than some GPU and might mislead users.
* Set default device ids to [0, 1] as we want automatically load the compiled unet to both neuron cores.

### DIFF
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -78,7 +78,7 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
         tokenizer: CLIPTokenizer,
         scheduler: Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler],
         feature_extractor: Optional[CLIPFeatureExtractor] = None,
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling_diffusion.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         tokenizer: CLIPTokenizer,
         scheduler: Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler],
         feature_extractor: Optional[CLIPFeatureExtractor] = None,
-        device_ids: Optional[List[int]] = [],
+        device_ids: Optional[List[int]] = [0, 1],
```
michaelbenayoun, line 81: Let's keep it to `None` here and set it to `[0, 1]` if `device_ids=None`. 
It is not good practice to set mutable objects as default values.

michaelbenayoun (approved):

PR #206, opened 2023-08-29
Bump Inference APIs to Neuron 2.13

JingyaHuang:
- [x] Unblock VAE encoder
- [x] Remove optimized cross-attention score workaround
- [x] Unblock dynamic batching test
- [x] Update documentation

### DIFF
--- a/docs/source/guides/export_model.mdx
+++ b/docs/source/guides/export_model.mdx
@@ -87,10 +87,12 @@ The Optimum Neuron export can be used through Optimum command-line:
 ```bash
 optimum-cli export neuron --help
 
-usage: optimum-cli export neuron [-h] -m MODEL [--task TASK] [--atol ATOL] [--cache_dir CACHE_DIR] [--trust-remote-code] [--disable-validation]
-                                 [--auto_cast {none,matmul,all}] [--auto_cast_type {bf16,fp16,tf32}] [--dynamic-batch-size]
-                                 [--batch_size BATCH_SIZE] [--sequence_length SEQUENCE_LENGTH] [--num_choices NUM_CHOICES]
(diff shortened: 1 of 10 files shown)


docs/source/guides/models.mdx
```diff
@@ -201,10 +201,10 @@ with torch.inference_mode():
 Optimum extends 🤗`Diffusers` to support inference on Neuron. To get started, make sure you have installed Diffusers:
 
 ```bash
-pip install diffusers>=0.17.0
+pip install diffusers>=0.20.0
```
philschmid, line 204: should we add an optional with [diffusers] to the setup.p?

docs/source/guides/export_model.mdx
```diff
@@ -226,6 +231,7 @@ optimum-cli export neuron --model stabilityai/stable-diffusion-2-1-base \
   --batch_size 1 \
   --height 512 `# height in pixels of generated image, eg. 512, 768` \
   --width 512 `# width in pixels of generated image, eg. 512, 768` \
+  --num_image_per_prompt 4 `# number of images to generate per prompt, defaults to 1` \
```
philschmid, line 234: is this needed with dynamic batching support now?

philschmid (approved):
LGTM!

dacorvo (approved):

michaelbenayoun (approved):
LGTM

docs/source/guides/models.mdx
JingyaHuang: We already have diffusers as a soft dependency in `setup.py`, let me update the doc using `optimum[neuronx, diffusers]` extra here.

docs/source/guides/export_model.mdx
JingyaHuang: If we turn on dynamic batching, we can input any batch size / number images per prompt, but by default the dynamic batch size is turned off, so we can keep the snippet this way I think.

Issue #193, opened 2023-08-24
T5-11B fine-tuning crashes


aws-rhsoln:
I tried running t5-11B fine-tuning using the run_summarization example and the `main` branch. This is the command I used to run:
```
NUM_WORKERS=32
WORLD_SIZE=1
NODE_ID=0
MASTER_ADDRESS=localhost
XLA_DOWNCAST_BF16=1  torchrun --nproc_per_node=$NUM_WORKERS --nnodes=$WORLD_SIZE --node_rank=$NODE_ID --master_addr=$MASTER_ADDRESS --master_port=44000 \
     run_summarization.py \
    --model_name_or_path google/flan-t5-small \
    --do_train \
    --do_eval false \
    --do_predict false \
    --dataset_name cnn_dailymail \
    --dataset_config "3.0.0" \
    --source_prefix "summarize: " \
    --output_dir /tmp/tst-summarization \
    --tensor_parallel_size $TP \
    --per_device_train_batch_size=1 \
    --per_device_eval_batch_size=1 \
    --zero_1 \
    --log_level info \
    --overwrite_output_dir
```

Here are the observations:
1. After running 3 compile steps (used neuron_parallel_compile), it results in HOST OOM. I see that the host memory utilization keeps growing at every step.
2. To debug the memory issue, I reduced the size of the model to t5-small making TP=2 and num_workers=2. Here, I see that new compilations are generated at every step. I see the size of the graph keeps increasing by 3K at every step. This can point to a missing mark_step or some tensors that are not detached before saving them for logging purposes.
3. I also noticed that the first 7-8 iterations doesn't produce any graph and the first graph appears directly at 9th step. This indicates that there is some sort of batching been done before executing the graph. 
4. I see that we are slicing the relative_attention_bias here: https://github.com/huggingface/optimum-neuron/blob/main/optimum/neuron/distributed/encoder_decoder_models.py#L59 . This would cause a graph per tp rank. One way we can avoid this is by having a special RelativeAttentionBiasEmbedding where we do something like this:
```
tp_rank = parallel_layers.parallel_state.get_tensor_model_parallel_rank()
            size_per_tp = model.decoder.block[0].layer[0].SelfAttention.relative_attention_bias.weight.shape[1]//parallel_layers.parallel_state.get_tensor_model_parallel_size()
[...]

PR #207, opened 2023-08-30
flan-t5 fix: `T5Parallelizer`, `NeuronCacheCallback` and `NeuronHash` refactors

michaelbenayoun:
This PR includes:

- [x] An embedding that is not the vocabulary embedding that is parallelized, and with data available in the checkpoint ends-up re-initialized as unparallelized. This is the case of the `relative_attention_bias` in T5. This is fixed here.  
- [x] Fixes `T5Parallelizer`. The way T5 was parallized was producing different graphs per tp ranks, and was incorrect. This is fixed here. (cc @aws-rhsoln)
- [x] A temporary directory was created by each worker as a temporary neuron cache. Now rank 0 creates a temporary neuron cache, and the other workers use it. All of this setup is done before `torch.distributed.init_process_group` for it to be taken into account by the Neuron compiler. The share of information is done by a `torch.distributed.TCPStore`.
- [x] `NeuronHash` refactor. The fact that a `NeuronHash` stored an instance of the model was causing issues during compilation for `flan-t5`. It is also unnecessary to keep a reference of the model instance inside each `NeuronHash`. Now:
     - A `NeuronHash` is instantiated with a model instance but it is not an attribute anymore
     - Instead we store private attributes that were before deduced from the model: `_model_name_or_path`, `_is_private` (could be inferred from `_model_name_or_path`) and `_model_type`.

Fixes: #193

### DIFF
--- a/optimum/neuron/distributed/encoder_decoder_models.py
+++ b/optimum/neuron/distributed/encoder_decoder_models.py
@@ -16,6 +16,9 @@
 
 from typing import TYPE_CHECKING, Dict, Optional
 
+import torch
+
+from ...utils import NormalizedConfigManager
 from .base import Parallelizer
 from .parallel_layers import ParallelEmbedding, ParallelMLP, ParallelSelfAttention
 from .utils import linear_to_parallel_linear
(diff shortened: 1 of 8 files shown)


optimum/neuron/distributed/encoder_decoder_models.py
```diff
@@ -47,21 +50,43 @@ def transform(
...
+
+        layer = super().transform(model, layer, orig_to_parallel=orig_to_parallel, device=device)
 
         orig_compute_bias = layer.compute_bias
 
         def compute_bias(self, query_length, key_length, device=None):
             """Compute binned relative position bias"""
             values = orig_compute_bias(query_length, key_length, device=device)
-            return values[:, tp_rank * num_heads : (tp_rank + 1) * num_heads, :]
+            return values[:, tp_rank * num_attention_heads_per_rank : (tp_rank + 1) * num_attention_heads_per_rank, :]
```
aws-rhsoln, line 78: THis line is going to cause a graph for each TP rank.

optimum/neuron/trainers.py
```diff
@@ -87,6 +87,12 @@
 # Used for torch.distributed.
 _ORIGINAL_NEURON_CACHE_PATH: Optional[Path] = None
 _TMP_NEURON_CACHE_DIR: Optional[TemporaryDirectory] = None
+_TMP_NEURON_CACHE_PATH: Optional[Path] = None
+_TCP_STORE_ADDRESS = "127.0.0.1"
+_TCP_STORE_PORT = 5000
+
+# TODO: remove that.
+os.environ["NEURON_CC_FLAGS"] = "--log_level=INFO"
```
dacorvo, line 95: Do not forget to actually remove that from the pull-request.

optimum/neuron/utils/cache_utils.py
```diff
@@ -564,13 +564,31 @@ class NeuronHash:
...
+            is_private = True
+        elif Path(model_name_or_path).exists():
+            is_private = True
+        else:
+            is_private = is_private_repo(model_name_or_path)
+
+        # Using object.__setattr__ to change the field value because NeuronHash is supposed to be frozen.
+        # Not very clear, but it should work here.
+        super().__setattr__("_is_private", is_private)
+        super().__setattr__("_model_type", model.config.model_type)
```
dacorvo, line 589: Not super clear indeed: all I can see is that these members are excluded from the hash computation down below ...
Can you clarify a bit ?

dacorvo (changes requested):

optimum/neuron/utils/cache_utils.py
michaelbenayoun: You can refer to the notes that I have just added to the PR for more information (last point in the list), but basically storing a reference to the model in the `NeuronHash` was causing issues during compilation. 

It is actually not needed per-se to have a reference to the model so I got rid of it. Instead now `model` is annotated as an `InitVar`, meaning that it will be a parameter to the initializer of the `NeuronHash` dataclass, but not an attribute. 
During initialization, the model argument is then used to compute values that are needed by the class that were previously inferred from the model reference.  

The reason `super().__setattr__` is used here is because `NeuronHash` is a frozen dataclass (to make it hashable), but we need to set those values at post_init time... This is a bit of hack, but it should be ok here since we never change anything after initialization.

dacorvo (approved):
LGTM. Thanks for the detailed explainations (very clear) !

PR #209, opened 2023-09-04
Align NeuronModelForCausalLM with latest transformers-neuronx and prepare TGI

dacorvo:
This pull-request includes several changes:

- first, it improves the Makefile build target,
- second, it improves `NeuronModelForCausalLM` compilation speed by using an experimental compiler flag,
- third, it moves the token selection code into a separate class that will be shared with the Inf2 TGI server.

### DIFF
--- a/optimum/neuron/generation/logits_process.py
+++ b/optimum/neuron/generation/logits_process.py
@@ -5,9 +5,8 @@ from transformers.generation import LogitsWarper
 class FastTopKLogitsWarper(LogitsWarper):
     r"""Returns [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores."""
 
-    def __init__(self, top_k: int, filter_value: float):
+    def __init__(self, top_k: int):
         self.top_k = top_k
-        self.filter_value = filter_value
 
     def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
         top_k = min(self.top_k, scores.size(-1))  # Safety check
--- a/optimum/neuron/generation/logits_process.py
+++ b/optimum/neuron/generation/logits_process.py
@@ -1,0 +1,15 @@
+import torch
+from transformers.generation import LogitsWarper
+
+
+class FastTopKLogitsWarper(LogitsWarper):
+    r"""Returns [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores."""
+
+    def __init__(self, top_k: int, filter_value: float):
+        self.top_k = top_k
+        self.filter_value = filter_value
+
+    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
+        top_k = min(self.top_k, scores.size(-1))  # Safety check
+        # Remove all tokens with a probability less than the last token of the top-k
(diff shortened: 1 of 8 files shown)


optimum/neuron/generation/logits_process.py
```diff
@@ -0,0 +1,15 @@
...
+    r"""Returns [batch_size, top_k] scores and indices instead of [batch_size, vocab_size] scores."""
+
+    def __init__(self, top_k: int, filter_value: float):
+        self.top_k = top_k
+        self.filter_value = filter_value
+
+    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
+        top_k = min(self.top_k, scores.size(-1))  # Safety check
+        # Remove all tokens with a probability less than the last token of the top-k
+        return torch.topk(scores, top_k)
```
michaelbenayoun, line 15: Where is `filter_value` used?

optimum/neuron/generation/token_selector.py
```diff
@@ -0,0 +1,179 @@
...
+            self.fast_topk = isinstance(last_warper, TopKLogitsWarper)
+            if self.fast_topk:
+                # Replace the last warping operation by a faster alternative
+                self.logits_warper[-1] = FastTopKLogitsWarper(last_warper.top_k, last_warper.filter_value)
+
+    @classmethod
+    def create(
+        cls, input_ids: torch.Tensor, generation_config: GenerationConfig, model: GenerationMixin, max_seq_length: int
+    ) -> "TokenSelector":
+        r"""Create the `TokenSelector` for a specific generation configuration.
```
michaelbenayoun, line 66:
```suggestion
        r"""Creates the `TokenSelector` for a specific generation configuration.
```

optimum/neuron/modeling_decoder.py
```diff
@@ -205,8 +205,12 @@ def _from_pretrained(
             neuronx_model._load_compiled_artifacts(compiled_path)
 
         # Compile the Neuron model (if present compiled artifacts will be reloaded instead of compiled)
-        os.environ["NEURON_CC_FLAGS"] = "--model-type=transformer-inference"
+        neuron_cc_flags = os.environ.get("NEURON_CC_FLAGS", "")
+        os.environ["NEURON_CC_FLAGS"] = (
+            neuron_cc_flags + " --model-type=transformer-inference --enable-experimental-O1"
+        )
```
michaelbenayoun, line 211: nit: I would use a regex to:

1. Check that `--model-type` was not already set, otherwise we would be setting two values
2. Check that `--enable-experimental-O1` was not already set as well. This is less problematic because it's a flag compared to `--model-type`.

optimum/neuron/generation/logits_process.py
dacorvo: Good catch: we don't need it because we don't return the filtered tokens. Removed.

optimum/neuron/modeling_decoder.py
dacorvo: Since I add the flags at the end of the existing flags, I actually override model-type, which is absolutely required (compilation will fail if it is not set).
I restore the previous flags afterwards, so it is safe.

michaelbenayoun (approved):
LGTM!

PR #214, opened 2023-09-06
AWS Inferentia2 TGI server

dacorvo:
This adds:

- a TGI-compatible python server to serve `optimum-neuron` causal LM models (`NeuronModelForCausalLM`),
- a docker image for deployment of a service on Inferentia 2 hosts.

Please refer to `text-generation-inference/README.md` for details.

### DIFF
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -52,6 +52,12 @@ if __name__ == "__main__":
         default="One of my fondest memory is",
         help="The prompts to use for generation, using | as separator.",
     )
+    parser.add_argument(
+        "--batch_size",
+        type=int,
+        default=None,
+        help="The batch size (optional). If not specified it wil be based on the number of prompts.",
+    )
     parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
     parser.add_argument(
         "--num_cores", type=int, default=2, help="The number of cores on which the model should be split."
@@ -72,7 +78,9 @@ if __name__ == "__main__":
     if args.seed is not None:
         set_seed(args.seed)
     prompts = args.prompts.split("|")
-    batch_size = len(prompts)
+    batch_size = len(prompts) if args.batch_size is None else args.batch_size
+    if len(prompts) < batch_size:
+        prompts = prompts + [prompts[-1]] * (batch_size - len(prompts))
     model = load_llm_optimum(args.model, batch_size, args.num_cores, args.auto_cast_type)
     tokenizer = AutoTokenizer.from_pretrained(args.model)
     outputs, latency = generate(model, tokenizer, prompts, args.length, args.temperature)
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -60,7 +60,7 @@ if __name__ == "__main__":
     )
     parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
     parser.add_argument(
-        "--num_cores", type=int, default=2, help="The number of cores on which the model should be split."
+        "--num_cores", type=int, default=1, help="The number of cores on which the model should be split."
     )
     parser.add_argument("--auto_cast_type", type=str, default="f32", help="One of f32, f16, bf16.")
     parser.add_argument(
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -56,7 +56,7 @@ if __name__ == "__main__":
         "--batch_size",
         type=int,
         default=None,
-        help="The batch size (optional). If not specified it wil be based on the number of prompts.",
+        help="The batch size (optional). If not specified it will be based on the number of prompts.",
     )
     parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
     parser.add_argument(
--- a/examples/text-generation/run_generation.py
+++ b/examples/text-generation/run_generation.py
@@ -62,7 +62,9 @@ if __name__ == "__main__":
     parser.add_argument(
         "--num_cores", type=int, default=1, help="The number of cores on which the model should be split."
     )
-    parser.add_argument("--auto_cast_type", type=str, default="f32", help="One of f32, f16, bf16.")
+    parser.add_argument(
+        "--auto_cast_type", type=str, default="f32", choices=["f32", "f16", "bf16"], help="One of f32, f16, bf16."
+    )
     parser.add_argument(
         "--temperature",
         type=float,
--- a/optimum/neuron/modeling.py
+++ b/optimum/neuron/modeling.py
@@ -688,29 +688,41 @@ class NeuronModelForCausalLM(NeuronDecoderModel, GenerationMixin):
 
         # Verify that the inputs are compatible with the model static input dimensions
         batch_size, sequence_length = input_ids.shape
-        if batch_size != self.batch_size:
-            raise ValueError(
-                f"The specified batch_size ({batch_size}) does not match the model static batch size ({self.batch_size})"
-            )
         if sequence_length > self.max_length:
             raise ValueError(
                 f"The input sequence length ({sequence_length}) exceeds the model static sequence length ({self.max_length})"
             )
-
+        padded_input_ids = input_ids
+        padded_attention_mask = attention_mask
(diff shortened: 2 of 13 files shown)


text-generation-inference/Dockerfile
```diff
@@ -0,0 +1,132 @@
+# Fetch and extract the TGI sources
+FROM alpine AS tgi
+RUN mkdir -p /tgi
+# Note that we should use a release instead of main, but v1.0.1 is broken
```
philschmid, line 4: there is 1.0.3 which you can test

text-generation-inference/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+    transformers-neuronx==0.5.58 \
+    --extra-index-url=https://pip.repos.neuron.amazonaws.com
+
+# Install HuggingFace packages
+RUN pip3 install \
+    hf_transfer
+
+# Install optimum-neuron
+COPY dist/optimum-neuron-${VERSION}.tar.gz optimum-neuron.tar.gz
+RUN pip3 install optimum-neuron.tar.gz
```
philschmid, line 113: Why copy a tar and not the code or not install with pip?

text-generation-inference/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+COPY text-generation-inference/server server
+COPY --from=tgi /tgi/proto proto
+RUN pip3 install -r server/build-requirements.txt
+RUN VERBOSE=1 BUILDDIR=/pyserver/build PROTODIR=/pyserver/proto make -C server gen-server
+
+# Neuron base image (used for deployment)
+FROM base AS neuron
+
+# VERSION is a mandatory parameter
+ARG VERSION
```
philschmid, line 76:
```suggestion
ARG OPTIMUM_NEURON_VERSION
```

philschmid (commented):
Left some comments on the container file. Will give it a try later 🔥

text-generation-inference/Dockerfile
dacorvo: I am installing with pip , but the docker COPY is required for the next RUN command to see the file.

text-generation-inference/Dockerfile
dacorvo: Yes, I need to bump the version. They added some fields to the protobuf lately that I don't support though, so I need to check.

text-generation-inference/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+RUN test -n ${VERSION:?}
+
```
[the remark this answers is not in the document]
dacorvo: If I do that, instead of updating the Dockerfile when we change version, I would need to update the top Makefile, which is equivalent.
Maybe we can do something smarter and set them once for the whole repository ? That would mean finding a way to reflect them in the setup.py.

examples/text-generation/run_generation.py
```diff
@@ -52,9 +52,15 @@ def generate(model, tokenizer, prompts, length, temperature):
         default="One of my fondest memory is",
         help="The prompts to use for generation, using | as separator.",
     )
+    parser.add_argument(
+        "--batch_size",
+        type=int,
+        default=None,
+        help="The batch size (optional). If not specified it wil be based on the number of prompts.",
```
michaelbenayoun, line 59:
```suggestion
        help="The batch size (optional). If not specified it will be based on the number of prompts.",
```

examples/text-generation/run_generation.py
```diff
@@ -52,9 +52,15 @@ def generate(model, tokenizer, prompts, length, temperature):
...
+    )
     parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
     parser.add_argument(
-        "--num_cores", type=int, default=2, help="The number of cores on which the model should be split."
+        "--num_cores", type=int, default=1, help="The number of cores on which the model should be split."
     )
     parser.add_argument("--auto_cast_type", type=str, default="f32", help="One of f32, f16, bf16.")
```
michaelbenayoun, line 65:
```suggestion
    parser.add_argument("--auto_cast_type", type=str, default="f32", choices=["f32", "f16", "bf16"], help="One of f32, f16, bf16.")
```
Also I would argue that it would be better if it was `fp32`, `fp16` and `bf16` as it is more aligned with PyTorch.

text-generation-inference/server/text_generation_server/cli.py
```diff
@@ -0,0 +1,101 @@
...
+app = typer.Typer()
+
+
+@app.command()
+def serve(
+    model_id: str,
+    revision: Optional[str] = None,
+    sharded: bool = False,
+    trust_remote_code: bool = None,
+    uds_path: Path = "/tmp/text-generation-server",
```
michaelbenayoun, line 18: nit: not really a `Path` here.
```suggestion
    uds_path: Union[Path, str] = "/tmp/text-generation-server",
```

text-generation-inference/server/text_generation_server/cli.py
```diff
@@ -0,0 +1,101 @@
...
+        model_id (`str`):
+            The *model_id* of a model on the HuggingFace hub or the path to a local model.
+        revision (`str`):
+            The revision of the model on the HuggingFace hub.
+        sharded (`bool`):
+            Whether the model must be sharded or not. Kept for compatibility with the
+            text-generation-launcher, but must be set to False.
+        trust-remote-code (`bool`):
+            Kept for compatibility with text-generation-launcher. Ignored.
+        uds_path (`str`):
```
michaelbenayoun, line 34:
```suggestion
        uds_path (`Union[Path, str]`):
```

text-generation-inference/server/text_generation_server/model.py
```diff
@@ -0,0 +1,37 @@
...
+def fetch_model(
+    model_id: str,
+    revision: Optional[str] = None,
+) -> str:
+    """Fetch a neuron model.
+
+    Args:
+        model_id (`str`):
+            The *model_id* of a model on the HuggingFace hub or the path to a local model.
+        revision (`str`):
```
michaelbenayoun, line 18:
```suggestion
        revision (`Optional[str]`, defaults to `None`):
```

michaelbenayoun (approved):
I am not very familiar with TGI so commented on what I could comment.
Awesome work @dacorvo !

JingyaHuang (approved):
LGTM, amazing work!

examples/text-generation/run_generation.py
dacorvo: These are the values used by the neuron compiler. I added the choices though.

examples/text-generation/run_generation.py
```diff
@@ -52,9 +52,15 @@ def generate(model, tokenizer, prompts, length, temperature):
...
+    parser.add_argument(
+        "--batch_size",
+        type=int,
+        default=None,
+        help="The batch size (optional). If not specified it wil be based on the number of prompts.",
+    )
     parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
     parser.add_argument(
-        "--num_cores", type=int, default=2, help="The number of cores on which the model should be split."
+        "--num_cores", type=int, default=1, help="The number of cores on which the model should be split."
```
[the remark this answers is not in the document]
dacorvo: Because I realized that for small models and small batch sizes it is much faster to use a single core than 2.
For instance for GPT2, even with batch size 16 it is still better to use a single core. For a batch size of 128 however, it will be faster if you use two cores.

optimum/neuron/modeling.py
```diff
@@ -732,7 +746,8 @@ def generate_tokens(
 
         """
         # keep track of which sequences are already finished
-        unfinished_sequences = torch.ones(input_ids.shape[0], dtype=torch.long, device=input_ids.device)
+        unfinished_sequences = torch.zeros(input_ids.shape[0], dtype=torch.long, device=input_ids.device)
```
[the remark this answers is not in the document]
dacorvo: it's actually cpu, so not really useful here. It was in the original transformers code for sample, so I left it but I could have removed it.

text-generation-inference/Dockerfile
dacorvo: Bumped version to 1.0.2.

PR #219, opened 2023-09-08
fix(tgi): typer does not support Union types

dacorvo:
In my latest pull-request I accepted a comment to be more accurate with type hinting, but unfortunately typer does not support Union types.


JingyaHuang (approved):
LGTM

PR #223, opened 2023-09-11
Add support for Llama inference through NeuronModelForCausalLM

dacorvo:
This pull-request enables the conversion of llama or llama v2 checkpoints into NeuronModelForCausalLM models.

The converted models can be pushed to the hub as specified in the documentation.

This pull-request also allows to use these models in Inferentia2 TGI containers.

Live demo here:
https://huggingface.co/spaces/dacorvo/inf2-chat

Know issues:
- there is a bug in the serialization that prevents precompiled llama models to be fully reloaded. As a consequence, some parts of the model will be recompiled on every instantiation (i.e. when `from_pretrained` is called),
- there is another bug that leads to corrupted outputs if the model static batch size is > 2.

These two issues should be fixed in an upcoming release of the `transformers-neuronx` library.

EDIT:

I had to add a fix to remove Marian hacks, that had been copied from an earlier version of the `transformers` `GenerationMixin`, and were removed in `4.32.0`: https://github.com/huggingface/transformers/pull/25294.

### DIFF
--- a/optimum/exporters/neuron/model_configs.py
+++ b/optimum/exporters/neuron/model_configs.py
@@ -351,3 +351,9 @@ class VaeDecoderNeuronConfig(VisionNeuronConfig):
 class GPT2NeuronConfig(TextNeuronDecoderConfig):
     NEURONX_ARGS = ["n_positions"]
     NEURONX_CLASS = "gpt2.model.GPT2ForSampling"
+
+
+@register_in_tasks_manager("llama", "text-generation")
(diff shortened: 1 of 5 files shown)


michaelbenayoun (approved):

JingyaHuang (approved):
LGTM

dacorvo:
> Spaces demo leads to error state for me.

Sorry, I had shut it down for the night !

PR #227, opened 2023-09-12
test_examples uses ExampleRunner

michaelbenayoun:
- [x] Updates `ExampleRunner` to make it both usable for Neuron cache filling and testing examples
- [x] Updates the `tests/test_examples.py` file
    * It is possible to run "tiny" versions of the models by setting the following flag `RUN_TINY=1`. It will run the example with the same model except that it contains much less blocks.
    * Each model, for a given task, is tested on different sharding strategies: no sharding, ZeRO-1, TP, TP + ZeRO-1 (when it makes sense)
    * Each model has a defined "coverage" associated to it: low, middle, high. By setting the flag `COVERAGE` it is possible to run test a subset of the models. The coverage of a model is defined by its uniqueness. Ideally running `tests/test_examples.py` with `COVERAGE=high` should cover most of the use cases.
- [x] Checks that the training loss is decreasing
- [x] Test example workflow is now triggered manually 

In following PRs:

- Validation is fixed and performed, each evaluation score is compared to a reference value, either hardcoded manually or computed by running a similar training job with Transformers on GPU.
- Enable ZeRO-1 tests once #222 is merged.

### DIFF
--- a/optimum/neuron/utils/runner.py
+++ b/optimum/neuron/utils/runner.py
@@ -150,7 +150,7 @@ class ExampleRunner:
                 "--target_lang en",
                 "--pad_to_max_length",
                 "--prediction_loss_only",
-                {"t5": "--source_prefix 'Translate Romanian to Enligsh: '"},
+                {"t5": "--source_prefix 'Translate Romanian to English: '"},
             ],
         },
         "image-classification": {
--- a/optimum/neuron/utils/runner.py
+++ b/optimum/neuron/utils/runner.py
@@ -90,6 +90,22 @@ class Precision(str, Enum):
     bf16 = "bf16"
 
 
+def run_command_with_realtime_output(cmd: List[str]) -> Tuple[int, str]:
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    stdout = []
+    decoder = codecs.getincrementaldecoder("utf-8")()
+    while True:
+        output = proc.stdout.read(1)
+        output = decoder.decode(output)
+        if output == "" and proc.poll() is not None:
+            break
+        if output != "":
+            stdout.append(output)
+            print(output, end="")
+    stdout = "".join(stdout)
+    return proc.returncode, stdout
+
+
 class ExampleRunner:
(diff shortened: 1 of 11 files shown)


optimum/neuron/utils/runner.py
```diff
@@ -143,6 +150,7 @@ class ExampleRunner:
                 "--target_lang en",
                 "--pad_to_max_length",
                 "--prediction_loss_only",
+                {"t5": "--source_prefix 'Translate Romanian to Enligsh: '"},
```
dacorvo, line 153:
```suggestion
                {"t5": "--source_prefix 'Translate Romanian to English: '"},
```

tests/test_examples.py
```diff
@@ -41,18 +40,20 @@
...
 # Doing it this way to be able to use this file in tools.
 path_tests = Path(__file__).parent
 sys.path.insert(0, str(path_tests))
-from utils import MODELS_TO_TEST_MAPPING  # noqa: E402
+
+# T =  TypeVar("T")
+# TypeOrDictOfType = Union[T, Dict[str, T]]
+# class TypeOrDictOfType(Generic[T]):  # This is fine
+#     pass
+# TypeOrDictOfType[
```
dacorvo, line 56:
```suggestion
```

dacorvo (changes requested):

tests/test_examples.py
```diff
@@ -61,6 +62,126 @@
...
+        - none: The model cannot be parallelized, either for shape mismatch as in the partial case, or because the
+        tensor parallelism support for this model type has not been added.
+    """
+
+    FULL = "full"
+    PARTIAL = "partial"
+    NONE = "none"
+
+
+class Priority(str, Enum):
```
JingyaHuang, line 84: Isn't it coverage other than priority?

JingyaHuang (approved):
LGTM!! Would be great if we can have all green CIs from now!

dacorvo (approved):
LGTM !

tests/test_examples.py
michaelbenayoun: Renamed to coverage, thanks for the suggestion!

PR #228, opened 2023-09-14
Remove experimental compilation flag for text-generation models

dacorvo:
Using the `--enable-experimental-01` flag speeds up the compilation, but also increases inference latency by 25 to 35 %.


JingyaHuang (approved):
I will remove it from the exporter as well, it looks very risky for the moment.

michaelbenayoun (approved):

PR #229, opened 2023-09-14
Patch for diffusers 0.21.0 release

JingyaHuang:
- [x] Check dependency availability
- [x] Fix import breaks from diffusers 0.21.0
- [x] Remove experimental compilation flag

### DIFF
--- a/optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
+++ b/optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
@@ -28,6 +28,10 @@ if is_diffusers_available():
     from diffusers.pipelines.stable_diffusion import StableDiffusionPipelineOutput
     from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import rescale_noise_cfg
     from diffusers.utils.torch_utils import randn_tensor
+else:
+    raise ImportError(
+        "Stable diffusion requires the diffusers library but it was not found in your environment. You can install it with pip: `pip install diffusers`. Please note that you may need to restart your runtime after installation."
+    )
 
 
 logger = logging.getLogger(__name__)
--- a/optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
+++ b/optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
@@ -18,20 +18,11 @@ import logging
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
-
-from ....utils import is_diffusers_available
-
-
-if is_diffusers_available():
-    from diffusers import StableDiffusionPipeline
-    from diffusers.loaders import LoraLoaderMixin, TextualInversionLoaderMixin
-    from diffusers.pipelines.stable_diffusion import StableDiffusionPipelineOutput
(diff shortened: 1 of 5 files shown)


optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
```diff
@@ -18,11 +18,16 @@
...
+
+from ....utils import is_diffusers_available
+
+
+if is_diffusers_available():
+    from diffusers import StableDiffusionPipeline
+    from diffusers.loaders import LoraLoaderMixin, TextualInversionLoaderMixin
+    from diffusers.pipelines.stable_diffusion import StableDiffusionPipelineOutput
+    from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import rescale_noise_cfg
+    from diffusers.utils.torch_utils import randn_tensor
```
michaelbenayoun, line 30: It will fail if diffusers is not installed no?
Because then we try to inherint from a class that was not imported.

JingyaHuang: The change is to tackle the import issue you met when launching training tests. I would raise an import error is diffusers not installed, wdyt?

JingyaHuang:
Oh right, I forgot that step again... will do @dacorvo !

optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion.py
michaelbenayoun: Where would you raise that?

optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion_xl.py
```diff
@@ -16,11 +16,20 @@
...
+
+
+if is_diffusers_available():
+    from diffusers import StableDiffusionXLPipeline
+    from diffusers.loaders import LoraLoaderMixin, TextualInversionLoaderMixin
+    from diffusers.pipelines.stable_diffusion_xl import StableDiffusionXLPipelineOutput
+    from diffusers.pipelines.stable_diffusion_xl.pipeline_stable_diffusion_xl import rescale_noise_cfg
+    from diffusers.utils.torch_utils import randn_tensor
+else:
+    raise ImportError(
```
JingyaHuang, line 30: here @michaelbenayoun

michaelbenayoun: No we dont want that right? Otherwise reading this file will end-up making everyone fail. Or we should make sure that it is never imported anywhere where `diffusers` is not checked for explicitly.

JingyaHuang: How it end up making __everyone__ fail? It doesn't solve the training test issue though. I will remove changes in pipeline files.
These pipelines are only imported under `modeling_diffusion.py` , will put the check there.

dacorvo (approved):

optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion_xl.py
michaelbenayoun: Alright!

michaelbenayoun (approved):

PR #231, opened 2023-09-14
Using the real model name instead of hard code "model"

davidshtian:
The model config using the hard code "model" as model name, which is not flexible and cannot distinct the real model. The fix used model.name_or_path property for the model name. Thanks~


JingyaHuang (approved):
Hi @davidshtian, thanks for contributing! The PR looks awesome. 

Let me run the tests offline, will merge it if every test goes well.

PR #233, opened 2023-09-15
Sequence parallelism

michaelbenayoun:
- [x] Removes depecrated `orig_to_parallel` parameter in all the tensor parallel related functions and classes.
- [x] Adds support for sequence parallelism for BERT, RoBERTa, GPT-Neo, LLama and T5
- [x] T5: the cross attention is also parallelized under a TP setting.
- [x] Fixes lazy loading with tied weights.

### DIFF
--- a/optimum/neuron/distributed/base.py
+++ b/optimum/neuron/distributed/base.py
@@ -27,6 +27,7 @@ from transformers.utils import WEIGHTS_NAME
 
 from ...utils import logging
 from ..utils import is_neuronx_distributed_available, is_torch_xla_available
+from .parallel_layers import LayerNormSequenceParallelizer
 from .utils import TENSOR_PARALLEL_SHARDS_DIR_NAME, ParameterMetadata, WeightInformation, load_tensor_for_weight
 
 
@@ -62,6 +63,8 @@ class Parallelizer(ABC):
     """
     Base abstract class that handles model parallelism.
     """
+
+    SEQUENCE_PARALLEL_LAYERNORM_PATTERNS: Optional[List[str]] = None
 
     def __init__(self):
         self._validate_required_libaries_are_available()
@@ -115,6 +118,15 @@ class Parallelizer(ABC):
         """
 
     @classmethod
+    def patch_attention_forward_for_sequence_parallelism(
+        cls, model: "PreTrainedModel", sequence_parallel_enabled: bool
+    ):
+        if sequence_parallel_enabled:
+            raise NotImplementedError(
+                f"No patching for the attention mechansim for sequence parallelism was implemented for {model.__class__}"
+            )
+
+    @classmethod
     def parallelize(
         cls,
         model: "PreTrainedModel",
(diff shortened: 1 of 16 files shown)


optimum/neuron/distributed/base.py
```diff
@@ -103,26 +113,35 @@ def _parallelize(
...
 
+    @classmethod
+    def patch_for_sequence_paralelism(cls, model: "PreTrainedModel", sequence_parallel_enabled: bool):
+        """
+        This method needs to be overriden. It must patch anything model-specfic to make the model compatible with
+        sequence parallelism.
+        """
+        if sequence_parallel_enabled:
+            raise NotImplementedError(
+                f"No patching for the attention mechansim for sequence parallelism was implemented for {model.__class__}"
```
dacorvo, line 135:
```suggestion
                f"No patching for the attention mechanism for sequence parallelism was implemented for {model.__class__}"
```

optimum/neuron/distributed/parallel_layers.py
```diff
@@ -705,3 +707,202 @@ def transform(
...
+    Represents a collective op to perform.
+
+    Example:
+
+        ```python
+        info = SequenceCollectiveOpInfo("gather", torch.nn.LayerNorm, "output", "last")
+        ```
+        Here `info` indicates that the output of the last `torch.nn.LayerNorm` must be gathered on the sequence axis.
+
+    """
```
dacorvo, line 724: You should provide a description of the Args matching the type hints below, detailing the allowed values for collective_op, io.

optimum/neuron/distributed/parallel_layers.py
```diff
@@ -705,3 +707,202 @@ def transform(
...
+
+    collective_op: Union[Literal["scatter"], Literal["gather"]]
+    layer: Union[Type["torch.nn.Module"], str]
+    io: Union[Literal["input"], Literal["output"]]
+    first_or_last: Union[Literal["first"], Literal["last"]]
```
dacorvo, line 729: Why not just use a boolean ?

optimum/neuron/utils/require_utils.py
```diff
@@ -15,14 +15,24 @@
...
-from typing import Any, Callable
+from typing import Any, Callable, Dict
 
 from transformers.utils import is_safetensors_available
 
 from . import is_neuronx_distributed_available, is_torch_neuronx_available, is_torch_xla_available
 
 
-def _create_requires_function(availability_function: Callable[[], bool], package_name: str) -> Callable[..., Any]:
+_AVAILIBILITIES: Dict[str, Callable[[], bool]] = {
```
dacorvo, line 25: AVAILABILITIES ?

tests/distributed/model_parallel_test_template.txt
```diff
@@ -95,18 +113,40 @@ for k, v in inputs.items():
...
+            gathered_output.append(gather_output(t, gather_dim))
+        result = output_type(gathered_output)
+    else:
+        result = gather_along_dim(output, gather_dim)
+    return result
+      
+if is_parallel:
+    # Because of parallelism (embeddings and sequence parallelism), some outputs need to be gathered.
+    # Since it is not possible to generically know which one, we save both the "regular" output and the gathered 
+    # version of it. We then compare both of them to the original output and fail both of them do not match.
```
dacorvo, line 140:
```suggestion
    # version of it. We then compare both of them to the original output and fail if both do not match.
```

tests/distributed/model_parallel_test_template.txt
```diff
@@ -95,18 +113,40 @@ for k, v in inputs.items():
...
     gathered_model_outputs =  dict()
     for name, output in model_outputs.items():
-        if name == "loss" or output is None or not isinstance(output, torch.Tensor) or output.shape[-1] != (vocab_size // {tp_size}):
-            gathered_model_outputs[name] = output
+        gathered_model_outputs[name] = output
+        if name == "loss" or output is None: # or not isinstance(output, torch.Tensor) or output.shape[-1] != (vocab_size // {tp_size}):
```
dacorvo, line 144: Is the comment intended ?

tests/distributed/test_model_parallelization.py
```diff
@@ -295,9 +342,45 @@ def test_model_parallel_without_parallelizing_embeddings(
             from_config=False,
             with_lazy_load=True,
             parallelize_embeddings=False,
+            sequence_parallel_enabled=True,
             overwrite_model_config=config_overwrite,
         )
 
+    # TODO: enable that once ParallelCrossEntropy works.
```
dacorvo, line 349: Instead of commenting these tests, you could just have skipped them unconditionally.

dacorvo (changes requested):

optimum/neuron/distributed/parallel_layers.py
michaelbenayoun: Because I find it easier to understand:

```python
SequenceCollectiveOpInfo("scatter", torch.nn.LayerNorm, "input", "first"),
```

Is more explicit than:

```python
SequenceCollectiveOpInfo(True, torch.nn.LayerNorm, False, True),
```

tests/distributed/model_parallel_test_template.txt
```diff
@@ -137,11 +138,11 @@ def gather_output(output, gather_dim):
 if is_parallel:
     # Because of parallelism (embeddings and sequence parallelism), some outputs need to be gathered.
     # Since it is not possible to generically know which one, we save both the "regular" output and the gathered 
-    # version of it. We then compare both of them to the original output and fail both of them do not match.
+    # version of it. We then compare both of them to the original output and fail both do not match.
```
dacorvo, line 141:
```suggestion
    # version of it. We then compare both of them to the original output and fail if both do not match.
```

dacorvo (approved):
Looks goods to me. Thanks again for working on this daunting task.

JingyaHuang (approved):
Don't have to much to say about sequence parallel, the code looks good. Thanks for adding the feature like a rocket!

PR #234, opened 2023-09-18
Replace transformers list of logits warpers by a fused logic warper

dacorvo:
The previous version of the `generate` method was using a list of `transformers` logits warpers to apply temperature and top-k / top-p filtering.

This lead to important slowdowns during inference, as revealed by a comparison between the latency when using `generate` on a llama model and using instead the simpler `sample` method included in `transformers-neuronx`.

A previous pull-request already had replaced the top-k filtering by a faster alternative whenever possible, but this goes one step further, replacing all three warpers by a single fused warper.

The numbers below highlight the benefit of the fused warper:

```
| 1000 tokens on llama2 7B             | top-k only | top-k + top-p |
|--------------------------------------|------------|---------------|
| transformers-neuronx custom sampling | 7.8 s      | 7.8 s         |
| transformers warpers                 | 7.9 s      | 9.7 s         |
| fused warper                         | 7.9 s      | 7.9 s         |
```

### DIFF
--- a/optimum/neuron/generation/__init__.py
+++ b/optimum/neuron/generation/__init__.py
@@ -13,6 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .logits_process import FastTopKLogitsWarper, FusedLogitsWarper
+from .logits_process import FusedLogitsWarper
 from .token_selector import TokenSelector
 from .utils import NeuronGenerationMixin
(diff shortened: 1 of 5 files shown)


michaelbenayoun (approved):
LGTM!

JingyaHuang (approved):
Thanks @dacorvo, the PR looks good.

PR #236, opened 2023-09-20
Use AWS Neuron SDK 2.14

dacorvo:
This bumps the version of the AWS Neuron system and python packages to those corresponding to AWS Neuron SDK 2.14.


JingyaHuang (approved):
LGTM, thanks for the update!

Issue #232, opened 2023-09-15
All 1. weights in `layer_norm` in lazy loaded T5 model


bocchris-aws:
When trying to finetune the google/flan-t5-base model with tensor parallelism using the `run_summarization.py` example, I observed losses that did not match with GPU. The problem is that all `layer_norm` layers get initialized with 1s only and respective weights are not loaded from the checkpoint.

PR #238, opened 2023-09-21
Weight loading after lazy loading fix

michaelbenayoun:
Fixes an important bug: parameters that did not play a part in parallelization for TP were not initialized from checkpoint weights.

Fixes: #232


JingyaHuang (approved):
LGTM

PR #243, opened 2023-09-26
Update text generation example

davidshtian:
Update text generation example default value to utilize all the Neuron cores within the instance, and modify the hard code "gpt2" model to actual model to be compared. Thanks~

### DIFF
--- a/.github/workflows/test_trainium_examples.yml
+++ b/.github/workflows/test_trainium_examples.yml
@@ -4,7 +4,8 @@ on:
   workflow_dispatch:
     inputs: 
       coverage: 
-        description: The coverage of the models to test, useful to perform filtering
+        description: Coverage
+        type: choice
         options:
           - all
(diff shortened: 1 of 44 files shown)


dacorvo (commented):
Thank you for the pull-request. I was actually refactoring this script to separate the conversion from the inference: https://github.com/huggingface/optimum-neuron/pull/245.
I would prefer integrate your changes after that if you don't mind.

dacorvo:
Here is a rebased pull-request with your num-cores change: https://github.com/huggingface/optimum-neuron/pull/247.
Please let me know if that is ok for you to proceed with that pull-request instead of this one.

davidshtian:
> Here is a rebased pull-request with your num-cores change: #247. Please let me know if that is ok for you to proceed with that pull-request instead of this one.

OK. Thanks~

PR #245, opened 2023-09-29
Rework text-generation example

dacorvo:
This modifies the text-generation example to clearly separate compilation from inference using two dedicated actions: `compile` and `run`.

This also bumps the `neuronx` compiler version to the one corresponding to `2.14.1`.


examples/text-generation/generation.py
```diff
@@ -0,0 +1,91 @@
...
+    run_parser.add_argument("--length", type=int, default=128, help="The number of tokens in the generated sequences.")
+    run_parser.add_argument(
+        "--temperature",
+        type=float,
+        default=1.0,
+        help="The temperature to generate. 1.0 has no effect, lower tend toward greedy sampling.",
+    )
+    run_parser.add_argument("--seed", type=int, default=None, help="Pass a seed for reproducibility.")
+    args = parser.parse_args()
+    if args.action == "compile":
```
JingyaHuang, line 66: I wonder if we shall use `export` instead of `compile` here. Back to the moment we chose `export` other than `compile`, and it's kinda nice to keep the example aligned with the API naming.

JingyaHuang (approved):
Looks good, thanks for PR, I just left a small nit.

examples/text-generation/generation.py
dacorvo: I hesitated a lot on that one. Will call it export.

PR #246, opened 2023-09-29
Parallel cross entropy fix

michaelbenayoun:
This PR adds support for parallel cross entropy.

Prior to this PR, the last `ParallelLinear` layer was gathering the output on the TP group, which is costly. 
Now no gathering is done, instead we use a parallel cross entropy that takes care of reducing the loss, which is much more lightweight.


5cp (commented):
In training_args.py, `disable_embedding_parallelization` defaults to True on line 60, and there is a related NotImplementedError() call on line 80 that needs to be removed.

optimum/neuron/trainers.py
```diff
@@ -182,6 +183,8 @@ def __init__(self, *args, **kwargs):
         # Make the model Neuron-compatible for generation.
         patch_generation_mixin_to_neuron_generation_mixin(self.model)
 
+        set_neuron_cc_optlevel_for_model(self.model, optlevel=self.args.neuron_cc_optlevel)
```
dacorvo, line 186: I don't see the benefit of adding yet another parameter that is more or less a duplicate of the compiler optlevel: if the llama models require optlevel 1 because compilation takes too long, then we put it somewhere in the guidelines and ask the user to export it accordingly.
This adds complexity, which will increase the maintenance cost and the test coverage, but it does not fix the real issue (llama compilation takes forever).

optimum/neuron/distributed/base.py
```diff
@@ -293,10 +305,35 @@ def parallelize(
...
+                    # First, we try to initialize the layer similarly as it would be done with the model.
+                    # To do that it is necessary to change the model class to that the `model._init_weights` method
+                    # considers this module as a `torch.nn.Linear` instance.
+                    orig_class = mod.__class__
+                    mod.__class__ = torch.nn.Linear
+                    left_uninitialized = try_to_hf_initialize(model, mod, parameter_names)
+                    mod.__class__ = orig_class
+                    if not left_uninitialized:
+                        continue
+                    initialize_parallel_linear(mod, left_uninitialized)
```
dacorvo, line 334: I might miss something, but this seems a bit convoluted just to set the weights to a normal distribution and the bias to zero (which is exactly what the HF init does for all Linear layers).

dacorvo (commented):
I fail to see what change is exactly related to the parallel cross entropy fix. Could you point to the relevant changed lines in that pull-request ?

I also added two comments on some of the changes. As a general comment, I think the training code is really complex, and I have sometimes difficulties understanding why: I understand the goal is to provide the same user experience as transformers by hiding most of the differences to the end-user (at the cost of code complexity), but there are actually hidden subtleties that defeat the purpose.

optimum/neuron/distributed/base.py
michaelbenayoun: So we are not sure it is always the case, and it is not, for instance in T5. 
Basically it tries to use the initialization defined by the model if possible, and use the default initialization scheme otherwise.

optimum/neuron/trainers.py
michaelbenayoun: The idea is to include this "knowledge" so that the user does not have to care about that.
For instance here: #313 .

I do not think this will add too much maintenance code, it is basically a function that sets a flag in the compiler. 

I think it's a nicer API for the end user to be able to specify the opt level, if they want to do that, as an argument instead of using the environment variable.

michaelbenayoun:
It did not work before for multiple reasons:

- Some model were not properly parallelized. For that you can check changes that were made in the `encoder_models.py`, `decoder_models.py` and `encoder_decoder_models.py` files
- **Minor**: doing an equivalent initialization between lazy loading and classic loading is not so easy, and this PR improves that a lot. You can check the `distributed/base.py` file for that
- Finally there is a bug from `torch_xla` or the Neuron compiler when trying to generate a tensor of random integers... So I had to fix the way we were generating dummy labels, you can check the `tests/distributed/utils.py` file for that

optimum/neuron/trainers.py
dacorvo: I think it is a bit superfluous, but agree to disagree.

optimum/neuron/distributed/parallel_layers.py
```diff
@@ -137,6 +146,21 @@ class ParallelEmbedding(ParallelLayer):
...
+
+        tp_size = get_tensor_model_parallel_size()
+        if cls.VOCAB_SIZE_NAME is not None:
+            attribute_names = cls.VOCAB_SIZE_NAME.split(".")
+            obj = layer
+            for name in attribute_names[:-1]:
+                obj = getattr(obj, name)
+            vocab_size_attribute_name = attribute_names[-1]
+            new_vocab_size = getattr(obj, vocab_size_attribute_name) // tp_size
+            setattr(obj, vocab_size_attribute_name, new_vocab_size)
```
dacorvo, line 162: OK, I see the change now, thanks.

dacorvo (approved):
Thanks for taking the time to fixing these tests: this looks like a daunting task.
I think I grasp the spirit of your changes, although I still don't get the full picture, but it is ok: I am trusting you on this.

optimum/neuron/distributed/utils.py
```diff
@@ -466,6 +473,65 @@ def gqa_key_value_slicing_when_tp_size_greater_than_num_key_value_heads(
...
+    """
+    if "weight" in parameter_names:
+        delete_tensor_model_parallel_attributes(mod.weight)
+        # It is needed to use `init_weight_cpu` instead of `_init_weights` because the initialization
+        # needs to happen on the full parameter and then scatter it accross TP ranks otherwise it will
+        # not be equivalent to the non-parallel case.
+        mod.init_weight_cpu()
+    if mod.bias is not None and "bias" in parameter_names:
+        mod._init_bias()
+        # mod.bias.data.zero_()
```
JingyaHuang, line 532: To delete?

JingyaHuang (approved):
LGTM. Thanks Michael.

optimum/neuron/distributed/utils.py
michaelbenayoun: Done.

Issue #253, opened 2023-10-06
Loss curve mismatch with tensor parallelism


bocchris-aws:
Running the summarization script on `main` in the following way leads to loss curve discrepancies when compared to GPU.

```
MODEL=google/flan-t5-xl
NUM_WORKERS=32
TRAIN_BS=1
EVAL_BS=1
SRC_LENGTH=512
TGT_LENGTH=64
GRAD_ACCUM=64

TP=8

export MALLOC_ARENA_MAX=64
export XLA_DOWNCAST_BF16=1

torchrun --nproc_per_node=$NUM_WORKERS --nnodes=1 \
     run_summarization.py \
    --model_name_or_path $MODEL \
    --dataset_name cnn_dailymail \
    --dataset_config "3.0.0" \
    --do_train \
    --max_steps 5000 \
    --save_strategy no \
    --do_eval \
    --max_eval_samples 64 \
    --max_source_length $SRC_LENGTH \
    --max_target_length $TGT_LENGTH \
    --val_max_target_length $TGT_LENGTH \
    --per_device_train_batch_size $TRAIN_BS \
    --per_device_eval_batch_size $EVAL_BS \
    --pad_to_max_length \
    --gradient_accumulation_steps $GRAD_ACCUM \
    --overwrite_output_dir \
    --seed 42 \
    --weight_decay 0.0 \
    --logging_steps 5 \
    --bf16 \
    --zero_1 \
    --logging_first_step true \
    --source_prefix "summarize: " \
    --tensor_parallel_size $TP
```

![Screenshot 2023-10-06 at 16 51 22](https://github.com/huggingface/optimum-neuron/assets/132611342/9a5996cf-d3d6-4ce0-a09b-9a7294e45807)

bocchris-aws:
Based on our discussion and the below screenshot, I will reopen this ticket for now. 
![Screenshot 2023-10-30 at 11 18 07](https://github.com/huggingface/optimum-neuron/assets/132611342/4609dc60-4c24-436c-aec1-de88d72e8403)

PR #257, opened 2023-10-11
Fix noisy loss for T5 when doing TP

michaelbenayoun:
Fixes: #253


bocchris-aws:
Problem of noisy loss remains after the fix as shown in Figure.
Screenshot 2023-10-16 at 11 04 18

PR #264, opened 2023-10-19
Use attention masks for TGI generation

dacorvo:
Attention masks were simply ignored during generation, leading to gibberish output if the inputs of some slots were padded (which typically happens when adding a new request during generation).


JingyaHuang (approved):
LGTM, thanks for the fix.

PR #265, opened 2023-10-19
Fix neuron pipelines

dacorvo:
This fixes a few issues with the neuron pipelines:

- not passing `input_shapes` to the `pipeline()` method with `export=False` led to errors if the model was not a model instance,
- the model static `batch_size` was never passed to the pipeline, leading to all inputs being processed individually,
- inconsistent parameters to `pipeline()` were not correctly checked,
- `hub_kwargs` was never passed to `from_pretrained()`.

Several tests are added for decoder models to check all features are correctly implemented. 

I also renamed `model_kwargs` to `compiler_args` to match the names used in other places.

Finally, I fixed the long-standing `load_tensor_for_weight` failure, that could have allowed me to get my first green CI, but alas there is an issue with the trainium test cache on the hub.

### DIFF
--- a/optimum/neuron/pipelines/transformers/base.py
+++ b/optimum/neuron/pipelines/transformers/base.py
@@ -158,7 +158,7 @@ def pipeline(
     feature_extractor: Optional[Union[str, PreTrainedFeatureExtractor]] = None,
     use_fast: bool = True,
     export: bool = False,
-    input_shapes: Optional[Dict[str, int]] = None,
+    input_shapes: Optional[Dict[str, int]] = {},
     token: Optional[Union[str, bool]] = None,
     revision: Optional[str] = None,
     trust_remote_code: Optional[bool] = None,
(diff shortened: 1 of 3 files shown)


optimum/neuron/pipelines/transformers/base.py
```diff
@@ -102,25 +104,24 @@ def load_pipeline(
     subfolder: str = "",
     token: Optional[Union[bool, str]] = None,
     revision: str = "main",
-    model_kwargs: Optional[Dict[str, Any]] = None,
+    compiler_args: Optional[Dict[str, Any]] = {},
     config: AutoConfig = None,
+    hub_kwargs: Optional[Dict[str, Any]] = {},
```
michaelbenayoun, line 109: It shoud be left to `None` and set inside the function instead.

optimum/neuron/pipelines/transformers/base.py
```diff
@@ -158,11 +159,11 @@ def pipeline(
     feature_extractor: Optional[Union[str, PreTrainedFeatureExtractor]] = None,
     use_fast: bool = True,
     export: bool = False,
-    input_shapes: Optional[Dict[str, int]] = None,
+    input_shapes: Optional[Dict[str, int]] = {},
+    compiler_args: Optional[Dict[str, int]] = {},
```
michaelbenayoun, line 163: Same comment.

michaelbenayoun (approved):

Issue #249, opened 2023-10-04
Tensor parallelism: saved model can't be loaded


bocchris-aws:
After running finetuning using `run_summarization.py` for a couple of steps, the directory to which the model shards have been saved does not contain a `config.json`. Hence, the model stored can't be loaded using `from_pretrained()`:`
```
OSError: PATH_TO_SAVED_MODEL_DIR does not appear to have a file named config.json.
```
Should the `config.json` be saved together with the shards?

michaelbenayoun:
You need to provide the parent directory instead. The parent directory contains all the config files.

PR #269, opened 2023-10-24
Save / load from checkpoint TP

michaelbenayoun:
As per title.

Fixes: #249

### DIFF
--- a/optimum/neuron/distributed/base.py
+++ b/optimum/neuron/distributed/base.py
@@ -28,8 +28,8 @@ from transformers.utils import WEIGHTS_NAME
 
 from ...utils import logging
 from ..utils import is_neuronx_distributed_available, is_torch_xla_available
-from ..utils.require_utils import requires_neuronx_distributed, requires_torch_xla
 from ..utils.deprecate_utils import deprecate
+from ..utils.require_utils import requires_neuronx_distributed
 from .parallel_layers import (
     IOSequenceParallelizer,
     LayerNormSequenceParallelizer,
@@ -312,6 +312,7 @@ class Parallelizer(ABC):
     @requires_neuronx_distributed
     def was_parallelized(cls, model: "PreTrainedModel") -> bool:
         from neuronx_distributed import parallel_layers
+
         parallel_layer_classes = (
             parallel_layers.ParallelEmbedding,
             parallel_layers.ColumnParallelLinear,
@@ -425,9 +426,12 @@ class Parallelizer(ABC):
     ):
         cls._check_model_was_parallelized(model)
 
-        import torch_xla.core.xla_model as xm
         import neuronx_distributed
-        from neuronx_distributed.parallel_layers.parallel_state import  get_data_parallel_rank, get_tensor_model_parallel_rank
+        import torch_xla.core.xla_model as xm
+        from neuronx_distributed.parallel_layers.parallel_state import (
+            get_data_parallel_rank,
+            get_tensor_model_parallel_rank,
+        )
 
         data_parallel_rank = get_data_parallel_rank()
         tensor_parallel_rank = get_tensor_model_parallel_rank()
@@ -527,6 +531,7 @@ class Parallelizer(ABC):
     def load_model_sharded_checkpoint(cls, model: "PreTrainedModel", load_dir: Union[str, Path]):
         cls._check_model_was_parallelized(model)
         from neuronx_distributed import parallel_layers
+
         if not isinstance(load_dir, Path):
             load_dir = Path(load_dir)
         parallel_layers.load(load_dir / TENSOR_PARALLEL_SHARDS_DIR_NAME, model=model, sharded=True)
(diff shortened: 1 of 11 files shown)


optimum/neuron/distributed/base.py
```diff
@@ -525,7 +529,48 @@ def load_model_checkpoint(cls, model: "PreTrainedModel", load_dir: Union[str, Pa
...
+            raise FileNotFoundError(f"Could not find a sharded checkpoint directory under {load_dir.as_posix()}.")
+
+    @classmethod
+    def load_optimizer_sharded_checkpoint(cls, optimizer: "torch.optim.Optimizer", load_dir: Union[str, Path]):
+        from neuronx_distributed.optimizer import NeuronZero1Optimizer
+
+        is_zero_1_optimizer = optimizer.__class__.__name__ == "NeuronAcceleratedOptimizer" and isinstance(optimizer.optimizer, NeuronZero1Optimizer)
+        is_zero_1_optimizer = is_zero_1_optimizer or isinstance(optimizer, NeuronZero1Optimizer)
+        if is_zero_1_optimizer:
+            raise NotImplementedError("It is not to load a sharded optimizer checkpoint when using ZeRO-1 yet.")
```
dacorvo, line 542:
```suggestion
            raise NotImplementedError("It is not possible to load a sharded optimizer checkpoint when using ZeRO-1 yet.")
```

optimum/neuron/distributed/base.py
```diff
@@ -525,7 +529,48 @@ def load_model_checkpoint(cls, model: "PreTrainedModel", load_dir: Union[str, Pa
...
             cls.load_model_sharded_checkpoint(model, load_dir)
-        elif (load_dir / WEIGHTS_NAME).is_file():
-            cls.load_model_regular_checkpoint(model, load_dir)
         else:
-            raise FileNotFoundError(f"Could not find a checkpoint file under {load_dir.as_posix()}.")
+            raise FileNotFoundError(f"Could not find a sharded checkpoint directory under {load_dir.as_posix()}.")
+
+    @classmethod
+    def load_optimizer_sharded_checkpoint(cls, optimizer: "torch.optim.Optimizer", load_dir: Union[str, Path]):
+        from neuronx_distributed.optimizer import NeuronZero1Optimizer
```
dacorvo, line 537: Any particular reason why the import happens here ? The method is not guarded by `@requires_neuronx_distributed`, and you have a check at the top of the file where you already import `parallel_layers`

optimum/neuron/distributed/base.py
```diff
@@ -525,7 +529,48 @@ def load_model_checkpoint(cls, model: "PreTrainedModel", load_dir: Union[str, Pa
...
+
+        is_zero_1_optimizer = optimizer.__class__.__name__ == "NeuronAcceleratedOptimizer" and isinstance(optimizer.optimizer, NeuronZero1Optimizer)
+        is_zero_1_optimizer = is_zero_1_optimizer or isinstance(optimizer, NeuronZero1Optimizer)
```
dacorvo, line 540: Isn't the check in the first line equivalent to `isinstance(optimizer, NeuronAcceleratedOptimizer)` ?
Anyway, since in the second line you also accept just `NeuronZero1Optimizer`, maybe it could be simplified to:
`is_zero_1_optimizer = isinstance(optimizer, NeuronZero1Optimizer`)

optimum/neuron/distributed/base.py
```diff
@@ -525,7 +529,48 @@ def load_model_checkpoint(cls, model: "PreTrainedModel", load_dir: Union[str, Pa
...
+        if is_zero_1_optimizer:
+            raise NotImplementedError("It is not to load a sharded optimizer checkpoint when using ZeRO-1 yet.")
+
+        if not isinstance(load_dir, Path):
+            load_dir = Path(load_dir)
+
+        from neuronx_distributed.parallel_layers.parallel_state import (
```
dacorvo, line 547: You already imported parallel_layers at the top of the file, so maybe you can import more symbols. Again, maybe I am missing something here ...

optimum/neuron/trainers.py
```diff
@@ -372,38 +373,82 @@ def _maybe_log_save_evaluate(self, tr_loss, model, trial, epoch, ignore_keys_for
...
+        self._save_xla(output_dir)
+
+        # Push to the Hub when `save_model` is called by the user.
+        if self.args.push_to_hub and not _internal_call:
+            self.push_to_hub(commit_message="Model save")
+
+    def _save_checkpoint(self, model, trial, metrics=None):
+        # In all cases, including ddp/dp/deepspeed, self.model is always a reference to the model we
+        # want to save except FullyShardedDDP.
+        # assert unwrap_model(model) is self.model, "internal model should be a reference to self.model"
```
dacorvo, line 428: Is the check needed or the comment above is enough ?

dacorvo (commented):
Thanks for addressing the specified issue. Wouldn't it be better to have a test verifying that the issue is solved ?

I have a few comments, although I am not able to grasp the rationale behind all the changes that come on top of code I am not familiar with.

optimum/neuron/distributed/base.py
michaelbenayoun: There are two ways of doing this;

1. Importing at the top of the file after having checked that the package was available
2. Do it on a per-function basis with the requires decorator

I was doing kind of both in the same file which is not great you're right, fixed it so follow the second way!

optimum/neuron/distributed/base.py
michaelbenayoun: No because you can have:

1. A `NeuronZero1Optimizer`, which we do not support yet
2. A `NeuronAcceleratedOptimizer` which is a wrapper around the optimizer that is always applied when using `accelerate` (I think we will always fall in that case but I wanted to cover all the cases here). This wrapper contains an `optimizer` attribute, and we check if it's a `NeuronZero1Optimizer`.

optimum/neuron/trainers.py
michaelbenayoun: Basically copied from `transformers`.

tests/distributed/training.py
```diff
@@ -0,0 +1,96 @@
+# coding=utf-8
```
dacorvo, line 1: Unless I am mistaken:
- this test will not be picked up by `pytest` automatically when parsing the `tests/distributed` directory because the file name does not start with `test_` (see https://docs.pytest.org/en/7.1.x/getting-started.html#run-multiple-tests).
- more, it does not have the `@is_trainium_test` decorator.
 
I think it will therefore be ignored by the CI, and it seems to be confirmed in the CI logs.

dacorvo (commented):
Thanks for the update and providing a test. However, I think this test is not picked up by te CI: is this intended ?

tests/distributed/training.py
michaelbenayoun: You are right!
I'm not done yet. Will finish now.

michaelbenayoun: Done.

dacorvo (approved):

dacorvo (changes requested):
CI is failing: FAILED tests/distributed/test_training.py::test_tp_save_and_resume_from_checkpoint - RuntimeError: You need to log in the Hugging Face Hub otherwise you will not be able to push anything.

dacorvo (approved):

dacorvo:
Thank you for fixing this !

Issue #272, opened 2023-10-24
tests/test_examples.py and tests/distributed failures with neuron 2.15


dtsengAmazon:
I have been running 2 tests with neuron 2.15 pre-release and here are the results:
This coverage test "`RUN_SLOW=true COVERAGE=high RUN_TINY=true USE_VENV=false pytest tests/test_examples.py -v`" results in "**10 failed, 9 passed in 7126.60s (1:58:46)**" 
and this distributed test "`pytest tests/distributed -v`" results in  **1 failed, 77 passed, 48 skipped in 3088.71s (0:51:28)**.

Through these tests, I have identified 3 groups of failures:

1. The 2 Question-Answering tests in the coverage test error out with ValueError: You must provide sequence_length for task question-answering. Which can likely be fixed with passing in the correct parameter sequence_length
2. All of the tp tests in the coverage test fail with errors similar to the following with respective model names:                                                                                                                                                                                                         
```
  File "/home/ubuntu/dtseng/optimum-neuron/examples/language-modeling/run_clm.py", line 639, in <module>                                                                                                                                                    
    main()                                                                                                                                                                                                                                                  
  File "/home/ubuntu/dtseng/optimum-neuron/examples/language-modeling/run_clm.py", line 414, in main                                                                                                                                                        
    model = AutoModelForCausalLM.from_pretrained(                                                                                                                                                                                                           
  File "/home/ubuntu/aws_neuron_venv_pytorch_p38/lib/python3.8/site-packages/transformers/models/auto/auto_factory.py", line 565, in from_pretrained                                                                                                        
    return model_class.from_pretrained(                                                                                                                                                                                                                     
  File "/home/ubuntu/aws_neuron_venv_pytorch_p38/lib/python3.8/site-packages/optimum/neuron/distributed/utils.py", line 512, in from_pretrained_for_tp                                                                                                      
    filenames, sharded_metadata = download_checkpoints_in_cache(                                                                                                                                                                                            
  File "/home/ubuntu/aws_neuron_venv_pytorch_p38/lib/python3.8/site-packages/optimum/neuron/utils/misc.py", line 287, in download_checkpoints_in_cache                                                                                                      
    raise EnvironmentError(                                                                                                                                                                                                                                 
OSError: Error no file named pytorch_model.bin, tf_model.h5, model.ckpt.index or flax_model.msgpack found in directory /tmp/tmp57f8xkdb/gpt-neo-125M. 
```
3. The single error in the distributed test is:
[...]

PR #274, opened 2023-10-25
Fix broken tests

michaelbenayoun:
Fixes some broken tests.

Fixes: #272


dacorvo:
Rebased on main to use the latest AWS Neuron SDK.

michaelbenayoun:
cc @Wauplin

.github/workflows/test_trainium_common.yml
```diff
@@ -16,45 +16,9 @@ concurrency:
...
-          iam-role-name: ${{ env.EC2_IAM_ROLE }}
-          aws-resource-tags: > # optional, requires additional permissions
-            [
-              {"Key": "Name", "Value": "ec2-optimum-github-runner"},
-              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
-            ]
   optimum-neuron-tests:
-    needs: start-runner # required to start the main job when the runner is ready
-    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    # needs: start-runner # required to start the main job when the runner is ready
```
dacorvo: We can remove that line now.

optimum/neuron/utils/cache_utils.py
```diff
@@ -719,7 +717,10 @@ def get_cached_model_on_the_hub(neuron_hash: NeuronHash) -> Optional[CachedModel
             repo_id, revision = repo_id
         else:
             revision = "main"
-        repo_filenames = HfApi().list_repo_files(repo_id, revision=revision, token=HfFolder.get_token())
+        try:
+            repo_filenames = HfApi().list_repo_files(repo_id, revision=revision, token=HfFolder.get_token())
+        except Exception:
+            continue
```
dacorvo, line 723: Do you really need this or was it required while you were investigating issues ?

tests/test_cache_utils.py
```diff
@@ -558,6 +547,7 @@ def test_push_to_hub_overwrite_existing(self):
             # With a directory
             with self.assertLogs("optimum", level="INFO") as cm:
                 push_to_cache_on_hub(neuron_hash, cache_dir, self.CUSTOM_PRIVATE_CACHE_REPO)
+                print(cm.output)
```
dacorvo, line 550: In similar tests you removed the print instruction. Do you still need it here ? Maybe it is because in other places you have inserted a mark_step ?

dacorvo (approved):
Thanks you very much ! LGTM, except a few nits.

.github/workflows/test_trainium_common.yml
michaelbenayoun: Removed in the last commit.

optimum/neuron/utils/cache_utils.py
michaelbenayoun: It is needed. 
Basically we are trying to list the files in the `repo_id`. But the `repo` might be private to us, which would lead to an error. 
Basically it is enforced when setting the main cache repo that it is not private to the user, but others might be (usually when playing around the cache repo during testing).

tests/test_cache_utils.py
michaelbenayoun: I think I leave this mostly for debugging purposes when I want to see the actual log output (which was captured by the context-manager if I recall properly). 

What you are referring to is different: basically before we were using `print` to force graph compilation. A print being equivalent to a `xm.mark_step()` + a computation of the resulting graph + transfert of the output to the CPU. Since the last compiler update it seems to not be working in this specific case, so just replaced it with `xm.mark_steps()`.

Issue #271, opened 2023-10-24
New instance ignores aws-neuron/optimum-neuron-cache and still compiles model


kct22aws:
By using `CUSTOM_CACHE_REPO="aws-neuron/optimum-neuron-cache" torchrun.....` as in the [image classification example](https://github.com/huggingface/optimum-neuron/blob/main/examples/image-classification/run_image_classification.py), I can push model to the repo. However, when in another instance, when I ran the same command and same training script as shown, it still proceed to compile the model. It seems Trainer is ignoring to look at the cache repo.

michaelbenayoun:
Alright, as a quick fix: can you try setting the cache repo with the CLI:

```
optimum-cli neuron cache set aws-neuron/optimum-neuron-cache
```

Also, are you logged in? If you're not logged in or do not have writing rights on this repo it will not push anything.

5cp:
I was able to reproduce the issue with optimum-neuron 0.0.12 and Neuron 2.14.

The ON ViT training job finds and attempts to download the files from the hub cache, but there are already 2 graphs that have triggered JiT compilation at that time:

```
INFO|trainer.py:1760] 2023-10-31 00:25:19,639 >> ***** Running training *****
[INFO|trainer.py:1761] 2023-10-31 00:25:19,639 >>   Num examples = 75,760
[INFO|trainer.py:1762] 2023-10-31 00:25:19,639 >>   Num Epochs = 1
[INFO|trainer.py:1763] 2023-10-31 00:25:19,639 >>   Instantaneous batch size per device = 16
[INFO|trainer.py:1766] 2023-10-31 00:25:19,639 >>   Total train batch size (w. parallel, distributed & accumulation) = 32
[INFO|trainer.py:1767] 2023-10-31 00:25:19,639 >>   Gradient Accumulation steps = 1
[INFO|trainer.py:1768] 2023-10-31 00:25:19,639 >>   Total optimization steps = 4,735
[INFO|trainer.py:1769] 2023-10-31 00:25:19,639 >>   Number of trainable parameters = 85,876,325
^M  0%|          | 0/4735 [00:00<?, ?it/s]
2023-10-31 00:25:19.000786:  45883  INFO ||NEURON_CACHE||: Compile cache path: /tmp/tmp7428jxhm/neuron-compile-cache
[...]

PR #280, opened 2023-10-31
Fix neuron cache starting compilation before fetching

michaelbenayoun:
This PR applies the [suggested fix ](https://github.com/huggingface/optimum-neuron/issues/271#issuecomment-1787748978) by @5cp to prevent compilation to start before finishing fetching the model compilation files from the Neuron cache.


kct22aws:
This fix will resolve issue as reported in https://github.com/huggingface/optimum-neuron/issues/271

PR #282, opened 2023-11-02
Fix doc build source code broken links

JingyaHuang:


regisss:
@JingyaHuang You may need to do it for PR doc builds also: https://github.com/huggingface/optimum-neuron/blob/a837ac2afee6c82b3795b40ade1b8bfde7639a9f/.github/workflows/doc-pr-build.yml#L39

PR #292, opened 2023-11-03
Add revision to push_to_hub

philschmid:
This PR adds the revision key to the `push_to_hub` method to push models, to specific revision, this is helpful ,e.g. to version compiled checkpoints, e.g. 2.15.0 revision.

Also removes/fixes fairscale change https://github.com/huggingface/transformers/pull/25702

### DIFF
--- a/optimum/neuron/modeling_decoder.py
+++ b/optimum/neuron/modeling_decoder.py
@@ -259,6 +259,7 @@ class NeuronDecoderModel(OptimizedModel):
         save_directory: str,
         repository_id: str,
         private: Optional[bool] = None,
+        revision: Optional[str] = None,
         use_auth_token: Union[bool, str] = True,
         endpoint: Optional[str] = None,
     ) -> str:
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling_base.py
```diff
@@ -317,6 +317,7 @@ def push_to_hub(
         save_directory: str,
         repository_id: str,
         private: Optional[bool] = None,
+        revision: Optional[str] = "main",
```
JingyaHuang, line 320: Can it be `None`?

philschmid: yes

dacorvo:
Thanks for this pull-request.
Regarding the addition of revision, I think it was omitted because the original method in `optimum` does not have it.
For consistency, it would be nice to add it also here: https://github.com/huggingface/optimum-neuron/blob/1d2a95e80b995a44a75aebdfd5b9aec19e1d79c6/optimum/neuron/modeling_decoder.py#L257.

michaelbenayoun (approved):

philschmid:
@dacorvo added it for the decoder as well.

PR #316, opened 2023-11-10
Transformers fix and doc update

dacorvo:
This modifies the serialization of `transformers-neuronx` checkpoints to avoid using `safetensors` (default option in `transformers >= 4.35.0`). To avoid further breaking changes, the `transformers` version is now pinned.

This also modifies the `NeuronModelForCausalLM.push_to_hub` method to avoid commiting each checkpoint file individually (each file is still uploaded individually though).

This also adds two new tutorials based on the `llama2 13B` notebook and the TGI README.

Finally, this sets two options in `NeuronModelForCausalLM.from_pretrained` to limit CPU memory usage.

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -10,12 +10,16 @@
       title: Overview
     - local: tutorials/fine_tune_bert
       title: Fine-tune BERT for Text Classification on AWS Trainium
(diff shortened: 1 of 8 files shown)


dacorvo:
Closing and reopening multiple PRs  as changes are accumulating and it will be easier to merge them individually.

PR #319, opened 2023-11-13
fix(pipelines): support passing decoder model + tokenizer

dacorvo:
When passing explicitly a neuron model to a pipeline, we check the model class. This modifies the check to accept not only NeuronBaseModel but also NeuronModelForCausalLM.

Update: cherry-picked @glegendre01 modifications to github workflows to reactivate INF1 CI.

### DIFF
--- a/optimum/neuron/pipelines/transformers/base.py
+++ b/optimum/neuron/pipelines/transformers/base.py
@@ -123,7 +123,7 @@ def load_pipeline(
             model, export=export, **compiler_args, **input_shapes, **hub_kwargs, **kwargs
         )
     # uses neuron model
-    elif isinstance(model, NeuronBaseModel):
+    elif isinstance(model, (NeuronBaseModel, NeuronModelForCausalLM)):
         if tokenizer is None and load_tokenizer:
             for preprocessor in model.preprocessors:
                 if isinstance(preprocessor, (PreTrainedTokenizer, PreTrainedTokenizerFast)):
--- a/.github/workflows/test_inf1.yml
+++ b/.github/workflows/test_inf1.yml
@@ -15,65 +15,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  start-runner:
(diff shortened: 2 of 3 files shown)


optimum/neuron/pipelines/transformers/base.py
```diff
@@ -123,7 +123,7 @@ def load_pipeline(
             model, export=export, **compiler_args, **input_shapes, **hub_kwargs, **kwargs
         )
     # uses neuron model
-    elif isinstance(model, NeuronBaseModel):
+    elif isinstance(model, (NeuronBaseModel, NeuronModelForCausalLM)):
```
michaelbenayoun, line 126: QQ: why is `NeuronModelForCausalLM` not a sublass of `NeuronBseModel`?

dacorvo: Because `NeuronBaseModel` is based on JIT models and implements the corresponding conversion logic. `NeuronModelForCausalLM` is a subclass of  `NeuronDecoderModel` that uses `transformers-neuronx` models instead.
We could refactor to add a common class to both though, since the latest subclasses of `NeuronBaseModel` are overriding pretty much all its methods now.

JingyaHuang: Wouldn't it make more sense to use `NeuronDecoderModel` here?

JingyaHuang (approved):
LGTM, thanks for the fix!

optimum/neuron/pipelines/transformers/base.py
michaelbenayoun: I guess it would be good to refactor it if possible. But also it's not pressing and can be postponed to when we have more bandwidth!

dacorvo: I will wait until @JingyaHuang 's pull-request on T5 is merged as it is a new subclass, and any change I make before that to the base class will result in nightmarish conflicts.

PR #329, opened 2023-11-16
Save function fix

michaelbenayoun:
This PR aims at fixing the save function. 

Since `transformers==4.35.0` the default saving function being used is the `safetensors.torch.save_file` one. This PR patches the saving mechanism to account for that.

### DIFF
--- a/optimum/neuron/trainers.py
+++ b/optimum/neuron/trainers.py
@@ -48,6 +48,7 @@ from .trainer_callback import NeuronCacheCallback
 from .utils import (
     DynamicPatch,
     ModelPatcher,
+    Patcher,
     is_torch_xla_available,
     patch_within_function,
 )
@@ -61,6 +62,7 @@ from .utils.training_utils import (
     patched_finfo,
     prepare_environment_for_neuron,
     skip_first_batches,
+    torch_xla_safe_save_file,
 )
 
 
@@ -392,20 +394,27 @@ class AugmentTrainerForNeuronMixin:
             xm.mark_step()
             parallelizer.save_model_checkpoint(self.model, output_dir, as_sharded=True, optimizer=self.optimizer)
         else:
+            safe_save_function_patcher = Patcher(
+                [("transformers.modeling_utils.safe_save_file", torch_xla_safe_save_file)]
+            )
             if not isinstance(self.model, PreTrainedModel):
                 if isinstance(unwrap_model(self.model), PreTrainedModel):
-                    unwrap_model(self.model).save_pretrained(
-                        output_dir,
-                        is_main_process=self.args.should_save,
-                        state_dict=self.model.state_dict(),
-                        save_function=xm.save,
-                    )
+                    with safe_save_function_patcher:
+                        unwrap_model(self.model).save_pretrained(
+                            output_dir,
+                            is_main_process=self.args.should_save,
+                            state_dict=self.model.state_dict(),
+                            save_function=xm.save,
+                        )
                 else:
                     logger.info("Trainer.model is not a `PreTrainedModel`, only saving its state dict.")
                     state_dict = self.model.state_dict()
                     xm.save(state_dict, os.path.join(output_dir, WEIGHTS_NAME))
             else:
-                self.model.save_pretrained(output_dir, is_main_process=self.args.should_save, save_function=xm.save)
+                with safe_save_function_patcher:
+                    self.model.save_pretrained(
+                        output_dir, is_main_process=self.args.should_save, save_function=xm.save
+                    )
 
         if self.tokenizer is not None and self.args.should_save:
             self.tokenizer.save_pretrained(output_dir)
--- a/optimum/neuron/utils/cache_utils.py
+++ b/optimum/neuron/utils/cache_utils.py
@@ -133,6 +133,7 @@ def create_custom_cache_repo(repo_id: str = CACHE_REPO_NAME, private: bool = Tru
 
 
 def is_private_repo(repo_id: str) -> bool:
+    """Tells whether `repo_id` is private to the current user logged-in."""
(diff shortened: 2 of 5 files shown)


optimum/neuron/trainers.py
```diff
@@ -402,20 +404,27 @@ def _save_xla(self, output_dir: Optional[str] = None):
...
+                            state_dict=self.model.state_dict(),
+                            save_function=xm.save,
+                        )
                 else:
                     logger.info("Trainer.model is not a `PreTrainedModel`, only saving its state dict.")
                     state_dict = self.model.state_dict()
                     xm.save(state_dict, os.path.join(output_dir, WEIGHTS_NAME))
             else:
-                self.model.save_pretrained(output_dir, is_main_process=self.args.should_save, save_function=xm.save)
+                with safe_save_function_patcher:
```
JingyaHuang, line 424: What is `self.model` under this case? If `unwrap_model(self.model)` is not an instance of `PreTrainedModel`, can it still apply `save_pretrained()`?

optimum/neuron/utils/cache_utils.py
```diff
@@ -133,14 +133,22 @@ def create_custom_cache_repo(repo_id: str = CACHE_REPO_NAME, private: bool = Tru
 
 
 def is_private_repo(repo_id: str) -> bool:
+    """Tells whether `repo_id` is private."""
```
JingyaHuang, line 136: Does this function check whether a repo is private to all public? If so why do we need to check if it's private to a particular user in advance?

optimum/neuron/utils/cache_utils.py
```diff
@@ -970,10 +993,7 @@ def push_to_cache_on_hub(
                 f"{local_cache_dir_or_file}"
             )
 
-    could_not_push_message = (
-        "Could not push the cached model to the repo {cache_repo_id}, most likely due to not having the write permission "
-        "for this repo. Exact error:\n{error}."
-    )
+    could_not_push_message = "Could not push the cached model to the repo {cache_repo_id}. Exact error:\n{error}."
```
JingyaHuang, line 996:
```suggestion
    could_not_push_message = "Could not push the cached model to the repo {cache_repo_id}. Error message:\n{error}."
```

optimum/neuron/utils/training_utils.py
```diff
@@ -315,6 +315,28 @@ def skip_first_batches(dataloader, num_batches=0):
...
+@requires_safetensors
+def torch_xla_safe_save_file(
+    tensors: Dict[str, torch.Tensor],
+    filename: Union[str, os.PathLike],
+    metadata: Optional[Dict[str, str]] = None,
+    master_only: bool = True,
+    global_master: bool = False,
+):
+    """
+    Torch XLA compatible implementation of `safetensors.torch.save_file`.
```
JingyaHuang, line 328: Could you elaborate and explain those arguments a bit? I am not familiar with what `master_only` and `global_master` indicate.

JingyaHuang (approved):
LGTM, thanks for the fix!

I just left some very minor nits, mostly because that I want to understand better.

optimum/neuron/trainers.py
```diff
@@ -402,20 +404,27 @@ def _save_xla(self, output_dir: Optional[str] = None):
...
             if not isinstance(self.model, PreTrainedModel):
                 if isinstance(unwrap_model(self.model), PreTrainedModel):
-                    unwrap_model(self.model).save_pretrained(
-                        output_dir,
-                        is_main_process=self.args.should_save,
-                        state_dict=self.model.state_dict(),
-                        save_function=xm.save,
-                    )
+                    with safe_save_function_patcher:
+                        unwrap_model(self.model).save_pretrained(
```
dacorvo, line 413: 1. Could you explain what you want to avoid/achieve here ?
2. Could you explain what will the exact chain of calls be here ?

optimum/neuron/utils/training_utils.py
michaelbenayoun: So those two parameters are related to distributed training.
Basically,  when `master_only` is `True`, only the master rank will be saving the file instead of all ranks, and `global_master` controls whether only the global master (accross multiple nodes) should be saving or the master of each node.

michaelbenayoun: More information [here](https://github.com/pytorch/xla/blob/30458aaae73a496f73c29bb468b04c283877217f/torch_xla/utils/serialization.py#L61-L69)

optimum/neuron/trainers.py
michaelbenayoun: So basically the chain of calls here is:

1. We enter a context manager that patches the `transformers.modeling_utils.safe_save_file` function which is `safetensors.torch.save_file` to be my `torch_xla` compatible version of this function.
2. The model is unwraped (it's not really important here, it's related to the Trainer and how sometimes the model is wrapped for feature I am not really sure we support).
3. The model is saved, and since last Transformers release, it will save the checkpoint using `safetensors`. [The issue is that when `safetensors` is used, the `save_function` parameter is ignored](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2372-L2378). *That is the reason why we do it like that instead of simply passing the `torch_xla_safe_save_file` as the value for `save_function`.
4. We exit the context manager and everything that was patched is restored to its original value.

optimum/neuron/utils/cache_utils.py
michaelbenayoun: This function checks whether is private to the general public but public to the current user.

optimum/neuron/trainers.py
dacorvo: Thanks for the explaination !

dacorvo (approved):
LGTM, thanks !

PR #332, opened 2023-11-16
update ckpt conversion flow to use the new sharded ckpt path structure

5cp:
This [recent change](https://github.com/aws-neuron/neuronx-distributed/commit/a8bce14394c80909f1f68ce9792cfd47da85d285#diff-03393251b3e1f7eba603073ff0b1c3b434e0c755a8b6276bacdd30147f76844cR81) in neuronx-distributed affected the checkpoint paths/filenames when saving sharded checkpoints.

This PR adjusts Optimum Neuron's checkpoint consolidation code to use the new path/filename structure.

### DIFF
--- a/optimum/neuron/distributed/checkpointing.py
+++ b/optimum/neuron/distributed/checkpointing.py
@@ -38,12 +38,10 @@ def consolidate_tensor_parallel_checkpoints(checkpoint_dir: Union[str, Path]) ->
 
     state_dicts = []
 
-    for sharded_checkpoint in checkpoint_dir.glob("tp_rank_*"):
+    for sharded_checkpoint in sorted(checkpoint_dir.glob("tp_rank_*/checkpoint.pt")):
         if not sharded_checkpoint.is_file():
             continue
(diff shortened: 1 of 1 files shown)


optimum/neuron/distributed/checkpointing.py
```diff
@@ -38,13 +38,11 @@ def consolidate_tensor_parallel_checkpoints(checkpoint_dir: Union[str, Path]) ->
 
     state_dicts = []
 
-    for sharded_checkpoint in checkpoint_dir.glob("tp_rank_*"):
+    for sharded_checkpoint in sorted(checkpoint_dir.glob("tp_rank_*/checkpoint.pt")):
```
michaelbenayoun, line 41: This will only work for TP right?
It will need to be updated for TP + PP. That's not an issue, just checking.

michaelbenayoun (approved):

dacorvo:
Merging it despite CI failure: CI tokens used to push models to the hub are not available for PR from forks.

Issue #310, opened 2023-11-09
Issue compiling with single Neuron core - no mark_step executed in training loop


mmcclean-aws:
When attempting to use Optimum Neuron for a single core there is an issue that there is no `xm.mark_step()` executed so many graphs are created for a model when compilation runs. Currently there needs to be at least 2 Neuron Cores running to properly compile a model

JingyaHuang:
I met the same issue before, a workaround would be setting up the `logging_steps` to evoke `xm.mark_step()`. 

https://github.com/huggingface/optimum-neuron/blob/6ed2ecae6023afcf0f8fd99a90ac3c5ba70a86d9/optimum/neuron/trainers.py#L312-L316

@michaelbenayoun shall have more ideas, and I would be happy to help on this topic as well.

michaelbenayoun:
Yes, I gave a quick fix to @mmcclean-aws offline. 
It's a very minor issue I will try to solve this week.

PR #340, opened 2023-11-20
Wrap around MpDeviceLoader all the time

michaelbenayoun:
This PR wraps the dataloader with `MpDeviceLoader` all the time in order to mark step even in the single worker setting.

Fixes: #310


dacorvo (commented):
LGTM but the CI failed on one test (compiler error). I relaunched the job to see if it keeps failing.

PR #344, opened 2023-11-21
Remove numpy pinning

aws-rishyraj:
numpy versioning is determined by neuronx-cc. Having pinning also here can cause issues post-installation when `neuronx-cc` changes its pinnings.

For the future release, `neuronx-cc` will have `numpy>=1.24.3, <=1.25.2`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ INSTALL_REQUIRES = [
     "accelerate == 0.23.0",
     "optimum >= 1.13.0",
     "huggingface_hub >= 0.14.0",
-    "numpy>=1.22.2, <=1.25.2",
     "protobuf<4",
 ]
 


michaelbenayoun:
Maybe we can set it to this new requirement then?

michaelbenayoun:
Alright. WDYT @JingyaHuang @dacorvo ?

dacorvo:
The thing is that `neuronx-cc` is only pulled as an option, so we might need to check if it is possible to do things like installing `optimum-neuron`, get the wrong `numpy` version, then install `optimum-neuron[neuronx]` and face the issue of conflicting `numpy` versions again.
Anyway, we now pin the version of `neuronx-cc` down to the patch version (because these are not really patch versions), so technically we can update the `numpy` pinning whenever we update `neuronx-cc`.

JingyaHuang:
I would agree to remove this and purely let the neuron compiler decide the numpy version, indeed `[neuron]` and `[neuronx]` are extras, but in reality, it should be one or the other case. 

Besides, keep in mind that optimum-neuron supports both inf1 and inf2, I wonder would the pinned numpy version range be different between them one day. If so for robustness, I would prefer letting neuron components make the choice.

dacorvo:
I did a quick test, and `pip install optimum_neuron` will install numpy 1.26.2 (and torch 2.1.1).
No big deal since `pip install optimum_neuron[neuronx]` will uninstall both and install instead the versions required by `neuronx` packages (note that we could also drop the torch 1.13.1.* req).

dacorvo:
As a side note: when the `neuronx_cc` compiler switches to a newer version of the numpy C API (like it did last time), what happens to the compiled artifacts ? Can they still be loaded ?

dacorvo (approved):

aws-rishyraj:
The compiled artifacts are independent of the numpy C API, so they can still be loaded.

PR #346, opened 2023-11-22
[Stable Diffusion - LCM] Add dp option for loading the whole pipe to all cores

JingyaHuang:
* Add `data_parallel_mode` option and set it as default for LCM (without classifier-free guidance, one neuron core will be left idle otherwise).
* Leverage passed `task` for defining compilation flags. @neo
* Accordingly change the doc and tests

### DIFF
--- a/optimum/exporters/neuron/convert.py
+++ b/optimum/exporters/neuron/convert.py
@@ -469,7 +469,7 @@ def add_stable_diffusion_compiler_args(config, compiler_args):
     if "unet" in identifier:
         sdxl_ids = ["stable-diffusion-xl", "sdxl"]
         # SDXL unet doesn't support fast loading neuron binaries
-        if not any(sdxl_id in identifier for sdxl_id in sdxl_ids):
+        if not getattr(config, "is_sdxl", False) and not any(sdxl_id in identifier for sdxl_id in sdxl_ids):
             compiler_args.append("--enable-fast-loading-neuron-binaries")
         compiler_args.append("--model-type=unet-inference")
     return compiler_args
--- a/optimum/exporters/neuron/convert.py
+++ b/optimum/exporters/neuron/convert.py
@@ -467,9 +467,8 @@ def add_stable_diffusion_compiler_args(config, compiler_args):
         compiler_args.append("--enable-fast-loading-neuron-binaries")
     # unet
     if "unet" in identifier:
-        sdxl_ids = ["stable-diffusion-xl", "sdxl"]
         # SDXL unet doesn't support fast loading neuron binaries
-        if not getattr(config, "is_sdxl", False) and not any(sdxl_id in identifier for sdxl_id in sdxl_ids):
+        if not getattr(config, "is_sdxl", False):
             compiler_args.append("--enable-fast-loading-neuron-binaries")
         compiler_args.append("--model-type=unet-inference")
     return compiler_args
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -21,7 +21,7 @@ import shutil
 from abc import abstractmethod
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
 
 import torch
 from huggingface_hub import snapshot_download
@@ -92,11 +92,11 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
         config: Dict[str, Any],
         tokenizer: CLIPTokenizer,
         scheduler: Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler, LCMScheduler],
+        data_parallel_mode: str,
         vae_encoder: Optional[Union[torch.jit._script.ScriptModule, "NeuronModelVaeEncoder"]] = None,
         text_encoder_2: Optional[Union[torch.jit._script.ScriptModule, "NeuronModelTextEncoder"]] = None,
         tokenizer_2: Optional[CLIPTokenizer] = None,
         feature_extractor: Optional[CLIPFeatureExtractor] = None,
-        device_ids: Optional[List[int]] = None,
         configs: Optional[Dict[str, "PretrainedConfig"]] = None,
         neuron_configs: Optional[Dict[str, "NeuronConfig"]] = None,
         model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
@@ -118,6 +118,9 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
                 [CLIPTokenizer](https://huggingface.co/docs/transformers/v4.21.0/en/model_doc/clip#transformers.CLIPTokenizer).
             scheduler (`Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler]`):
                 A scheduler to be used in combination with the U-NET component to denoise the encoded image latents.
+            data_parallel_mode (`str`):
+                Mode to decide what components to load into both NeuronCores of a Neuron device. Can be "none"(no data parallel), "unet"(only
+                load unet into both cores of each device), "all"(load the whole pipeline into both cores).
             vae_encoder (`Optional[torch.jit._script.ScriptModule]`, defaults to `None`):
                 The Neuron TorchScript module associated to the VAE encoder.
             text_encoder_2 (`Optional[torch.jit._script.ScriptModule]`, defaults to `None`):
@@ -127,8 +130,6 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
                 [CLIPTokenizer](https://huggingface.co/docs/transformers/v4.21.0/en/model_doc/clip#transformers.CLIPTokenizer).
             feature_extractor (`Optional[CLIPFeatureExtractor]`, defaults to `None`):
                 A model extracting features from generated images to be used as inputs for the `safety_checker`
-            device_ids (Optional[List[int]], defaults to `None`):
-                A list of integers that specify the NeuronCores to use for parallelization, by default will load the UNet model into 2 neuron cores.
             configs (Optional[Dict[str, "PretrainedConfig"]], defaults to `None`):
                 A dictionary configurations for components of the pipeline.
             neuron_configs (Optional["NeuronConfig"], defaults to `None`):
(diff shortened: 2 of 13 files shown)


optimum/exporters/neuron/convert.py
```diff
@@ -469,7 +469,7 @@ def add_stable_diffusion_compiler_args(config, compiler_args):
     if "unet" in identifier:
         sdxl_ids = ["stable-diffusion-xl", "sdxl"]
         # SDXL unet doesn't support fast loading neuron binaries
-        if not any(sdxl_id in identifier for sdxl_id in sdxl_ids):
+        if not getattr(config, "is_sdxl", False) and not any(sdxl_id in identifier for sdxl_id in sdxl_ids):
```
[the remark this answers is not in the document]
JingyaHuang: Yep , [`_infer_task_from_model_name_or_path`](https://github.com/huggingface/optimum/blob/521d0696f2f252939b6dcbee011d56b63f4f3101/optimum/exporters/tasks.py#L1412) do the job, I will remove the second part and just depends on the task then.

optimum/neuron/modeling_diffusion.py
```diff
@@ -262,35 +268,58 @@ def load_model(
...
-            unet = torch_neuronx.DataParallel(
+        submodels = {
+            "text_encoder": text_encoder_path,
+            "unet": unet_path,
+            "vae_decoder": vae_decoder_path,
+            "vae_encoder": vae_encoder_path,
+            "text_encoder_2": text_encoder_2_path,
+        }
+        if data_parallel_mode == "all":
+            logger.info("Loading the whole pipeline into both Neuron Cores...")
```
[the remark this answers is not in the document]
JingyaHuang: In the modeling API we only consider the core allocation on each device, and each neuron device has two cores max. If users run with an instance with multiple neuron device, then they can launch multiple processes and each with `NEURON_RT_NUM_CORES=2`.

dacorvo (approved):

optimum/neuron/modeling_diffusion.py
```diff
@@ -92,11 +92,11 @@ def __init__(
         config: Dict[str, Any],
         tokenizer: CLIPTokenizer,
         scheduler: Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler, LCMScheduler],
+        data_parallel_mode: str,
```
neo, line 95:
```suggestion
        data_parallel_mode: Literal["none", "unet", "all"],
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -118,6 +118,9 @@ def __init__(
                 [CLIPTokenizer](https://huggingface.co/docs/transformers/v4.21.0/en/model_doc/clip#transformers.CLIPTokenizer).
             scheduler (`Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler]`):
                 A scheduler to be used in combination with the U-NET component to denoise the encoded image latents.
+            data_parallel_mode (`str`):
```
neo, line 121:
```suggestion
            data_parallel_mode (`Literal["none", "unet", "all"]`):
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -231,27 +234,30 @@ def __init__(
...
+    @staticmethod
+    def is_lcm(unet_config):
         patterns = ["lcm", "latent-consistency"]
-        unet_name_or_path = getattr(self.unet.config, "_name_or_path", "").lower()
+        unet_name_or_path = getattr(unet_config, "_name_or_path", "").lower()
         return any(pattern in unet_name_or_path for pattern in patterns)
 
     @staticmethod
     def load_model(
+        data_parallel_mode: Optional[str],
```
neo, line 245:
```suggestion
        data_parallel_mode: Optional[Literal["none", "unet", "all"]],
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -231,27 +234,30 @@ def __init__(
...
         text_encoder_2_path: Optional[Union[str, Path]] = None,
-        device_ids: Optional[List[int]] = None,
         dynamic_batch_size: bool = False,
     ):
         """
         Loads Stable Diffusion TorchScript modules compiled by neuron(x)-cc compiler. It will be first loaded onto CPU and then moved to
         one or multiple [NeuronCore](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/arch/neuron-hardware/neuroncores-arch.html).
 
         Args:
+            data_parallel_mode (`Optional[str]`):
```
neo, line 258:
```suggestion
            data_parallel_mode (`Optional[Literal["none", "unet", "all"]]`):
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -262,35 +268,58 @@ def load_model(
...
-            "vae_encoder": vae_encoder,
-            "text_encoder_2": text_encoder_2,
-        }
+            raise ValueError("You need to pass `data_parallel_mode` to define Neuron Core allocation.")
+
+        return submodels
+
+    @staticmethod
+    def set_default_dp_mode(unet_config):
+        if NeuronStableDiffusionPipelineBase.is_lcm(unet_config) is True:
```
neo, line 317:
```suggestion
        if NeuronStableDiffusionPipelineBase.is_lcm(unet_config):
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -381,7 +410,7 @@ def _from_pretrained(
         vae_decoder: Optional["NeuronModelVaeDecoder"] = None,
         local_files_only: bool = False,
         model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
-        device_ids: Optional[List[int]] = None,
+        data_parallel_mode: Optional[str] = None,
```
neo, line 413:
```suggestion
        data_parallel_mode: Optional[Literal["none", "unet", "all"]] = None,
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -509,9 +541,63 @@ def _from_transformers(
         disable_fast_relayout: Optional[bool] = False,
         disable_fallback: bool = False,
         dynamic_batch_size: bool = False,
-        device_ids: Optional[List[int]] = None,
+        data_parallel_mode: Optional[str] = None,
```
neo, line 544:
```suggestion
        data_parallel_mode: Optional[Literal["none", "unet", "all"]] = None,
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -509,9 +541,63 @@ def _from_transformers(
...
+                The data type to cast FP32 operations to when auto-cast mode is enabled. Can be `"bf16"`, `"fp16"` or `"tf32"`.
+            disable_fast_relayout (`Optional[str]`, defaults to `None`):
+                (INF1 ONLY) Whether to disable fast relayout optimization which improves performance by using the matrix multiplier for tensor transpose.
+            disable_fallback (`bool`, defaults to `False`):
+                (INF1 ONLY) Whether to disable CPU partitioning to force operations to Neuron. Defaults to `False`, as without fallback, there could be
+                some compilation failures or performance problems.
+            dynamic_batch_size (`bool`, defaults to `False`):
+                Whether to enable dynamic batch size for neuron compiled model. If this option is enabled, the input batch size can be a multiple of the
+                batch size during the compilation, but it comes with a potential tradeoff in terms of latency.
+            data_parallel_mode (`Optional[str]`, defaults to `None`):
```
neo, line 595:
```suggestion
            data_parallel_mode (`Optional[Literal["none", "unet", "all"]]`, defaults to `None`):
```

optimum/neuron/modeling_diffusion.py
```diff
@@ -719,11 +805,11 @@ def __init__(
         config: Dict[str, Any],
         tokenizer: CLIPTokenizer,
         scheduler: Union[DDIMScheduler, PNDMScheduler, LMSDiscreteScheduler],
+        data_parallel_mode: str,
```
neo, line 808:
```suggestion
        data_parallel_mode: Literal["none", "unet", "all"],
```

optimum/neuron/pipelines/diffusers/pipeline_utils.py
```diff
@@ -24,7 +24,11 @@
 
 class DiffusionBasePipelineMixin:
     def check_num_images_per_prompt(self, prompt_batch_size: int, neuron_batch_size: int, num_images_per_prompt: int):
-        if not self.dynamic_batch_size and neuron_batch_size != prompt_batch_size * num_images_per_prompt:
+        if (
+            not self.data_parallel_mode == "all"
```
neo, line 28: shall we do a separate calculation check for when the mode is "all" instead of just letting it pass unconditionally when it's "all"?

neo (commented):
one code consideration and another typing suggestion 😆

neo:
lol was a bit too late 🤦 

but I do have a question: does this mean lcm will be even faster with this because it's loading everything into both cores instead of just the unet?

optimum/neuron/pipelines/diffusers/pipeline_utils.py
JingyaHuang: Currently there are some constraints on compiling LCM with bs>1. I would leave a placeholder here and do the check better once I sort the compilation issue out with the Annapurna team.
(for now LCM is the only case where we need to load the "all" pipeline into both cores)

neo:
oh ok, that makes sense – that without the need to do cfg in the other core, it can do another image at the same time

thank you for clarifying!

because my follow up question would be can I make single inference run in multiple cores or even multiple devices to further reduce the latency and I'm guessing the answer is not possible, is it?

JingyaHuang:
Each Neuron device has two cores, so if you use a larger instance with more than one Neuron device, you can run multiple processes to use them all. Eg. With inf2.24xlarge (6 neuron devices, 12 Neuron cores), you could generate 12 images simultaneously.

neo:
yes that's my understand, but not generating one image 12x faster, right?

JingyaHuang:
nope.

neo:
coming back with another question: I read from discussion saying that if we use negative prompt with cfg in LCM, it would be "(2x?) slower because of the two passes"; while I couldn't really understand whether that's true or why with my very limited understanding, I do wonder whether we can put the "two passes" of LCM with negative prompt in the two different cores? If so, does it essentially mean that it would return to be the same as the non LCM unet approach being loaded in to two cores?

thanks! 🙇

PR #350, opened 2023-11-23
Support decoding sequences of byte tokens

dacorvo:
Some tokens must be decoded in sequence to produce non-standard characters like emoji or ideograms.

For instance, :smiley: is encoded as [243, 162, 155, 134].

The decoding is correct if you process all tokens:
```
>>> tokenizer.decode([243, 162, 155, 134])
'😃'
``` 
But when streaming the current TGI server decodes and returns tokens one at a time:
```
>>> tokenizer.decode(243), tokenizer.decode(162), tokenizer.decode(155), tokenizer.decode(134)
'�', '�', '�', �'
```

This change attempts to decode each token and discard the result until it produces something meaningful.
This means that during the generation, when decoding a sequence of n bytes, the server will return n-1 empty strings before the final correct string.

In other words, when decoding '😃', the server will return "", "", "", "😃".

### DIFF
--- a/text-generation-inference/server/text_generation_server/generator.py
+++ b/text-generation-inference/server/text_generation_server/generator.py
@@ -94,8 +94,9 @@ class Slot:
         PAUSE = 1
         READY = 2
 
-    def __init__(self, id: int):
+    def __init__(self, id: int, tokenizer: PreTrainedTokenizerBase):
         self._id = id
+        self._tokenizer = tokenizer
         self.clear()
 
     def clear(self):
@@ -108,7 +109,9 @@ class Slot:
         self._mask = []
         self._selector = None
         self._generated_tokens = 0
-        self._next_token_text = ""
+        self._next_text_token_start = 0
+        self._next_text_token_end = 0
+        self._next_text = ""
 
     @property
     def id(self) -> int:
@@ -131,7 +134,7 @@ class Slot:
         return self._generation_config
 
     @property
-    def generated_tokens(self) -> torch.LongTensor:
+    def generated_tokens(self) -> int:
         return self._generated_tokens
 
     def assign(self, request: Request, generation_config: GenerationConfig):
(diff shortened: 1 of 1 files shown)


text-generation-inference/server/text_generation_server/generator.py
```diff
@@ -188,7 +193,32 @@ def resume(self):
             self._mask = torch.cat([self._mask, torch.LongTensor([1])])
         self._state = Slot.State.READY
 
-    def append(self, next_token: int, next_token_text: str):
+    def _decode_next_tokens(
+        self,
+    ) -> str:
+        """Hack to hopefully support generate_stream for the maximum number of tokenizers"""
```
philschmid, line 199: 😅 
can we test that?

text-generation-inference/server/text_generation_server/generator.py
```diff
@@ -188,7 +193,32 @@ def resume(self):
...
+        # We need to include the tokens that produced the last text to defeat cleanup algorithms in the decode
+        # which decide to add a space or not depending on the surrounding ids.
+        new_text = self._tokenizer.decode(self._tokens[self._next_text_token_start :], skip_special_tokens=False)
+        if new_text.endswith("�"):
```
philschmid, line 203: feels weird no?

philschmid (approved):

text-generation-inference/server/text_generation_server/generator.py
dacorvo: I tested it manually. I would need a bit of work to test individually some parts of the TGI server, without loading the whole thing. I will do that in an upcoming pull-request so that this goes into the release today.

philschmid: SGTM!

text-generation-inference/server/text_generation_server/generator.py
dacorvo: https://github.com/huggingface/text-generation-inference/blob/b226e469c95d5abe4a67ca0aff758b49338654b0/server/text_generation_server/models/model.py#L68C14-L68C14

dacorvo: I shamelessly copied the code from the TGI implementation.

PR #351, opened 2023-11-23
Extend lcm doc

JingyaHuang:
Extend a bit the doc of LCM

### DIFF
--- a/docs/source/tutorials/stable_diffusion.mdx
+++ b/docs/source/tutorials/stable_diffusion.mdx
@@ -356,7 +356,9 @@ To avoid Neuron device out of memory, it's suggested to finish all base inferenc
 
 Latent Consistency Models (LCMs) were proposed in [Latent Consistency Models: Synthesizing High-Resolution Images with Few-Step Inference by Simian Luo, Yiqin Tan, Longbo Huang, Jian Li, and Hang Zhao](https://huggingface.co/papers/2310.04378). LCMs enable inference with fewer steps on any pre-trained LDMs, including Stable Diffusion and SDXL.
 
-In `Optimum-Neuron`, you can use the class `NeuronLatentConsistencyModelPipeline` to compile and run inference of LCMs distilled from Stable Diffusion (SD) models, and continue to use the class `NeuronStableDiffusionXLPipeline` for LCMs distilled from SDXL models.
+In `optimum-neuron`, you can:
+  - Use the class `NeuronLatentConsistencyModelPipeline` to compile and run inference of LCMs distilled from Stable Diffusion (SD) models, 
+  - And continue to use the class `NeuronStableDiffusionXLPipeline` for LCMs distilled from SDXL models.
 
 Here are examples to compile the LCMs of Stable Diffusion ( [SimianLuo/LCM_Dreamshaper_v7](https://huggingface.co/SimianLuo/LCM_Dreamshaper_v7) ) and Stable Diffusion XL( [latent-consistency/lcm-sdxl](https://huggingface.co/latent-consistency/lcm-sdxl) ), and then run inference on AWS Inferentia 2 :
 
--- a/docs/source/tutorials/stable_diffusion.mdx
+++ b/docs/source/tutorials/stable_diffusion.mdx
@@ -413,11 +413,9 @@ Now we can generate images from text prompts on Inf2 using the pre-compiled mode
 from optimum.neuron import NeuronLatentConsistencyModelPipeline
 
 pipe = NeuronLatentConsistencyModelPipeline.from_pretrained("Jingya/LCM_Dreamshaper_v7_neuronx")
+prompts = ["Self-portrait oil painting, a beautiful cyborg with golden hair, 8k"] * 2
 
-num_images_per_prompt = 2
-prompt = ["Self-portrait oil painting, a beautiful cyborg with golden hair, 8k"] * num_images_per_prompt
-
-images = pipe(prompt=prompt, num_inference_steps=4, guidance_scale=8.0).images
+images = pipe(prompt=prompts, num_inference_steps=4, guidance_scale=8.0).images
 ```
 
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/stable_diffusion.mdx
```diff
@@ -378,20 +400,43 @@ save_directory = "lcm_sdxl_neuron/"
...
 
+***LCM of Stable Diffusion***
+
+```python
+from optimum.neuron import NeuronLatentConsistencyModelPipeline
+
+pipe = NeuronLatentConsistencyModelPipeline.from_pretrained("Jingya/LCM_Dreamshaper_v7_neuronx")
+
+num_images_per_prompt = 2
+prompt = ["Self-portrait oil painting, a beautiful cyborg with golden hair, 8k"] * num_images_per_prompt
```
dacorvo, line 418: It should actually be prompts there, no ? It is a bit confuzing to have a batch size of 1 but passing multiple prompts.

JingyaHuang: In the modeling, `prompt` can either be a str or a list of prompts, so I find both work for me

https://github.com/huggingface/optimum-neuron/blob/ac90dc3828e13623b878f19b559998d011f7d66b/optimum/neuron/pipelines/diffusers/pipeline_stable_diffusion_xl.py#L40

docs/source/tutorials/stable_diffusion.mdx
```diff
@@ -356,11 +356,33 @@ To avoid Neuron device out of memory, it's suggested to finish all base inferenc
 
 Latent Consistency Models (LCMs) were proposed in [Latent Consistency Models: Synthesizing High-Resolution Images with Few-Step Inference by Simian Luo, Yiqin Tan, Longbo Huang, Jian Li, and Hang Zhao](https://huggingface.co/papers/2310.04378). LCMs enable inference with fewer steps on any pre-trained LDMs, including Stable Diffusion and SDXL.
 
-Here is an example to compile a LCM of Stable Diffusion XL: [latent-consistency/lcm-sdxl](https://huggingface.co/latent-consistency/lcm-sdxl) and run inference on AWS Inferentia 2 :
+In `Optimum-Neuron`, you can use the class `NeuronLatentConsistencyModelPipeline` to compile and run inference of LCMs distilled from Stable Diffusion (SD) models, and continue to use the class `NeuronStableDiffusionXLPipeline` for LCMs distilled from SDXL models.
```
michaelbenayoun, line 359:
```suggestion
In `optimum-neuron`, you can:
  - Use the class `NeuronLatentConsistencyModelPipeline` to compile and run inference of LCMs distilled from Stable Diffusion (SD) models, 
  - And continue to use the class `NeuronStableDiffusionXLPipeline` for LCMs distilled from SDXL models.
```

docs/source/tutorials/stable_diffusion.mdx
```diff
@@ -356,11 +356,33 @@ To avoid Neuron device out of memory, it's suggested to finish all base inferenc
...
-***Export via Python API***
+***LCM of Stable Diffusion***
+
+```python
+from optimum.neuron import NeuronLatentConsistencyModelPipeline
+
+model_id = "SimianLuo/LCM_Dreamshaper_v7"
+num_images_per_prompt = 1
+input_shapes = {"batch_size": 1, "height": 768, "width": 768, "num_images_per_prompt": num_images_per_prompt}
+compiler_args = {"auto_cast": "matmul", "auto_cast_type": "bf16"}
```
michaelbenayoun, line 373: I would format those in multiple lines as I found it easier to read in the doc.

docs/source/tutorials/stable_diffusion.mdx
```diff
@@ -356,11 +356,33 @@ To avoid Neuron device out of memory, it's suggested to finish all base inferenc
...
+
+stable_diffusion = NeuronLatentConsistencyModelPipeline.from_pretrained(
+    model_id, export=True, **compiler_args, **input_shapes
+)
+save_directory = "lcm_sd_neuron/"
+stable_diffusion.save_pretrained(save_directory)
+
+# Push to hub
+stable_diffusion.push_to_hub(save_directory, repository_id="Jingya/LCM_Dreamshaper_v7_neuronx", use_auth_token=True)
```
michaelbenayoun, line 382: Same comment.

michaelbenayoun (approved):
LGTM!
Left a few nits

docs/source/tutorials/stable_diffusion.mdx
JingyaHuang: Yeah, but as we put all features of sd under the same section, I try to let the snippet not look too long...

So far with the scroll bar I find it ok to read so preferably keep it that way.

docs/source/tutorials/stable_diffusion.mdx
dacorvo: Sorry, my comment was not specific enough. It is ok to pass multiple prompts, but the normal behaviour for a pipe is to process them one at a time, since you did not specify a batch_size of 1 when instantiating it, and the model has been compiled with batch_size 1 too.
I am assuming here that you will actually process two images in parallel, which should correspond to an actual batch size of 2.
But maybe I am getting it wrong.

dacorvo: At least that's the behaviour of the transformers pipelines ... but maybe for diffusers, it is different.

JingyaHuang: LCM currently can only be compiled with bs=1, during the inference, since we are doing dp for the whole pipe, yes exactly we would be able to run bs=2 in parallel.

I know it's confusing that this is not processed inside the pipeline with an argument `num_images_per_prompt`, but we need to pass the prompt twice into the pipeline... Well I would say it's temporary, I hope that Annapurna could unblock me with the compilation issue so that I can better organise the bs along with arguments like `num_images_per_prompt`.

PR #354, opened 2023-11-24
Push after running `neuron_parallel_compile`

michaelbenayoun:
This PR adds support for pushing to the cache repository on the Hugging Face Hub after using `neuron_paralell_compile`. 

Basically, when running a training after doing `neuron_parallel_compile`, all the compilation files will be pushed to the Hub before starting actual training.

cc @5cp

### DIFF
--- a/optimum/neuron/utils/cache_utils.py
+++ b/optimum/neuron/utils/cache_utils.py
@@ -52,7 +52,7 @@ from .version_utils import get_neuronxcc_version
 
 logger = logging.get_logger()
 
-SHOULD_LOG = should_log()
+# SHOULD_LOG = should_log()
 
 HOME = Path.home()
 DEFAULT_HF_HOME = f"{HOME}/.cache/huggingface"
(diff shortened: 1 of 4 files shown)


optimum/neuron/utils/cache_utils.py
```diff
@@ -41,20 +41,19 @@
...
 
 
-if TYPE_CHECKING:
-    from transformers import PretrainedConfig, PreTrainedModel
-
-
 logger = logging.get_logger()
 
+# SHOULD_LOG = should_log()
+
```
dacorvo, line 56:
```suggestion
```

optimum/neuron/utils/misc.py
```diff
@@ -38,12 +38,21 @@
...
 
 from ...utils import logging
+from .import_utils import is_torch_xla_available
 from .require_utils import requires_safetensors
 
 
 logger = logging.get_logger()
 
 
+def should_log() -> bool:
```
dacorvo, line 48: This is very specific to neuronx_distributed, and not really related to logging. Can you move it under the neuronx_distributed directory, and choose a name that clearly states that this verifies we are in the process of the first worker (that's what I understood).

optimum/neuron/utils/misc.py
```diff
@@ -45,7 +45,7 @@
 logger = logging.get_logger()
 
 
-def should_log() -> bool:
+def should_current_worker_log() -> bool:
```
dacorvo, line 48: It still has nothing to do with log. I would call it is_main_worker, wdyt ?

dacorvo (approved):
Maybe change the name, otherwise LGTM ...

optimum/neuron/utils/misc.py
michaelbenayoun: Alright, doing it.

PR #361, opened 2023-12-04
[Inference] T5 Tensor Parallel support

JingyaHuang:
* Add support for exporting T5 when apply tensor parallel with `neuronx_distributed`
* Add support for inference with the traced parallel T5 model
* Tests
* Documentation for T5 (small ones and large ones with TP)


JingyaHuang:
not stale

PR #108, opened 2023-06-23
Enable various sampling techniques

bocchris-aws:
Enabling multiple sampling techniques by wrapping the models forward function. Before each forward pass we pad the input and move respective tensors to the XLA device. Directly after the forward, we "unpad" the logits such that any subsequent processing will work. This way we can call the forward pass on the XLA device without recompilations. 
Tested on T5-small and BART-base for constrained beam search.

PR #370, opened 2023-12-07
Add general support for generation on TRN with NxD

aws-tianquaw:
This pull request adds general support for generative LM generation on TRN. Currently, the generation can only be done separately from the training job with `transformer_neuronx` inference feature. The limitation is that it needs to be run as a separate program from the training job. With this support added, the users will be able to run model evaluation as part of the training job and choose to run it every n steps or epochs. Although we have implemented `NeuronGenerationMixin` class that delivers a similar feature, it has some drawbacks and functionality issues which I'll specify soon.

The generation will be done on both Neuron device and CPU in the following way:
1. Model forward pass will be executed on TRN
2. All other logics including padding, searching, and sampling will be handled by general device (CPU).

This enables us to support various of search/sampling methods provided by huggingface with minimal code changes. It has a couple of advantages comparing to [the existing implementation of `NeuronGenerationMixin` class](https://github.com/huggingface/optimum-neuron/blob/main/optimum/neuron/generation/utils.py):
1. The new implementation only overrides a small portion of `generate` function and call `super().generate()` to reuse most of HuggingFace implementation. Therefore, it will be more stable and less likely to brake as HuggingFace updates transformers package.
2. The new implementation has general supports all HF search/sampling methods while the existing one only supports greedy search and beam search (which doesn't work with latest transformers)
3. The new implementation is much more light-weighted so it's much easier to maintain.

I've tested the functionality of text generation with GPT2 and Llama-2-7b models using [this script](https://github.com/huggingface/transformers/blob/v4.35.0/examples/pytorch/text-generation/run_generation.py). The output results on TRN exactly matches the CPU outputs.

Reference: https://github.com/huggingface/optimum-neuron/pull/108


optimum/neuron/utils/generation_utils.py
```diff
@@ -0,0 +1,188 @@
+# coding=utf-8
```
dacorvo, line 1: I'd rather not add yet another util file here: I think all these functions can go into the same `generation/utils.py` file.

Also, I would prefix them with an underscore, since they are never used anywhere else.

optimum/neuron/utils/generation_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        `torch.Tensor`: Padded `input_ids`.
+    """
+    bsz = input_ids.size(0)
+    input_ids = torch.cat(
+        [input_ids, torch.ones((bsz, num_padding_values), device=input_ids.device, dtype=torch.long) * pad_token_id], 1
+    )
+    return input_ids
+
+
+def get_fwd_for_general_sampling(
```
dacorvo, line 103: Since eventually you are overloading forward, you could as well have done it statically in your mixin, which for now contains a single overloaded method (generate).

optimum/neuron/utils/generation_utils.py
```diff
@@ -0,0 +1,188 @@
...
+                position_ids = kwargs["attention_mask"].long().cumsum(-1) - 1
+                position_ids.masked_fill_(kwargs["attention_mask"] == 0, 1)
+                kwargs["position_ids"] = position_ids
+
+        # Move inputs to device
+        move_dict_args_to_device(kwargs, main_device)
+
+        # Forward
+        kwargs = args_and_kwargs_to_kwargs_only(current_fwd, args, kwargs)
+        outputs = current_fwd(**kwargs)
```
dacorvo, line 167: If you overload, here you can simply call super().forward

optimum/neuron/utils/generation_utils.py
```diff
@@ -0,0 +1,188 @@
...
+                dim=-1,
+                groups=parallel_state.get_tensor_model_parallel_group(as_list=True),
+            )
+        xm.mark_step()
+
+        # Move to CPU
+        move_dict_args_to_device(outputs, to_device)
+
+        # Post-process output as a function of cur_len
+        outputs["logits"] = outputs["logits"][:, :cur_len, ...].to(torch.float32)
```
dacorvo, line 184: Maybe restore instead the dtype of the original outputs (could be float 16 ?).

optimum/neuron/utils/training_utils.py
```diff
@@ -263,6 +263,32 @@ def patch_generation_mixin_to_neuron_generation_mixin(model: "PreTrainedModel"):
         cls.__bases__ = tuple(new_bases)
 
 
+def patch_generation_mixin_to_general_neuron_generation_mixin(model: "PreTrainedModel"):
```
dacorvo, line 266: You need to add unit tests for this new feature.

At the very minimum, these tests should prove that greedy generation and beam search are working as expected, with the basic generation configurations.
You can take as a reference the corresponding tests for T5:
https://github.com/huggingface/optimum-neuron/blob/main/tests/generation/test_generate.py

dacorvo (changes requested):
Thank you very much for this pull-request.
I am fully aligned with the fact that the existing `NeuronGenerationMixin` needs to be replaced by a more robust and easier to maintain implementation, and this goes definitely in the right direction.
An additional benefit is that this new class will be compatible with the logits processors, some of which are really important (like repetition penalty).
I have made a few comments on the implementation itself, mostly related to how things are organized (I would rather put everything at the same place, if possible inside the mixin).
My main request however is related to the tests: I am not comfortable with integrating new code without proper unit tests.
These tests will not only allow us to detect early bugs or configurations we don't support, but will also help us detecting regressions when bumping the transformers version.

optimum/neuron/utils/generation_utils.py
aws-tianquaw: According to HuggingFace's implementation, `GenerationMixin` class is the base class of `PretraineModel` classes. I cannot call `super().forward` in `GeneralNeuronGenerationMixin` class because its base class doesn't implement any `forward` method.  I can access the model's forward using `self.forward` in GeneralNeuronGenerationMixin` class because the child model classes never implement `generate` function.

optimum/neuron/utils/generation_utils.py
aws-tianquaw: Yes I agree, I'll move the functions in this class to `generation/utils.py`

optimum/neuron/utils/generation_utils.py
aws-tianquaw: Good point! This is mainly for fixing the datatype that is upcast in some NxD neuron models. It's better to restore to original data type here

optimum/neuron/utils/training_utils.py
aws-tianquaw: Sure, I'll add unit tests following the example

optimum/neuron/utils/generation_utils.py
dacorvo: You're right: `forward` is provided by the child class whose `GenerationMixin` inheritance is replaced by this new class, so your only option is to overload it dynamically.
I am wondering what happens when you call generate a second time: since you never restore the original forward, will it not result in wrapping forward again and again ?
I would personally have tried doing it in the constructor: the model class always needs to call super() in its constructor because of the `torch.nn.Module` inheritance.

aws-tianquaw: I'm working on more testings right now, and I found similar issue regarding wrapping the forward. Once we replaced the forward method, we need to restore it at the end of the generation in general. Otherwise, the following training steps will brake. There are probably two ways to fix it:
1. Make the new forward general enough for both training and evaluation, and wrap it once in the constructor.
2. In the generation function, wrap the forward and then restore it after generation is done.

I'll explore both and share updates.

dacorvo: If you want to wrap temporarily the forward, instead of patching, you could also use the pytorch built-in hooks.
I personally like wrapping them in a context like illustrated here:
https://github.com/huggingface/quanto/blob/main/quanto/quantization/calibrate.py

aws-tianquaw: It seems like we the `kwargs` which we need to update will not be passed to the hook, so I think maybe we should not use the hooks in this case. See https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L189-L190

dacorvo: Ah yes, you're right: these hooks are rather designed for modules whose forward method is usually simpler than for models.

dacorvo (changes requested):
Thanks you very much for this pull-request, and taking the time to fully address my comments.
Before merging, you need to:
- apply styling (`make style`),
- rebase on main (another pull-request has been merged to use the latest AWS Neuron SDK, sorry about that ...),
- ping us to trigger the CI tests again.

dacorvo (changes requested):
The test_runner failures in the CI are related to the fact that you don't have access to the HF internal token, and you should not worry about them.
However, two of the tests that you added are failing.
```
tests/test_cache_utils.py ......ssssssssss                               [ 21%]
tests/test_generate.py sssss                                             [ 28%]
tests/test_runner.py EEEEEE                                              [ 36%]
tests/test_trainer_callback.py ssss                                      [ 41%]
tests/test_trainers.py sssssss                                           [ 50%]
tests/test_utils.py .                                                    [ 52%]
tests/cli/test_neuron_cache_cli.py ssssss                                [ 60%]
tests/generation/test_generate.py ............FF................         [100%]
```

aws-tianquaw:
> The test_runner failures in the CI are related to the fact that you don't have access to the HF internal token, and you should not worry about them. However, two of the tests that you added are failing.
> 
> ```
> tests/test_cache_utils.py ......ssssssssss                               [ 21%]
> tests/test_generate.py sssss                                             [ 28%]
> tests/test_runner.py EEEEEE                                              [ 36%]
> tests/test_trainer_callback.py ssss                                      [ 41%]
> tests/test_trainers.py sssssss                                           [ 50%]
> tests/test_utils.py .                                                    [ 52%]
> tests/cli/test_neuron_cache_cli.py ssssss                                [ 60%]
> tests/generation/test_generate.py ............FF................         [100%]
> ```

The tests are now passing. Could you please check if there is any other merge blocker? If no, when can we target to merge this PR?

dacorvo (approved):
Thank you so much for your patience and dedication !

PR #372, opened 2023-12-11
[Export] Add extra args for exporter CLI

JingyaHuang:
* Add `compiler_workdir` to store neff files during the compilation
* Add `optlevel` ("O1" / "O2" / "O3")


tests/cli/test_export_cli.py
```diff
@@ -152,6 +152,57 @@ def test_dynamic_batching(self):
...
+                    "1",
+                    "--task",
+                    "text-classification",
+                    "--compiler_workdir",
+                    f"{tempdir}/neff",
+                    tempdir,
+                ],
+                shell=False,
+                check=True,
+            )
```
dacorvo, line 204: Maybe here you could check that there is a NEFF file in the tempdir/neff directory.

tests/inference/test_modeling.py
```diff
@@ -123,6 +123,15 @@ def test_trust_remote_code(self):
...
         self.assertIsInstance(model.config, PretrainedConfig)
 
+    @requires_neuronx
+    def test_save_compiler_intermediary_files(self):
+        with tempfile.TemporaryDirectory() as tempdir:
+            save_path = f"{tempdir}/neff"
+            _ = NeuronModelForSequenceClassification.from_pretrained(
+                self.MODEL_ID, export=True, compiler_workdir=save_path, **self.STATIC_INPUTS_SHAPES
+            )
+            self.assertTrue(os.path.isdir(save_path))
```
dacorvo, line 133: Maybe also check there is at least one neff in it ?

dacorvo (approved):
Thanks for the pull-request ! LGTM, but maybe you can improve a little the tests.

tests/cli/test_export_cli.py
JingyaHuang: sounds good, I will add a check for that!

tests/inference/test_modeling.py
JingyaHuang: same.

Issue #373, opened 2023-12-11
Need to setup env. var MALLOC_ARENA_MAX when training with Optimum Neuron


5cp:
Due to a [known XLA/glibc memory issue](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/release-notes/torch/torch-neuronx/index.html#known-issues-and-limitations-training), we need to set env. var MALLOC_ARENA_MAX during training jobs in order to avoid OOMs. As an example, please see the [NxD Llama2-7B example launch script](https://github.com/aws-neuron/aws-neuron-samples/blob/master/torch-neuronx/training/llama2/tp_zero1_llama2_7b_hf_pretrain/tp_zero1_llama2_7b_hf_pretrain.sh#L15). Without this in place, it is common to observe OOMs during checkpointing.

Can this env. var be enabled via Optimum Neuron? Or should we recommend that it be specified on the command-line when launching the training job?

5cp:
Update - I tried setting MALLOC_ARENA_MAX in several places, both within Optimum Neuron as well as run_clm.py. For maximum memory savings this env. var must be set before invoking torchrun.

I suggest we add this to the llama2-7B fine-tuning example, ex:

```
MALLOC_ARENA_MAX=64 torchrun \
 --nproc_per_node=32 \
 optimum-neuron-git/examples/language-modeling/run_clm.py \
 --model_name_or_path NousResearch/Llama-2-7b-hf \
 --dataset_name timdettmers/openassistant-guanaco \
 --do_train True \
 --bf16 True \
 --learning_rate 5e-5 \
 --output_dir test_output \
 --overwrite_output_dir True \
 --per_device_train_batch_size 1 \
 --gradient_checkpointing True \
 --tensor_parallel_size 8 \
 --block_size 4096 \
 --max_steps 25 \
 --logging_steps 5 \
 --gradient_accumulation_steps 2
```

Issue #368, opened 2023-12-07
Neuron `consolidate` leads to size mismatch. 


philschmid:
When trying to load a model after training, which was consolidated with `optimum-cli neuron consolidate dolly_llama/tensor_parallel_shards dolly_llama` you get an tensor size miss match error. 
I tried to fine-tune llama on dolly dataset, the training succeeds with TP=8. Afterwards i consolidated the weights and tried to load the model to make sure it learned the dolly format. For this i tried to load the model with the `NeuronModelForCausalLM` and `AutoModelForCausalLM` class, e.g.

```python
from transformers import AutoTokenizer, AutoModelForCausalLM

tokenizer = AutoTokenizer.from_pretrained("dolly_llama")
model = AutoModelForCausalLM.from_pretrained("dolly_llama")
```

This leads to the error

```bash
Error(s) in loading state_dict for LlamaForCausalLM:
	size mismatch for model.embed_tokens.weight: copying a param with shape torch.Size([32000, 4096]) from checkpoint, the shape in current model is torch.Size([4000, 4096]).
	size mismatch for lm_head.weight: copying a param with shape torch.Size([32000, 4096]) from checkpoint, the shape in current model is torch.Size([4000, 4096]).
	You may consider adding `ignore_mismatched_sizes=True` in the model `from_pretrained` method.
```

This feels that there is some error when consolidating the weights from the sharded ones since 32000/8=4000.

5cp:
Hi @philschmid 

I was able to reproduce this. The `vocab_size` in the fine-tuned model's config.json is incorrectly set to 4000 and is not adjusted during consolidation. I manually changed this to 32000 and was able to load the checkpoint.

michaelbenayoun:
Closing since I think this is solved in #378

Issue #367, opened 2023-12-07
Neuron `consolidate` is not creating `model.safetensors.index.json`


philschmid:
When training a model with TP, weights are saved as sharded TP weights. This means we need to consolidate/convert them into regular weights using `optimum neuron consolidate` command. But the command currently is not creating the correct structure. It only saves the `model-00001-of-00002.safetensors` and misses the `model.safetensors.index.json`. 

Additionally we should also copy the `config.json` to the target directory if available. Since it is also always needed to run the model. 

![Bildschirmfoto 2023-12-07 um 13 55 28](https://github.com/huggingface/optimum-neuron/assets/32632186/25f3bfb5-5394-443c-a1fe-a73c0153f4cb)

philschmid:
## Addition: 

It seems the `model.safetensors.index.json` is being created but at the wrong level, when using. 
```bash
optimum-cli neuron consolidate dolly_llama_sharded/tensor_parallel_shards dolly_llama_sharded
```

the `model.safetensors.index.json` is added at `../dolly_llama_sharded` but should be in `dolly_llama_sharded` next to the weights. 

![Bildschirmfoto 2023-12-07 um 14 05 34](https://github.com/huggingface/optimum-neuron/assets/32632186/8f7da7e3-cb6b-4bab-888b-36b1eecc488e)

5cp:
This small change appears to resolve the issue
```diff
diff --git a/optimum/neuron/distributed/checkpointing.py b/optimum/neuron/distributed/checkpointing.py
index 58a11f8..9fcb986 100644
--- a/optimum/neuron/distributed/checkpointing.py
+++ b/optimum/neuron/distributed/checkpointing.py
@@ -15,6 +15,7 @@
 """Functions handling checkpointing under parallel settings."""

 import json
+import os
 from pathlib import Path
 from typing import Dict, Literal, Union

@@ -95,6 +96,6 @@ def consolidate_tensor_parallel_checkpoints_to_unified_checkpoint(
             torch.save(shard, output_dir / shard_file)
     if index is not None:
         save_index_file = SAFE_WEIGHTS_INDEX_NAME if save_format == "safetensors" else WEIGHTS_INDEX_NAME
-        with open(save_index_file, "w") as fp:
+        with open(os.path.join(output_dir, save_index_file), "w") as fp:
             content = json.dumps(index, indent=2, sort_keys=True) + "\n"
             fp.write(content)
```

PR #378, opened 2023-12-14
Fix checkpoint saving and consolidation for TP

michaelbenayoun:
As per title.

Fixes: #373 #368 #367


dacorvo (approved):

dacorvo:
These are important fixes: thank you @michaelbenayoun . Let's merge this quickly.

PR #387, opened 2023-12-20
Use official serialization API instead of beta.

aws-yishanm:
Update the serialization portion of the code to use the official API. This api now checks the validity of the HLOs and NEFFs (meaning that a recompilation may be necessary). Please use neuron_model.save('./nameofdir') to save the new artifacts.


dacorvo (approved):
Thank you for this pull-request. Some of the tests are failing because pull-requests from forks do not have the correct credentials, but I tested locally and it works as expected.

PR #393, opened 2023-12-22
fix(decoder): remove call to beta serialization api

dacorvo:
This is the continuation of pull-request #387, also removing the call to the experimental serialization API when saving the model.

### DIFF
--- a/optimum/neuron/modeling_decoder.py
+++ b/optimum/neuron/modeling_decoder.py
@@ -246,7 +246,7 @@ class NeuronDecoderModel(OptimizedModel):
 
         if src_compiled_path is None:
             # The compiled model has never been serialized: do it now
-            self.model._save_compiled_artifacts(dst_compiled_path)
+            self.model.save(dst_compiled_path)
         else:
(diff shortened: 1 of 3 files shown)


optimum/neuron/modeling_decoder.py
```diff
@@ -246,7 +246,7 @@ def _save_pretrained(self, save_directory: Union[str, Path]):
 
         if src_compiled_path is None:
             # The compiled model has never been serialized: do it now
-            self.model._save_compiled_artifacts(dst_compiled_path)
+            self.model.save(dst_compiled_path)
```
JingyaHuang, line 249: Do all these stable APIs also exist in 2.15.2? Otherwise do we need to bump the transformers-neuronx's version in the setup?

JingyaHuang (approved):
LGTM, thanks @dacorvo, just left a small nit.

optimum/neuron/modeling_decoder.py
dacorvo: No they are already there (introduced in 2.15.0). I had not seen them, but the doc had indeed been updated.

PR #398, opened 2024-01-08
Use AWS Neuron SDK 2.16 packages

dacorvo:
This bumps the version of all AWS neuron packages referenced in setup.py and text-generation-inference Dockerfile. 

The latest AWS Neuron SDK pip packages are not compatible with the old drivers in the legacy CI AMI.

The runners need to be updated to use the latest DLAMI from AWS that corresponds to the AWS Neuron SDK 2.16: ami-0fbea04d7389bcd4e.

Runners status:
- inf2: DONE
- inf1: DONE
- trn1: DONE

All tests are OK, except for the `trainium` common tests that are failing;
- three test_runner tests (summarization, translation, image_classification),
- one staging test on optimum cache.

I would suggest to merge that pull-request so that @michaelbenayoun can further investigate the failing tests.


.github/workflows/test_inf2.yml
```diff
@@ -33,7 +35,23 @@ jobs:
           python -m pip install -U pip
           python -m pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
           python -m pip install .[neuronx,tests]
-      - name: Run tests
+      - name: Run CLI tests
```
michaelbenayoun, line 38: Is it only CLI tests?

optimum/neuron/utils/runner.py
```diff
@@ -213,7 +214,7 @@ def __init__(
         self.use_venv = use_venv
         self.should_install_requirements = install_requirements
         self.venv_dir = TemporaryDirectory()
-        self.python_name = "python"
+        self.python_name = sys.executable
         self.pip_name = "pip"
         self.torchrun_name = "torchrun"
```
michaelbenayoun, line 219: This case still works considering that you needed to change the name for `self.python_name` and we either use `self.python_name` or `self.torchrun_name`?

tests/exporters/test_export.py
```diff
@@ -80,8 +81,9 @@ def _get_models_to_test(export_models_dict: Dict, random_pick: Optional[int] = N
                     (f"{model_type}_{task}", model_type, model_name, task, neuron_config_constructor)
                 )
 
+    random_pick = os.environ.get("MAX_EXPORT_TEST_COMBINATIONS", None)
     if random_pick is not None:
-        return sorted(random.choices(models_to_test, k=random_pick))
+        return sorted(random.choices(models_to_test, k=int(random_pick)))
```
JingyaHuang, line 86: Are the CIs running all tests now? Did you set `MAX_EXPORT_TEST_COMBINATIONS`? I wonder if it would take too long.

michaelbenayoun (approved):

JingyaHuang (approved):
LGTM! Thanks for fixing the CIs!

tests/exporters/test_export.py
dacorvo: I used it while debugging the tests, but eventually I ran all of them because if I set it to just 1 it runs just ... one test.
We would need to figure out a compromise for these tests. Maybe have a few quick steps with just one test to catch early failures, then the whole list that takes tens of minutes.

PR #402, opened 2024-01-09
[Pre Neuron Inf Cache system]Support neff/weights decoupling

JingyaHuang:
With [`inline_weights_to_neff`](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/frameworks/torch/torch-neuronx/api-reference-guide/inference/api-torch-neuronx-trace.html#torch-neuronx-trace-api) argument, we are now able to **decouple the weights and neff graph during the compilation**. It means that we can set up a **caching system for reusing shared compiled neff graphs and load with weights from different checkpoints** during the inference. This could largely **save the compilation time** which could take up to hours.

- [x] Support `inline_weights_to_neff` in Neuron exporter
- [x] Replace weights function that works for Optimum wrapped models
- [x] Support replacing weights in the modeling
- [x] Tests

[Next step]
Initial caching for inference of encoder models
* Set up hashing for inference
* Set up caching mechanism based on current caching of models with independent neff files  (`optimum-cli` for creating public/private cache)
* Checker & replace weights on the modeling `from_pretrained` when export=True and cache exist.
    - Exist (local / remote) -> load precompiled neff + replace weights with current checkpoint
    - Non-exist -> Compile from scratch and maybe cache


optimum/commands/export/neuronx.py
```diff
@@ -85,6 +85,11 @@ def parse_args_neuronx(parser: "ArgumentParser"):
         type=Path,
         help="Path indicating the directory where to store intermediary files generated by Neuronx compiler.",
     )
+    optional_group.add_argument(
+        "--disable-weights-neff-inline",
```
michaelbenayoun, line 89:
```suggestion
        "--weights-neff-inline",
```

michaelbenayoun: According to the help string it seems that when set you want to actually inline weights no?

optimum/neuron/modeling_base.py
```diff
@@ -103,7 +109,13 @@ def load_model(path: Union[str, Path]) -> torch.jit._script.ScriptModule:
             path = Path(path)
 
         if path.is_file():
-            return torch.jit.load(path)
+            model = torch.jit.load(path)
+            return model
+
+    def replace_wights(self, weights: Union[Dict[str, torch.Tensor], torch.nn.Module] = None):
```
michaelbenayoun, line 115:
```suggestion
    def replace_weights(self, weights: Optional[Union[Dict[str, torch.Tensor], torch.nn.Module]] = None):
```

optimum/neuron/utils/misc.py
```diff
@@ -508,3 +511,38 @@ def download_checkpoints_in_cache(
...
+    module_paths = [module_path for module_path in module_paths if module_path != ""]
+
+    for module_path in module_paths:
+        if len(re.findall("\w\d+", module_path)) > 0:
+            continue
+        else:
+            model.weights._c.setattr(module_path, weights[module_path.replace(prefix + "->", "").replace("->", ".")])
+
+
+def check_if_weights_replacable(config: "PretrainedConfig", weights: Union[Dict[str, torch.Tensor], torch.nn.Module]):
```
michaelbenayoun, line 541:
```suggestion
def check_if_weights_replacable(config: "PretrainedConfig", weights: Optional[Union[Dict[str, torch.Tensor], torch.nn.Module]]):
```

optimum/commands/export/neuronx.py
JingyaHuang: I will change a bit the help to make it clear.
In the beginning, I wanted to separate weights / neff by default, then I hit some edge cases eg. some models arch are not well supported, dynamic batching not yet supported but will come in the future. So I prefer to inline weight by default for now.

tests/inference/test_modeling.py
```diff
@@ -136,6 +136,26 @@ def test_save_compiler_intermediary_files(self):
...
+                export=True,
+                compiler_workdir=save_path,
+                inline_weights_to_neff=False,
+                **self.STATIC_INPUTS_SHAPES,
+            )
+            self.assertFalse(neuron_model.config.neuron.get("inline_weights_to_neff"))
+
+            # replace weights
+            model = AutoModelForSequenceClassification.from_pretrained(self.MODEL_ID)
+            neuron_model.replace_wights(weights=model)
```
dacorvo, line 155:
```suggestion
            neuron_model.replace_weights(weights=model)
```

dacorvo (commented):
LGTM, but you have a typo in your tests.

tests/inference/test_modeling.py
JingyaHuang: And that's why the CI failed this time, thanks for helping me debug. Guess I deleted a letter randomly while fixing the doc string 😅.

dacorvo (approved):

PR #403, opened 2024-01-10
Bump hf libraries versions

JingyaHuang:
To solve the tokenizers vs huggingface_hub version conflicts. @dacorvo

### DIFF
--- a/.github/workflows/test_inf2.yml
+++ b/.github/workflows/test_inf2.yml
@@ -38,20 +38,20 @@ jobs:
       - name: Run CLI tests
         run: |
           source aws_neuron_venv_pytorch/bin/activate
-          HF_TOKEN_OPTIMUM_NEURON_CI=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/cli
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/cli
       - name: Run exporters tests
         run: |
           source aws_neuron_venv_pytorch/bin/activate
-          HF_TOKEN_OPTIMUM_NEURON_CI=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/exporters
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -13,10 +13,10 @@
 
 
 INSTALL_REQUIRES = [
-    "transformers == 4.35.0",
+    "transformers == 4.36.2",
     "accelerate == 0.23.0",
     "optimum >= 1.14.0",
-    "huggingface_hub >= 0.14.0",
+    "huggingface_hub >= 0.19.4",
```
michaelbenayoun, line 19:
```suggestion
    "huggingface_hub >= 0.20.1",
```
I think it'll be needed for us to use the `huggingface_hub.get_token()` function, which is the recommended way to ge the value of the token from now on.

.github/workflows/test_inf2.yml
```diff
@@ -38,20 +38,20 @@ jobs:
       - name: Run CLI tests
         run: |
           source aws_neuron_venv_pytorch/bin/activate
-          HF_TOKEN_OPTIMUM_NEURON_CI=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/cli
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/cli
       - name: Run exporters tests
         run: |
           source aws_neuron_venv_pytorch/bin/activate
-          HF_TOKEN_OPTIMUM_NEURON_CI=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/exporters
```
michaelbenayoun, line 45: Is this environment variable only used in the two workflows you changed?

JingyaHuang: I saw Trainium tests are using it as well, do you want me to change trainium CIs as well?

michaelbenayoun: Let's do it in another PR.

michaelbenayoun (approved):
The CI for training do not pass, I think that's ok. We can iterate later. I think that #279 will solve many of these issues.

PR #405, opened 2024-01-12
Skip pushing if the user does not have write access to the cache repo

### LINKED ISSUE #391: Trainer always wants to push to cache and let training fail
I noticed that in the last version (0.0.16), all the training fails if you don't add the training argument `skip_cache_push True`. Without the argument the trainer always wants to push to the cache, this raises and error and failts the training with

```bash
ValueError: Could not push the cached model to aws-neuron/optimum-neuron-cache because you do not have write access to this repo.
```

michaelbenayoun:
As per title.

Fixes #391


dacorvo (approved):
LGTM, thanks !

PR #413, opened 2024-01-17
Do not upload NeuronModelForCausalLM weights when they can be reconstructed from the hub

dacorvo:
When serializing a model, the checkpoint files (i.e. the original
weights formatted for neuron) and the compiled artifacts are stored
under two folders in the model directory.
This modifies the `NeuronForCausalLM.push_to_hub` method to exclude
the folder containing the checkpoint files if they can be instead
reconstructed from a model on the hub.

Note that it is still possible to upload checkpoint files by using the
`huggingface_cli` directly on the saved model folder instead of
push_to_hub.

To allow `optimum-neuron` to fetch and reconstruct the checkpoint when
fetchin a neuron model from the hub, the original repository that was
used to export the model and its revision are stored in the neuron config.

Example model here:

https://huggingface.co/dacorvo/llama-2-7b-chat-hf-neuronx-bs1-seq2048-no-checkpoint/tree/main

```json

"neuron": {
  "auto_cast_type": "fp16",
  "batch_size": 1,
  "checkpoint_id": "meta-llama/Llama-2-7b-chat-hf",
  "checkpoint_revision": "c1b0db933684edbfe29a06fa47eb19cc48025e93",
  "compiler_type": "neuronx-cc",
  "compiler_version": "2.12.54.0+f631c2365",
  "num_cores": 2,
  "sequence_length": 2048,
  "task": "text-generation"
}

```

### DIFF
--- a/tests/generation/test_hub.py
+++ b/tests/generation/test_hub.py
@@ -32,22 +32,26 @@ def test_model_from_hub():
     check_neuron_model(model, batch_size=16, sequence_length=512, num_cores=2, auto_cast_type="fp32")
 
 
-@is_inferentia_test
-@requires_neuronx
-def test_push_to_hub(neuron_decoder_path, neuron_push_decoder_id):
-    model = NeuronModelForCausalLM.from_pretrained(neuron_decoder_path)
-    model.push_to_hub(neuron_decoder_path, neuron_push_decoder_id, use_auth_token=TOKEN, endpoint=ENDPOINT_STAGING)
+def _test_push_to_hub(model, model_path, repo_id):
+    model.push_to_hub(model_path, repo_id, use_auth_token=TOKEN, endpoint=ENDPOINT_STAGING)
     api = HfApi(endpoint=ENDPOINT_STAGING, token=TOKEN)
     try:
-        hub_files_info = api.list_files_info(neuron_push_decoder_id)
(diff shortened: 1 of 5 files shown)


michaelbenayoun (approved):
LGTM

tests/generation/test_hub.py
```diff
@@ -25,29 +27,45 @@
...
 @requires_neuronx
-def test_model_from_hub():
-    model = NeuronModelForCausalLM.from_pretrained(
-        "dacorvo/tiny-random-gpt2-neuronx", revision="1b3456cf877cc42c053ee8464f1067021eccde4b"
-    )
+@pytest.mark.parametrize(
+    "model_id, revision",
+    [
+        ["dacorvo/tiny-random-gpt2-neuronx", "1b3456cf877cc42c053ee8464f1067021eccde4b"],
+        ["dacorvo/tiny-random-gpt2-neuronx-no-checkpoint", "78eb2313ab7e149bbc22ff32257db93ba09e3033"],
```
JingyaHuang, line 34: For the test with no checkpoint, shall we check there is no checkpoint after push to hub?

JingyaHuang (approved):
LGTM!

tests/generation/test_hub.py
dacorvo: It is a download here: the check is done in the other test.

PR #422, opened 2024-01-18
Allow exporting decoder models using optimum-cli

dacorvo:
A long awaited feature ... it is now possible to export decoder models just like any other models using the optimum-cli.

### DIFF
--- a/docs/source/guides/export_model.mdx
+++ b/docs/source/guides/export_model.mdx
@@ -25,7 +25,7 @@ optimum-cli export neuron \
   --model bert-base-uncased \
   --sequence_length 128 \
   --batch_size 1 \
-  bert_neuron/  
+  bert_neuron/
 ```
 
 Check out the help for more options:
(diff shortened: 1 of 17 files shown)


philschmid (approved):
Should be add a test for it?

dacorvo:
> Should be add a test for it?

I thought I did add one, but I might have dreamed it, or I lost it in my rebase ...

dacorvo:
> > Should be add a test for it?
> 
> I thought I did add one, but I might have dreamed it, or I lost it in my rebase ...

Done.

docs/source/guides/export_model.mdx
```diff
@@ -49,18 +49,18 @@ To understand a little bit more about the compilation, here are general steps ex
 </Tip>
 
 Although pre-compilation avoids overhead during the inference, traced Neuron module has some limitations:
-* Traced Neuron module will be static, which requires fixed input shapes and data types used passed during the compilation. As the model won't be dynamically recompiled, the inference will fail if any of the above conditions change. 
+* Traced Neuron module will be static, which requires fixed input shapes and data types used passed during the compilation. As the model won't be dynamically recompiled, the inference will fail if any of the above conditions change.
```
michaelbenayoun, line 52:
```suggestion
* Traced Neuron module will be static, which requires fixed input shapes and data types used during the compilation. As the model won't be dynamically recompiled, the inference will fail if any of the above conditions change.
```

optimum/neuron/modeling_decoder.py
```diff
@@ -208,10 +208,15 @@ def _export(
...
 
+        if batch_size is None:
+            batch_size = 1
         # If the sequence_length was not specified, deduce it from the model configuration
         if sequence_length is None:
             # Note: for older models, max_position_embeddings is an alias for n_positions
             sequence_length = config.max_position_embeddings
+        if num_cores is None:
+            # Use all available cores
+            num_cores = len(os.listdir("/sys/class/neuron_device/")) * 2
```
michaelbenayoun, line 219: This will fail if the directory does not exist. Are we sure that this code can only be executed on a machine that has neuron devices?

michaelbenayoun (approved):

optimum/neuron/modeling_decoder.py
dacorvo: I added a check on top of the function. It is not possible to instantiate the `transformers-neuronx` model otherwise.

PR #438, opened 2024-01-24
Correctly identify special tokens during generation

### LINKED ISSUE #437: TGI for Inf2 does not special=True in the last token of a stream
As of https://github.com/huggingface/optimum-neuron/commit/3b3afa4dad427c0d43d3c799431a0e2619ac636a, when calling a TGI endpoint for Inf2, it does not set `special=True` in the last token of a stream. As a result, callers have to guess the end of the stream based on an assumption of what the last token will be.

Here is an example last token I get from the TGI endpoint:

```
data:{"token":{"id":2,"text":"</s>","logprob":0.0,"special":false},"generated_text":"/inf2 How are you?\n\nI'm doing well, thanks for asking! I'm always happy to chat with you and help with any questions or topics you'd like to discuss. How about we talk about something you're interested in or passionate about?</s>","details":null}

```

Note that this is not the case for regular TGI.

dacorvo:
The special tokens were never identified as such. This fixes #437.

### DIFF
--- a/text-generation-inference/server/text_generation_server/generator.py
+++ b/text-generation-inference/server/text_generation_server/generator.py
@@ -459,7 +459,7 @@ class NeuronGenerator(Generator):
                     token_id=next_token,
                     token_logprob=None,
                     token_text=next_token_text,
-                    token_is_special=(next_token in [self.special_tokens]),
+                    token_is_special=(next_token in self.special_tokens),
                     generated_text=generated_text,
                 )
             )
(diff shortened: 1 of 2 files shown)


philschmid (approved):

michaelbenayoun (approved):

text-generation-inference/server/text_generation_server/generator.py
```diff
@@ -291,7 +291,7 @@ def __init__(
         tokenizer.pad_token_id = tokenizer.eos_token_id
         tokenizer.padding_side = "left"
         self.tokenizer = tokenizer
-        self.special_tokens = [self.tokenizer.eos_token_id, self.tokenizer.pad_token_id]
+        self.special_tokens = self.tokenizer.all_special_ids
```
mfuntowicz, line 294: Do we want to use a `set()` here to use optimized lookup?

PR #449, opened 2024-01-30
AWS Neuron SDK 2.16.1: update neuronxcc

dacorvo:
The Neuron X compiler has been updated with the latest release of the SDK.

This bumps the version used by `optimum-neuron` and adds a new `__sdk_version__` variable to reflect on which AWS Neuron SDK version it is based.

### DIFF
--- a/optimum/neuron/modeling_decoder.py
+++ b/optimum/neuron/modeling_decoder.py
@@ -123,14 +123,11 @@ class NeuronDecoderModel(OptimizedModel):
             neuronx_model.load(compiled_dir)
 
         # Compile the Neuron model (if present compiled artifacts will be reloaded instead of compiled)
-        neuron_cc_flags = os.environ.get("NEURON_CC_FLAGS", "")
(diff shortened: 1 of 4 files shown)


optimum/neuron/modeling_decoder.py
```diff
@@ -123,14 +123,11 @@ def __init__(
             neuronx_model.load(compiled_dir)
 
         # Compile the Neuron model (if present compiled artifacts will be reloaded instead of compiled)
-        neuron_cc_flags = os.environ.get("NEURON_CC_FLAGS", "")
-        os.environ["NEURON_CC_FLAGS"] = neuron_cc_flags + " --model-type=transformer"
```
philschmid, line 127: Is that no longer needed?

philschmid (approved):
Great idea on adding the sdk_version

optimum/neuron/modeling_decoder.py
dacorvo: No, it is now the default.

PR #455, opened 2024-02-01
Split inferentia CI workflows

dacorvo:
These workflows take too long: they need to be split.

@JingyaHuang for some reason, one of the stable diffusion test is failing repeatedly, while it was not before ...

### DIFF
--- a/.github/workflows/test_inf1_export.yml
+++ b/.github/workflows/test_inf1_export.yml
@@ -1,4 +1,4 @@
-name: Optimum neuron / Test INF2
+name: Optimum neuron / Test INF1 export
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


.github/workflows/test_inf2.yml
```diff
@@ -47,15 +47,3 @@ jobs:
         run: |
           source aws_neuron_venv_pytorch/bin/activate
           HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m is_inferentia_test tests/generation
-      - name: Run exporters tests
```
JingyaHuang, line 50: could we name the file `test_inf2_common` and job name with `name: Optimum neuron / Test INF2 Common`?

JingyaHuang (commented):
LGTM, just left one nit on the naming, thanks for the PR!

.github/workflows/test_inf2.yml
dacorvo: I agree test_inf2.yml is a poor name, but these are not really common tests: all decoder tests are included. I was about to call it test_inf2_decoder, but there are some CLI tests also, so I kept the name ...

JingyaHuang: Oki doki 😤, then we keep it this way for now.

JingyaHuang (approved):

PR #462, opened 2024-02-05
Fix path, update versions

shub-kris:


.github/workflows/build-ami.yml
```diff
@@ -29,18 +29,22 @@ jobs:
...
+            uses: actions/checkout@v4
             with:
               # If the workflow is triggered manually or by schedule, uses the tag, otherwise uses the current branch  
               ref: ${{ github.event.inputs.tag || github.ref }}
           
           - name: Setup Packer
             uses: hashicorp/setup-packer@main
+            id: setup-packer
+            with:
+              version: ${{ env.PACKER_VERSION }}
```
philschmid, line 44: We can fix it here no? Or is something which changes often?

philschmid (approved):

.github/workflows/build-ami.yml
shub-kris: It is the version of the packer and not the action, so it changes often. 
https://github.com/hashicorp/packer/tags. So something I will also day to fix until we run into some kind of warnings as github actions was complaining today

shub-kris: @philschmid good to merge?

philschmid (approved):
LGTM! can you try an build a version v0.0.18?

shub-kris:
> LGTM! can you try an build a version v0.0.18?

It will never work with versions that doesn't have the path: `infrastructure/ami`. It will only work with the branches which will be now branched out from main.

philschmid:
> > LGTM! can you try an build a version v0.0.18?
> 
> It will never work with versions that doesn't have the path: `infrastructure/ami`. And the branch v0.0.18 won't work because it is branched out from a previous timestamp. So the packer files aren't there in v0.0.18 or any other branch before. https://github.com/huggingface/optimum-neuron/tree/v0.0.18-release
> 
> It will only work with the branches which will be now branched out from main.

Okay but what if, e.g. make release 0.0.19 and noticed the AMI pipeline fails since there was some change in the base AMI or we need to update a version or another file in `infrastructure/ami`? 
How can we then release 0.0.19. From what you shared we would not be able to release this version since it would check out the AMI code from the release which would not have the fixed or changes.

shub-kris:
@philschmid with this change, if any change is made in the repo and a PR is made to the main, the workflow will be triggered and the ami will be built. 

```
run: packer validate -var "optimum_version=${{ github.event.inputs.tag || github.head_ref || github.ref_name }}" -var "region=${{ env.AWS_REGION }}" hcl2-files

``` 
When someone manually specifies it will use that tag and during pull request it will use the name of the source branch (gtithub.head_ref is only set during PR) and if pushed to main it will use github.ref_name which will be main.

philschmid:
> So from now on whenever we will make the release, we would be sure that the build-ami was successful.

Thats not a dependency we should have! The AMI is not directly related to the code. IMO we should be able to build AMIs for a specific optimum version, but the AMI code doesn't need to in the library code. 

Why can't we use the `inputs.tag` and pass that into the packer script and based on this either install from pypi or install from git. We could remove all the `with` in the checkout here
```yaml
            uses: actions/checkout@v4
            with:
              # If the workflow is triggered manually or by schedule, uses the tag, otherwise uses the current branch  
              ref: ${{ github.event.inputs.tag || github.ref }}
```
this would then always check out the "branch/event" which triggered the workflow and we pass the `inputs.tag` into ` packer validate -var "optimum_version=`

And then in the packer bash we simply do 
```bash
# Check if OPTIMUM_VERSION starts with 'v'
if [[ $OPTIMUM_VERSION == v* ]]; then
    # If it starts with 'v', install that version using pip
    pip install optimum-neuron==$OPTIMUM_VERSION
else
    pip install git+https://github.com/huggingface/optimum-neuron.git@$OPTIMUM_VERSION
fi
```
^ you have to validate it works

This will allow us to build an AMI based on the "dispatch" + "tag" and if it is triggered by action, e.g. pull request we us that branch?

shub-kris:
@philschmid I removed all the checkout code. I agree we should use the branch that triggered the workflow. 

We don't have to do this logic:

```
# Check if OPTIMUM_VERSION starts with 'v'
if [[ $OPTIMUM_VERSION == v* ]]; then
    # If it starts with 'v', install that version using pip
    pip install optimum-neuron==${OPTIMUM_VERSION#v}
else
    pip install git+https://github.com/huggingface/optimum-neuron.git@$OPTIMUM_VERSION
fi

```

As this thing is always taken care by: 
```
optimum_version=${{ github.event.inputs.tag || github.head_ref || github.ref_name }}"
```
It will always be the value which triggers the workflow. 


Also as we are copying some files from the git repo to the local machine while building the image the pip logic won't copy the files for that we will have to clone the repo anyhow.

shub-kris:
Once merged you will also be able to build with `v0.0.18`, I tested it with the current workflow

philschmid (approved):

PR #463, opened 2024-02-05
Add issue and PR templates & build optimum env cli for Neuron

JingyaHuang:
as per title. As there are community members asking for support for some model arch, I would like to complete the templates to make it easier for them.

------

Ouput of the cli env:

```bash
optimum-cli env
```

```
Copy-and-paste the text below in your GitHub issue:


Platform:

- Platform: Linux-5.15.0-1052-aws-x86_64-with-glibc2.29
- Python version: 3.8.10


Python packages:

- `optimum-neuron` version: 0.0.19.dev0
- `neuron-sdk` version: 2.16.1
- `optimum` version: 1.16.2
- `transformers` version: 4.36.2
- `huggingface_hub` version: 0.20.3
- `torch` version: 1.13.1+cu117
- `aws-neuronx-runtime-discovery` version: 2.9
- `libneuronxla` version: 0.5.669
- `neuronx-cc` version: 2.12.68.0+4480452af
- `neuronx-distributed` version: 0.6.0
- `neuronx-hwm` version: 2.12.0.0+422c9037c
- `torch-neuronx` version: 1.13.1.1.13.0
- `torch-xla` version: 1.13.1+torchneurond
- `transformers-neuronx` version: 0.9.474


Neuron Driver:


WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

aws-neuronx-collectives/unknown,now 2.19.7.0-530fb3064 amd64 [installed]
aws-neuronx-dkms/unknown,now 2.15.9.0 amd64 [installed]
aws-neuronx-runtime-lib/unknown,now 2.19.5.0-97e2d271b amd64 [installed]
aws-neuronx-tools/unknown,now 2.16.1.0 amd64 [installed]
```

### DIFF
--- a/optimum/commands/env.py
+++ b/optimum/commands/env.py
@@ -43,11 +43,12 @@ class EnvironmentCommand(BaseOptimumCLICommand):
                 except Exception:
                     num_version = "NA"
                 info[f"`{pkg}` version"] = num_version
+        return info
 
     @staticmethod
     def print_apt_pkgs():
-        info = subprocess.run("apt list --installed | grep aws-neuron", capture_output=True, shell=True)
-        pkgs_list = info.stdout.decode().split("\n")
+        info = subprocess.getoutput("apt list --installed | grep aws-neuron")
+        pkgs_list = info.split("\n")[3:]
         for pkg in pkgs_list:
             print(pkg)
 
@@ -87,7 +88,7 @@ class EnvironmentCommand(BaseOptimumCLICommand):
         else:
             neuron_python_pkgs = None
 
-        self.get_pip_pkgs_version(neuron_python_pkgs, info)
+        info = self.get_pip_pkgs_version(neuron_python_pkgs, info)
 
         print("\nCopy-and-paste the text below in your GitHub issue:\n")
         print("\nPlatform:\n")
--- a/optimum/commands/env.py
+++ b/optimum/commands/env.py
@@ -1,0 +1,98 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


dacorvo (approved):
Nice ! Thanks !

optimum/commands/env.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @staticmethod
+    def get_pip_pkgs_version(pkg_list: Optional[List], info: Dict):
+        if pkg_list is not None:
+            for pkg in pkg_list:
+                try:
+                    num_version = get_distribution(pkg).version
+                except Exception:
+                    num_version = "NA"
+                info[f"`{pkg}` version"] = num_version
```
michaelbenayoun, line 45: Why is it updating the `info` dictionary instead of returning the value as "regular" function?

optimum/commands/env.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @staticmethod
+    def print_apt_pkgs():
+        info = subprocess.run("apt list --installed | grep aws-neuron", capture_output=True, shell=True)
```
michaelbenayoun, line 49: nit: Would not use `shell=True` unless absolutely necessary.

JingyaHuang: My previous concern was without `shell=True`, it captures only staging outputs, something like:
```
Neuron Driver:

Listing...
```
instead of the neuron driver information that I want to show... There is `subprocess.getoutput` but under the hood, it still can only execute the cmd in a shell as well. 

I just put a workaround to avoid shell=True.

optimum/commands/env.py
JingyaHuang: The function returns info now.

michaelbenayoun (approved):

PR #472, opened 2024-02-09
Integrate new cache system for training

michaelbenayoun:
# What does this PR do?

Integrate the native cache system to training.

To do in following PRs:

- Update the documentation
- Remove former cache util code that is not useful anymore, this includes the functions and the tests
- Unify `optimum/neuron/utils/cache_utils.py` and `optimum/neuron/utils/hub_neuronx_cache.py`.

cc @5cp


optimum/neuron/utils/neuron_cc_wrapper
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python3
```
5cp, line 1: This assumes a system-wide install of Optimum Neuron, which isn't always the case.

Changing shebang to `#!/usr/bin/env python3` worked for my setup.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -274,6 +277,44 @@ def hf_create_compile_cache(cache_url):
...
+            shutil.move(src, dst)
+        if restore:
+            src = main_dir / "neuron_cc_wrapper_backup"
+            dst = main_dir / "neuron_cc_wrapper"
+        else:
+            src = Path(__file__).parent / "neuron_cc_wrapper"
+            dst = main_dir / "neuron_cc_wrapper"
+        shutil.copy(src, dst)
+
+    try:
```
5cp, line 307: I think we should have an `except` clause here. I ran into an issue where the `patch()` failed and my job was hanging at the subsequent `xm.rendezvous()` with no indication of what went wrong.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -231,22 +233,47 @@ def hash(self):
 
 
 REGISTRY_FOLDER = f"0_REGISTRY/{__version__}"
```
dacorvo, line 235: We might also use something like REGISTRY/version/mode. For now let's leave it as it is.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -282,11 +349,16 @@ def synchronize_hub_cache(cache_repo_id: Optional[str] = None):
         repo_id (`Optional[str]`, default to None):
             The id of the HuggingFace cache repository, in the form 'org|user/name'.
     """
+    # Not pushing anything if neuron parallel compile.
```
dacorvo, line 352: I think the check should rather be done in the module calling this. The cache should not know anything about parallel compile.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -274,6 +301,46 @@ def hf_create_compile_cache(cache_url):
...
+    """
+    Patches the `neuron_cc_wrapper` file to force it use our own version of it which essentially makes sure that it
+    uses our caching system.
+    """
+
+    import torch_xla.core.xla_model as xm
+
+    def patch(restore: bool = False):
+        path = os.environ["PATH"]
+        main_dir = Path(path.split(":")[0])
```
dacorvo, line 317: This won't work if main_dir is not writable, but with our deployments it seems the first item in the PATH is $HOME/.local/bin.  Let's leave that issue for later and cross fingers !

dacorvo (commented):
One minor comment: if you could move the check for parallel compile in the parallel compile related code that would be awesome.
Otherwise LGTM.

optimum/neuron/utils/hub_neuronx_cache.py
michaelbenayoun: Want to do that now? It's not a difficult change to make.

optimum/neuron/utils/hub_neuronx_cache.py
michaelbenayoun: I was thinking the same thing, will make the change.

optimum/neuron/utils/hub_neuronx_cache.py
michaelbenayoun: Ok, just got an idea to make things much cleaner instead of patching an existing file and everyting. What about creating a temporary directory and copying the executable there?

optimum/neuron/utils/hub_neuronx_cache.py
michaelbenayoun: Ok, I have changed it. If the user has a compiler version under 2.17, it will use the "old" naming convention for inference, otherwise it will use the new one (the one you suggested)

dacorvo (approved):
LGTM, thanks for the pull-request: a brand new training cache !

PR #481, opened 2024-02-15
[documentation] Add Container overview page.

philschmid:
# What does this PR do?

Adds a new documentation page with available containers.

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -5,6 +5,8 @@
     title: Installation
   - local: quickstart
     title: Quickstart
+  - local: containers
+    title: Optimum Containers
   - sections:
     - local: tutorials/overview
       title: Overview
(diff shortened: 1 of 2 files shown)


dacorvo (approved):
The TGI 0.0.18 version should be available but I think it does not work because of the change from /data to /tmp.

dacorvo:
> > The TGI 0.0.18 version should be available but I think it does not work because of the change from /data to /tmp.
> 
> /tmp is what we need. Is it /data now?
> 
> FYI ETA for 0.0.18 is 2/16 and SDK 2/21

No, it is /tmp as requested, but now their sanity test is failing ... I think /tmp is not writable ...

dacorvo:
The TGI 0.0.18 container has been fixed: https://github.com/awslabs/llm-hosting-container/commit/222cd7dde632578aec27c6d2cc93ddcc4f1dfdd4

PR #488, opened 2024-02-15
Add support for transformers-neuronx continuous batching

dacorvo:
# What does this PR do?

This adds support for `transformers-neuronx` continuous batching.

This also activates automatically the continuous batching feature for any model that supports it as soon as the batch size is > 1.

This allows to workaround the fact that static batching is broken in `transformers-neuronx` for these models: https://github.com/aws-neuron/transformers-neuronx/issues/79.

But this is also a game changer for NeuronX TGI, as it unlocks:
- higher batch sizes for `Llama` and `Mistral` models thanks to `transformers-neuronx` "continuous batching" graphs that share KV cache between nodes,
- unlimited concurrent requests thanks to TGI 1.4.1 `max-batch-size` parameter.

### DIFF
--- a/optimum/exporters/neuron/base.py
+++ b/optimum/exporters/neuron/base.py
@@ -379,7 +379,7 @@ class NeuronDecoderConfig(NeuronConfig):
         be passed to export the model,
     - NEURONX_CLASS (`str`) -- the name of the transformers-neuronx class to instantiate for the model.
     It is a full class name defined relatively to the transformers-neuronx module, e.g. `gpt2.model.GPT2ForSampling`
-    [`~optimum.utils.DummyInputGenerator`] specifying how to create dummy inputs.
+    - CONTINUOUS_BATCHING (`bool`) -- Whether the model supports continuous batching or not.
 
     The NEURONX_CLASS must always be defined in each model configuration.
 
@@ -389,6 +389,7 @@ class NeuronDecoderConfig(NeuronConfig):
 
     INPUT_ARGS = ("batch_size", "sequence_length")
     NEURONX_CLASS = None
+    CONTINUOUS_BATCHING = False
 
     def __init__(self, task: str):
         if not is_transformers_neuronx_available():
@@ -404,3 +405,7 @@ class NeuronDecoderConfig(NeuronConfig):
     @property
     def neuronx_class(self):
         return self._neuronx_class
+
+    @property
+    def continuous_batching(self):
+        return self.CONTINUOUS_BATCHING
--- a/optimum/exporters/neuron/base.py
+++ b/optimum/exporters/neuron/base.py
@@ -379,7 +379,7 @@ class NeuronDecoderConfig(NeuronConfig):
         be passed to export the model,
     - NEURONX_CLASS (`str`) -- the name of the transformers-neuronx class to instantiate for the model.
     It is a full class name defined relatively to the transformers-neuronx module, e.g. `gpt2.model.GPT2ForSampling`
-    - CONTINUOUS_BATCHING (`bool`) -- Whether the model supports continuous batching or not.
+    - CONTINUOUS_BATCHING (`bool`, , defaults to `False`) -- Whether the model supports continuous batching or not.
 
     The NEURONX_CLASS must always be defined in each model configuration.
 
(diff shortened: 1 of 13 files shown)


optimum/exporters/neuron/base.py
```diff
@@ -379,7 +379,7 @@ class NeuronDecoderConfig(NeuronConfig):
         be passed to export the model,
     - NEURONX_CLASS (`str`) -- the name of the transformers-neuronx class to instantiate for the model.
     It is a full class name defined relatively to the transformers-neuronx module, e.g. `gpt2.model.GPT2ForSampling`
-    [`~optimum.utils.DummyInputGenerator`] specifying how to create dummy inputs.
+    - CONTINUOUS_BATCHING (`bool`) -- Whether the model supports continuous batching or not.
```
michaelbenayoun, line 382:
```suggestion
    - CONTINUOUS_BATCHING (`bool`, defaults to `False`) -- Whether the model supports continuous batching or not.
```

optimum/neuron/modeling.py
```diff
@@ -688,32 +688,82 @@ def forward(
...
+        seq_ids: Optional[torch.Tensor] = None,
+    ):
+        # The start_ids parameter has different meanings:
+        # - for continuous (unpadded) batching it corresponds to the sequence id,
+        # - for static batching it corresponds to the start of the padded sequence.
+        if self.continuous_batching:
+            if seq_ids is None:
+                seq_ids = torch.arange(input_ids.shape[0])
+            else:
+                assert seq_ids.shape[0] == input_ids.shape[0]
```
michaelbenayoun, line 704: nit: maybe raise an explicit error with a message saying that the shapes should match in this case

optimum/neuron/modeling.py
```diff
@@ -688,32 +688,82 @@ def forward(
...
+            return cache_ids
+        # Static batching
+        return None if prefill else torch.tensor([cache_len - 1], dtype=torch.int32)
+
+    def prepare_inputs_for_prefill(
+        self,
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        seq_ids: Optional[List[int]] = None,
+        **kwargs,
```
michaelbenayoun, line 733: Is this necessary? I tend to avoid it whenever possible.

optimum/neuron/modeling_decoder.py
```diff
@@ -131,16 +132,26 @@ def __init__(
 
         exporter = get_exporter(config, task)
 
-        # transformers-neuronx uses f32/f16 instead of fp32/fp16
-        auto_cast_type = auto_cast_type.replace("p", "")
+        tnx_kwargs = {
+            "batch_size": batch_size,
+            "tp_degree": num_cores,
```
michaelbenayoun, line 137: The TP degree is always exactly the number of neuron cores used?

optimum/neuron/modeling.py
JingyaHuang, line 704: Can we raise with error information here?

text-generation-inference/Dockerfile
```diff
@@ -92,8 +92,8 @@ RUN wget -qO - https://apt.repos.neuron.amazonaws.com/GPG-PUB-KEY-AMAZON-AWS-NEU
 RUN apt-get update -y \
  && apt-get install -y --no-install-recommends \
     aws-neuronx-dkms=2.15.9.0 \
-    aws-neuronx-collectives=2.20.11.0 \
-    aws-neuronx-runtime-lib=2.20.11.0 \
+    aws-neuronx-collectives=2.20.11.0-c101c322e \
```
JingyaHuang, line 95: What are these numbers?

JingyaHuang (approved):
LGTM, thanks David!

optimum/neuron/modeling.py
dacorvo: I am using an assert because these methods are always called internally, so it is rather to catch internal programming errors.

optimum/neuron/modeling_decoder.py
dacorvo: Yes. I use the num_cores name in optimum-neuron because I use it for two things: to set the TP degree (here) and also to restrict the number of cores I reserve at initialization (otherwise the TNX runtime takes them all).

PR #494, opened 2024-02-21
fix(cache): omit irrelevant config parameters in lookup

dacorvo:
# What does this PR do?

It adds a white list of parameters to omit when comparing config.json in cache lookups.

The current white-list is:

- `_name_or_path`,
- `transformers_version`,
- `eos_token_id`,
- `bos_token_id`,
- `pad_token_id`.


philschmid (approved):
LGTM! Thanks, makes it easy to update too

PR #498, opened 2024-02-28
GQA optimization for TP

michaelbenayoun:
# What does this PR do?

- Adds support for `GQAQKVColumnParallelLinear`, which makes it possible to have `tp_size >>> num_key_value_heads`
- Adds tests for this use case
- Adds tests for checkpoint consolidation after distributed training
- ~~Initialize the parallel layers directly on the `xla` device to save host memory~~, **in another PR**.


dacorvo (commented):
Very hard to review in details as it is a very complex pull-request with many changes.
It looks good at first glance, but I have one question: you seem to sometimes use Tensor.copy_() (just creates a new tensor) and sometimes Tensor.clone() without detach() (data is still attached to old Tensor). Is it on on purpose ?

michaelbenayoun:
> Very hard to review in details as it is a very complex pull-request with many changes. It looks good at first glance, but I have one question: you seem to sometimes use Tensor.copy_() (just creates a new tensor) and sometimes Tensor.clone() without detach() (data is still attached to old Tensor). Is it on on purpose ?

We use `Tensor.copy_()` when loading weights to a model parameter. It will copy the tensor data into the parameter, and even move it to the proper device if needs be, so it is very convenient.
We use `Tensor.clone()` when we want to perform some task on a copy of the tensor, it should not affect anything here since we do that on `cpu` tensors during parallelization, or during checkpointing. I detach them now as your suggestion since it makes things safer!

optimum/neuron/accelerate/accelerator.py
```diff
@@ -422,7 +422,8 @@ def _prepare_model_for_mp(
         cpu_ids = {name: id(param) for name, param in model.named_parameters()}
         tied_parameters_dict = get_tied_parameters_dict(model)
         model_main_input_name = getattr(model, "main_input_name", None)
-        model = self.state.mp_plugin.parallelize_model(model, device=self.device)
+        # TODO: use self.device.
```
dacorvo, line 425: You changed that back also ... Is it because the tests were failing ?

michaelbenayoun: Yes, and since it is not related to the PR, I will work on that on another PR.

optimum/neuron/accelerate/accelerator.py
```diff
@@ -435,6 +436,11 @@ def _prepare_model_for_mp(
         else:
             model_to_cast = model
 
+        # Update CPU ids
+        original_parameter_names_to_gqa_qkv_names = model._gqa_qkv_metadata["original_names_to_gqa_qkv_names"]
+        for key in list(cpu_ids.keys()):
```
JingyaHuang, line 441:
```suggestion
        for key in cpu_ids.keys():
```

optimum/neuron/distributed/checkpointing.py
```diff
@@ -16,32 +16,58 @@
...
+def create_gqa_query_or_output_projection_weight_from_full_weight(
+    full_weight: torch.Tensor,
+    tp_size: int,
+    num_attention_heads: int,
+    num_key_value_heads: int,
+    kv_size_multiplier: int,
+    query_or_output: Union[Literal["query"], Literal["output"]],
+):
+    assert query_or_output in ["query", "output"]
+    assert full_weight.device == torch.device("cpu")
```
JingyaHuang, line 38: Maybe raise with information?

JingyaHuang (approved):
LGTM, thanks for adding the feature. I'm not familiar with this feature so not many meaningful advice for that...

optimum/neuron/accelerate/accelerator.py
michaelbenayoun: No, because in the for loop I update the keys so I want to work on a copy of the initial keys.

optimum/neuron/distributed/checkpointing.py
michaelbenayoun: I do not think it's needed as it's quite low level functions.

PR #501, opened 2024-02-29
Update Inference supported models list

JingyaHuang:
# What does this PR do?

as per title


JingyaHuang:
And perhaps we should place the list in a more visible place instead of under `exporter`, WDYT @philschmid ?

michaelbenayoun (approved):

PR #506, opened 2024-03-06
TGI: optimize continuous batching and improve export

dacorvo:
# What does this PR do?

1. This first modifies the TGI continuous batching implementation to take advantage of `transformers-neuronx` implementation.

Instead of dropping the KV cache when adding new requests and rebuilding it from cached texts, we simply omit the pending requests when calling model.forward, specifying only the indices of the new requests to prefill.

A llama TGI unit test is specifically added to verify the results are still correct after that change (for Llama and Mistral, `transformers-neuronx` continuous batching is always on).

2. For Sagemaker deployment, some disk usage logs are added when fetching/exporting a model.

3. During export, the model generation config is fetched to provide default values.

### DIFF
--- a/text-generation-inference/server/text_generation_server/model.py
+++ b/text-generation-inference/server/text_generation_server/model.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 import time
 from typing import Optional
 
@@ -50,6 +51,16 @@ def is_cached(model_id, neuron_config):
     return in_cache
 
 
+def log_cache_size():
+    path = HF_HUB_CACHE
+    if os.path.exists(path):
+        usage = shutil.disk_usage(path)
+        gb = 2**30
+        logger.info(f"Cache disk [{path}]: total = {usage.total/gb:.2f} G, free = {usage.free/gb:.2f} G")
+    else:
+        raise ValueError(f"The cache directory ({path}) does not exist.")
+
+
 def fetch_model(
     model_id: str,
     revision: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


text-generation-inference/server/text_generation_server/model.py
```diff
@@ -99,14 +111,22 @@ def fetch_model(
     logger.warning(f"{model_id} is not a neuron model: it will be exported using cached artifacts.")
     start = time.time()
     logger.info(f"Exporting model to neuron with config {neuron_config}.")
+    log_cache_size()
     start = time.time()
     model = NeuronModelForCausalLM.from_pretrained(model_id, export=True, **export_kwargs)
     end = time.time()
     logger.info(f"Model successfully exported in {end - start:.2f} s.")
     logger.info(f"Saving exported model to local storage under {export_path}.")
+    log_cache_size()
```
JingyaHuang, line 120: Will there be anything cached in `HF_HUB_CACHE`? I thought there would only be a cache in `neuron_cache_path`(`/var/tmp/neuron-compile-cache`) after a fresh export.

JingyaHuang (approved):
LGTM, just left a question.

text-generation-inference/server/text_generation_server/model.py
dacorvo: We need some space to expand the checkpoints and store the cached artifacts when we export a hub model.
Whenever we use the `hubclientlibrary` to fetch something (when calling from_pretrained), it ends up under `HF_HUB_CACHE`, so it is supposed to be writeable.
I therefore use the same destination when exporting.
When we deploy our own containers, we mount a writable `/data` volume.
With SageMaker, we only have a writable `/tmp` volume of 30G that can be expanded to 512G.
This is why I added those logs to identifiy issues when exporting the models.

PR #510, opened 2024-03-11
[Inference] Neuron cache for traced torchscript models (encoders, stable diffusion)

### LINKED ISSUE #284: Support for inference in Neuron cache
Since many customers will use Neuron for inference, in order to speed up deployment of the inference of common models it would help substantially if they could download the compiled artefacts from the Neuron cache for inference. Today it only supports training

JingyaHuang:
# What does this PR do?

Fixes #284 

- [x] Cache compiled artifacts (with preprocessor, config, etc.) during export.
- [x] Synchronize with hub remote cache.
- [x] Look up cached models through `optimum-cli`.
- [x] Check cache existence while `export=True` while using the modeling API. 

## Before submitting

-----

TEST

* Set remote repo

```bash
optimum-cli neuron cache set Jingya/optimum-neuronx-cache
```

* Synchronize

(enable setting custom cache repo for synchronize)

```bash
optimum-cli neuron cache synchronize (--repo_id Jingya/optimum-neuronx-cache)
```

* Lookup

```bash
 optimum-cli neuron cache lookup google-bert/bert-base-cased --mode inference
```

(eg. `google-bert/bert-base-uncased` can reuse the cached artifacts from `google-bert/bert-base-cased`)

----

**Supported**
* Encoders, stable diffusion

**Next steps**
* T5 (will come in the next PR)
* Doc for cache system
* Auto-fill cache tools


optimum/neuron/modeling_base.py
```diff
@@ -316,34 +278,86 @@ def _export(
...
+                compiler_version=NEURON_COMPILER_VERSION,
+                inline_weights_to_neff=inline_weights_to_neff,
+                optlevel=optlevel,
+                model_type=getattr(config, "model_type", None),
+                task=task,
+            )
+            cache_config = build_cache_config(compilation_config)
+            cache_entry = ModelCacheEntry(model_id=model_id, config=cache_config)
+            cache_repo_id = load_custom_cache_repo_name_from_hf_home()
+            compile_cache = _create_hub_compile_cache_proxy(cache_repo_id=cache_repo_id)
```
dacorvo, line 300: You should modify this internal helper to `create_hub_compile_cache_proxy`.
Also, you should not pass cache_repo_id, and let the helper get the default value for you.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -167,6 +212,10 @@ def upload_file(self, cache_path: str, src_path: str):
         # Only upload to the default cache: use synchronize to populate the Hub cache
         self.default_cache.upload_file(cache_path, src_path)
 
+    def upload_folder(self, cache_dir: str, src_dir: str):
+        # Upload folder the default cache: use synchronize to populate the Hub cache
```
dacorvo, line 216:
```suggestion
        # Upload folder to the default cache: use synchronize to populate the Hub cache
```

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,151 @@ def get_hub_cached_entries(
...
+            # seq-to-seq
+            clean_configs["model_type"] = next(iter(clean_configs.values()))["model_type"]
+
+        return clean_configs
+    else:
+        return next(iter(clean_configs.values()))
+
+
+def cache_aot_neuron_artifacts(neuron_dir: Path, cache_config_hash: str):
+    cache_repo_id = load_custom_cache_repo_name_from_hf_home()
```
dacorvo, line 562: 1. This is where you should call the context.
2. No need to reevaluate cache_repo_id, as you use the default value.
3. Inside the context, you should call directly `create_compile_cache` (which has been patched to _create...)

tests/cache/test_neuronx_cache.py
```diff
@@ -40,7 +53,12 @@ def cache_repos():
     if api.repo_exists(cache_repo_id):
         api.delete_repo(cache_repo_id)
     cache_repo_id = api.create_repo(cache_repo_id, private=True).repo_id
+    api.repo_info(cache_repo_id, repo_type="model")
```
dacorvo, line 56: Do you need this ?

tests/cache/test_neuronx_cache.py
```diff
@@ -40,7 +53,12 @@ def cache_repos():
...
     cache_dir = TemporaryDirectory()
+    set_custom_cache_repo_name_in_hf_home(
```
dacorvo, line 58: You don't need to do that because I already set the environment variable: please remove it.

tests/cache/test_neuronx_cache.py
```diff
@@ -84,18 +150,40 @@ def check_decoder_generation(model):
...
+    assert isinstance(image, PIL.Image.Image)
+
+
 def get_local_cached_files(cache_path, extension="*"):
     links = glob.glob(f"{cache_path}/**/*/*.{extension}", recursive=True)
     return [link for link in links if os.path.isfile(link)]
 
 
-def check_cache_entry(model, cache_path):
+def check_jit_cache_entry(model, cache_path):
```
dacorvo, line 171: The decoder cache is not really a jit cache.

tests/cache/test_neuronx_cache.py
```diff
@@ -152,6 +335,7 @@ def test_decoder_cache(cache_repos):
     ids=["invalid_repo", "invalid_endpoint", "invalid_token"],
 )
 def test_decoder_cache_unavailable(cache_repos, var, value, match):
+    unset_custom_cache_repo_name_in_hf_home()  # clean the repo set by cli since it's prioritized than env variable
```
dacorvo, line 338: Please remove this

tests/cache/test_neuronx_cache.py
```diff
@@ -182,3 +366,4 @@ def test_optimum_neuron_cli_cache_synchronize(cache_repos):
     stdout = stdout.decode("utf-8")
     assert p.returncode == 0
     assert f"1 entrie(s) found in cache for {model_id}" in stdout
+    unset_custom_cache_repo_name_in_hf_home()
```
dacorvo, line 369: And this

dacorvo (commented):
Looks good to me on principle, but I have suggested a few changes in the neuronx cache. Also, we need to sync on the env vs explicit call for cache repo id, as I am not aligned with your changes, that have unwanted side-effects.

tests/cache/test_neuronx_cache.py
JingyaHuang: Changed to `check_decoder_cache_entry`.

tests/cache/test_neuronx_cache.py
JingyaHuang: removed

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: Applied the first two suggestions, and I would keep with `create_hub_compile_cache_proxy()` instead of `hf_create_compile_cache` since it asks for the required positional argument: 'cache_url'.

optimum/neuron/modeling_base.py
JingyaHuang: Great catch, applied!

optimum/exporters/neuron/__main__.py
```diff
@@ -537,6 +548,8 @@ def decoder_export(
     output: Union[str, Path],
     **kwargs,
 ):
+    from ...neuron import NeuronModelForCausalLM
```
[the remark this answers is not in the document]
JingyaHuang: Crossed import as we discussed. Let's apply the workaround here and better tackle it in a new PR.

optimum/neuron/modeling_base.py
```diff
@@ -316,34 +277,85 @@ def _export(
...
+                    library_name=library_name,
+                    cache_dir=cache_dir,
+                    use_auth_token=use_auth_token,
+                    local_files_only=local_files_only,
+                    force_download=force_download,
+                    trust_remote_code=trust_remote_code,
+                )
+                # replace weights
+                neuron_model.replace_weights(weights=model)
+                return neuron_model
```
michaelbenayoun, line 323: nit: I would have one return in the end to make the function more readable.

optimum/neuron/modeling_diffusion.py
```diff
@@ -541,16 +577,15 @@ def _export(
         force_download: bool = True,
         cache_dir: Optional[str] = None,
         compiler_workdir: Optional[str] = None,
-        inline_weights_to_neff: bool = True,
+        disable_neuron_cache: Optional[bool] = False,
```
michaelbenayoun, line 580: Why `Optional`? It's described as `bool` only in the docstring below.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -151,6 +169,33 @@ def download_file(self, filename: str, dst_path: str):
...
+        if self.default_cache.exists(folder_path):
+            # cached locally
+            return True
+        else:
+            rel_folder_path = self._rel_path(folder_path)
+            try:
+                folder_info = list(self.api.list_repo_tree(self.repo_id, rel_folder_path))
+                folder_exists = len(folder_info) > 1
+            except Exception as e:
+                logger.info(f"{rel_folder_path} not found in {self.repo_id}: {e} \nThe model will be recompiled.")
```
michaelbenayoun, line 183: It should be log if `folder_exists` is `False`, not only when it fails?

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -167,6 +212,10 @@ def upload_file(self, cache_path: str, src_path: str):
         # Only upload to the default cache: use synchronize to populate the Hub cache
         self.default_cache.upload_file(cache_path, src_path)
 
+    def upload_folder(self, cache_dir: str, src_dir: str):
+        # Upload folder to the default cache: use synchronize to populate the Hub cache
+        shutil.copytree(src_dir, cache_dir, dirs_exist_ok=True)
```
michaelbenayoun, line 217: nit: maybe give the method another name if it does no upload anything? Or you are overriding an existing method?

dacorvo (approved):
Thank you so much for this very important feature. Nice and nicely done !

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -185,10 +234,14 @@ def download_file_to_string(self, filename: str, limit: int = None):
 
 def get_hub_cache():
     HUB_CACHE = "aws-neuron/optimum-neuron-cache"
-    return os.getenv("CUSTOM_CACHE_REPO", HUB_CACHE)
+    custom_hub_cache = load_custom_cache_repo_name_from_hf_home()
+    if custom_hub_cache is not None and len(custom_hub_cache) > 0:
+        return custom_hub_cache
+    else:
+        return os.getenv("CUSTOM_CACHE_REPO", HUB_CACHE)
```
michaelbenayoun, line 241: We already have a function that does all of this in `optimum/neuron/cache_utils.py` called `get_hf_hub_cache_repos`. I plan to "merge" the utils from this file with the `hub_neuronx_cache` utils that are in common, but just wanted to let you know.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -214,16 +267,16 @@ class ModelCacheEntry:
 
     """
 
-    def __init__(self, model_id: str, config: PretrainedConfig):
+    def __init__(self, model_id: str, config: Union[PretrainedConfig, Dict]):
```
michaelbenayoun, line 270:
```suggestion
    def __init__(self, model_id: str, config: Union[PretrainedConfig, Dict[str, Any]]):
```

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -214,16 +267,16 @@ class ModelCacheEntry:
...
         self.model_id = model_id
         # Remove keys set to default values
-        self.config = config.to_diff_dict()
+        self.config = config.to_diff_dict() if isinstance(config, PretrainedConfig) else config
```
michaelbenayoun, line 273: Since `self.config` is changed afterwards by popping some elements, I would recommend working on a copy of it.
```suggestion
        self.config = config.to_diff_dict() if isinstance(config, PretrainedConfig) else dict(config)
```

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+            if value != target_entry_config[name]:
+                is_matched = False
+    if is_matched:
+        neuron_config.pop("model_type", None)
+        model_entries.append(neuron_config)
+
+    return model_entries
+
+
+def _prepare_config_for_matching(entry_config, target_entry, model_type):
```
michaelbenayoun, line 468: nit: I would define this function above the other since it's being used by the other (for visibility)

michaelbenayoun, line 468: Can you add type annotations please? I am a bit lost on what is a config and what is a `ModelCacheEntry`.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+
+    return model_entries
+
+
+def lookup_matched_entries(entry_config, target_entry, white_list, model_entries, model_type):
+    is_matched = True
+    entry_config, target_entry_config, neuron_config = _prepare_config_for_matching(
+        entry_config, target_entry, model_type
+    )
+    for name, value in entry_config.items():
```
michaelbenayoun, line 450: `entry_config` can never be `None`?

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+        if isinstance(value, Dict):
```
michaelbenayoun, line 451:
```suggestion
        if isinstance(value, dict):
```

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+            for param in white_list:
+                value.pop(param, None)
+                target_entry_config[name].pop(param, None)
```
michaelbenayoun, line 454: Same as before since we are popping things, make sure that you prepare function returns a copy of the original configs.

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+            for term in set(entry_config[name]).intersection(set(target_entry_config[name])):
+                if entry_config[name][term] != target_entry_config[name][term]:
+                    is_matched = False
```
michaelbenayoun, line 457: Should we break here?

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+        else:
+            if value != target_entry_config[name]:
+                is_matched = False
```
michaelbenayoun, line 460: Same here for breaking?

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+    if neuron_white_list is None:
+        neuron_white_list = NEURON_CONFIG_WHITE_LIST
+
+    for param in white_list:
+        config.pop(param, None)
+
+    for param in neuron_white_list:
+        config["neuron"].pop(param, None)
+
+    return config
```
michaelbenayoun, line 524: Same as before: should we work on a copy instead?

optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -364,25 +417,153 @@ def get_hub_cached_entries(
...
+
+
+# Only applied on traced TorchScript models
```
michaelbenayoun, line 527: I would more add a docstring than a comment like that.

optimum/neuron/utils/misc.py
```diff
@@ -532,7 +540,7 @@ def download_checkpoints_in_cache(
```
michaelbenayoun, line 541: @requires_torch_neuronx

optimum/neuron/utils/misc.py
```diff
@@ -543,7 +551,11 @@ def replace_weights(
         weights = weights.state_dict()
```
michaelbenayoun, line 552: `from torch_neuronx.xla_impl.data_parallel import DataParallel`

optimum/neuron/utils/misc.py
```diff
@@ -623,3 +646,29 @@ def maybe_save_preprocessors(
         src_name_or_path, subfolder=src_subfolder, trust_remote_code=trust_remote_code
     ):
         preprocessor.save_pretrained(dest_dir)
+
+
+class DiffusersPretrainedConfig(PretrainedConfig):
+    # override to update `model_type`
```
michaelbenayoun, line 652: Add as docstring instead.

optimum/commands/export/neuron.py
```diff
@@ -79,7 +84,7 @@ def parse_args_neuron(parser: "ArgumentParser"):
         help="Path indicating the directory where to store intermediary files generated by Neuron compiler.",
     )
     optional_group.add_argument(
-        "--disable-weights-neff-inline",
+        "--inline-weights-neff",
```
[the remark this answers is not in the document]
JingyaHuang: inlined weights/neff won't be able to apply the cache so I would prefer the default to be able to leverage cache afterward.

optimum/neuron/modeling_base.py
```diff
@@ -229,7 +240,8 @@ def _export(
         force_download: bool = False,
         cache_dir: Optional[str] = None,
         compiler_workdir: Optional[Union[str, Path]] = None,
-        inline_weights_to_neff: bool = True,
+        disable_neuron_cache: Optional[bool] = False,
```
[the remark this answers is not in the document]
JingyaHuang: It can't be None, good catch!

optimum/neuron/modeling_diffusion.py
JingyaHuang: sort of typo, it can't be None.

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: It's the `upload_file` method for traced inference, so I tried to name them in a similar way, `upload_file` in `libneuronxla` also copy file from one path to another like `upload_folder` here.

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: That will be awesome!! Yeah I also used utils from cache_utils, will be nice if we can organize them better.

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: No, here it could be that the model has never been cached so unfound, I only want report when there is an issue of reading the repo (like access, wrong repo_id etc.) but not warn when it's just empty.

optimum/neuron/utils/misc.py
JingyaHuang: For `replace_weights`? actually it shall work for inf1 as well.

optimum/neuron/utils/misc.py
JingyaHuang: I tried, but the doc builder will complain if we import directly `torch_neuronx`...

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: ok! will change it to docstring.

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: Will do!

JingyaHuang: will do!

optimum/neuron/utils/misc.py
michaelbenayoun: Alright!

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: These configs are dynamically built just for comparing, they are destroyed right after the lookup and not used anywhere else.

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: yes!

PR #524, opened 2024-03-21
Skip weight load during parallel compile

michaelbenayoun:
# What does this PR do?

Since the real weights are not needed during precompilation, we can skip weight initialization / loading.


dacorvo (approved):
LGTM, thanks (just saw a typo in base.py line 677, but keep it for another pr).

PR #529, opened 2024-03-26
Set up tgi environment values with the ones used to build the model

oOraph:
Need this to workaround the model static params, for the docker entrypoint to adapt tgi environment accordingly to the specified model This will make usage of the image easier: default params (e.g not specifying anything) should be enough for most models


oOraph:
note: the associated generated image is available for testing here :)
[docker.io]/raphael31415/neuronx-tgi:0.0.21.dev0

text-generation-inference/tgi_env.py
```diff
@@ -0,0 +1,236 @@
...
+        neuron_config,
+    )
+
+    if revision:
+        checkpoint_revision = neuron_config["checkpoint_revision"]
+        if revision != checkpoint_revision:
+            logger.debug(
+                "Specified revision does not match with the model one (%s != %s)", revision, checkpoint_revision
+            )
+            return False
```
dacorvo, line 141: You can remove this check: the revision in the neuron config is the one from the original model, that will be used to fetch the weights. It is not the neuron model revision passed as parameter to the script.

text-generation-inference/tgi_env.py
```diff
@@ -0,0 +1,236 @@
...
+]
+
+# By the end of this script all env var should be specified properly
+env_vars = tgi_server_env_vars + tgi_router_env_vars
+
+available_cores = get_available_cores()
+neuronxcc_version = get_neuronxcc_version()
+
+
+def parse_cmdline(argv: List[str] = None) -> argparse.Namespace:
```
dacorvo, line 38: This function actually aligns the env variables also, so I think its name should reflect that too.

text-generation-inference/tgi_env.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # max_prefill_tokens = os.getenv("MAX_BATCH_PREFILL_TOKENS")
+        # if not max_prefill_tokens:
+        #     max_prefill_tokens = max_input_length
+        # f.write("export MAX_BATCH_PREFILL_TOKENS={}\n".format(max_prefill_tokens))
+        #
+        # max_batch_total_tokens = os.getenv("MAX_BATCH_TOTAL_TOKENS")
+        # if not max_batch_total_tokens:
+        #     max_batch_total_tokens = str(neuron_config["sequence_length"] * neuron_config["batch_size"])
+        # f.write("export MAX_BATCH_TOTAL_TOKENS={}\n".format(max_batch_total_tokens))
```
dacorvo, line 99: You should remove this

dacorvo (changes requested):
This looks good to me, but I am a bit worried some configurations might not work.
Could you add integration tests under https://github.com/huggingface/optimum-neuron/tree/main/text-generation-inference/integration-tests ?
I also need to add a github workflow to build the image and run the integration tests (make tgi_docker_test).

oOraph:
> This looks good to me, but I am a bit worried some configurations might not work. Could you add integration tests under https://github.com/huggingface/optimum-neuron/tree/main/text-generation-inference/integration-tests ? I also need to add a github workflow to build the image and run the integration tests (make tgi_docker_test).

done, both -> tgi_implicit_env.py and workflow

oOraph:
Actually I remove the workflow, the integration test test_gpt2.py cannot work for the local_neuron variant, reason:

Some directory is filled with data here:
https://github.com/huggingface/optimum-neuron/blob/6856557565c20c16311191409adf7968d41253ea/text-generation-inference/integration-tests/test_gpt2.py#L27

then this directory is expected to be shared with the docker container, here:
https://github.com/huggingface/optimum-neuron/blob/6856557565c20c16311191409adf7968d41253ea/text-generation-inference/integration-tests/conftest.py#L115

the problem is that this cannot work if tests are run within a container + docker dind environment as the volume filled in the first container won't be available on the host and thus won't end up on the second (hence tgi will launch with an empty dir)

-> so either we remove the local_neuron variant or we find a way to share the volume between the container running pytests and the one spawned by pytests

oOraph:
I had to deactivate/remove all tests related to aws-neuron/gpt2-neuronx-bs4-seqlen1024 because of the neuronx-cc upgrade v2.13.xxx

dacorvo (commented):
LGTM, thanks. Before merging, could you:
- bump dev version,
- use a single workflow for TGI,
- simplify a bit the implicit env test to just check you received a response to a single request.

.github/workflows/test_inf2_tgi.yml
```diff
@@ -50,3 +50,10 @@ jobs:
           sudo apt install gawk -y
           source aws_neuron_venv_pytorch/bin/activate
           HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} make tgi_test
+      - name: Run integration tests
+        shell: bash
+        run: |
+          # gawk is required when invoking the Makefile targets
+          sudo apt install gawk -y
```
dacorvo, line 57: It has already been installed in the previous step, but nevermind

text-generation-inference/Dockerfile
```diff
@@ -77,6 +80,9 @@ RUN VERBOSE=1 BUILDDIR=/pyserver/build PROTODIR=/pyserver/proto VERSION=${VERSIO
 # Neuron base image (used for deployment)
 FROM base AS neuron
 
+ARG VERSION
```
dacorvo, line 83: Do we really need to repeat VERSION multiple times ?

dacorvo (approved):
Thank you very much for the pull-request !

text-generation-inference/Dockerfile
oOraph: yep I wasn't even able to fully explain it: without it, on github ci the docker build process failed as you can see here:
https://github.com/huggingface/optimum-neuron/actions/runs/8552046964/job/23432354170
my guess is it's due to a buildkit version... not 100% sure though

.github/workflows/test_inf2_tgi.yml
oOraph: fixed

oOraph:
Side note: I bumped the version to 0.0.22.dev0 which will temporarily break tests (as there are no compatible cached models for the CI yet: gpt 2 compiled with neuronx-cc 2.13.66.0+6dfecc895 on 1 or 2 cores)

PR #531, opened 2024-03-27
Fix GQA permutation computation and sequential weight initialization / loading when doing TP

michaelbenayoun:
# What does this PR do?

- Fixes the way indices are compute for GQA permutation of the query and output projection, and add a test case to make sure everything works
- Add the possibility to specify the number of concurrent ranks that can initialize or load the model weights at the same time under TP. It can be useful to avoid going out-of-memory. 
- Fixes a typo `indicies -> indices`


examples/language-modeling/run_clm.py
michaelbenayoun, line 1: Will restore this change before merging.

optimum/neuron/distributed/base.py
```diff
@@ -698,14 +705,20 @@ def should_parallelize_layer_predicate_func(layer):
...
-                cls._maybe_load_weights_to_parallel_linears(model)
-
-            if skip_linear_weight_load or any(p.device == torch.device("meta") for p in model.parameters()):
-                # Initialize or load the weights for the parallelized model if it was lazily loaded.
-                cls._initialize_or_load_weights(model, names_of_the_parameters_to_consider, device=device)
-
+            local_rank = xm.get_local_ordinal()
+            if num_ranks_per_loading_step < 0:
+                num_ranks_per_loading_step = get_local_world_size()
+            for worker in range(math.ceil(get_local_world_size() / num_ranks_per_loading_step)):
```
JingyaHuang, line 711: So there will be two workers (0, 1) if `get_local_world_size() / num_ranks_per_loading_step` > 1 and < 2?

JingyaHuang (approved):
LGTM. Thanks for the fix!

optimum/neuron/distributed/base.py
michaelbenayoun: I can set `num_ranks_per_loading_step = min(num_ranks_per_loading_step, get_local_world_size())` as a safety measure.

michaelbenayoun:
Ran the distributed tests locally and they pass. It's just a flaky test that needs to be solved.

Issue #532, opened 2024-03-27
optimum-cli does not support exporting to Neuron by default in HF DL AMI v20240318


mlopezr:
### System Info

```shell
The Hugging Face Deep Learning AMI v20240318 ships with optimum-1.17.1 and optimum-neuron-0.0.20.dev0
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

1. Launch an inf2.24xlarge instance with Hugging Face Deep Learning AMI v20240318 in us-east-1 (AMI ID ami-0d608117d5f482a2f).
2. Run `optimum-cli export neuron --help`:
```
$ optimum-cli export neuron --help
usage: optimum-cli export [-h] {onnx,tflite} ...
optimum-cli export: error: invalid choice: 'neuron' (choose from 'onnx', 'tflite')
```
3. To solve it, run `pip uninstall optimum optimum-neuron` (which uninstalled optimum-1.17.1 and optimum-neuron-0.0.20.dev0) and then `pip install optimum[neuronx]` (which installs libneuronxla-0.5.669 optimum-1.18.0 optimum-neuron-0.0.18 torch-neuronx-1.13.1.1.13.0 transformers-neuronx-0.9.474)

### Expected behavior

I would expect `optimum-cli export neuron --help` to work out of the box in the latest HF DLAMI.

Issue #142, opened 2023-07-17
optimum-cli export does not take 'neuron' as a valid choice


adekunleoajayi:
I am following this [tutorial ](https://www.philschmid.de/optimize-deploy-bert-inf2) on how to deploy LLM to AWS inferentia2.

When I try to convert the model to AWS neuron using the optimum-cli 

`optimum-cli export neuron --model yiyanghkust/finbert-tone --sequence_length 128 --batch_size 1 tmp/`

I get the following error

```
usage: optimum-cli export [-h] {onnx,tflite} ...
optimum-cli export: error: invalid choice: 'neuron' (choose from 'onnx', 'tflite')
```

philschmid:
@adekunleoajayi can you please share your installed version and environment information?

evellasques:
Hi @adekunleoajayi, I was running into the same issue. I think this is something with the development version of Optimum Neuron. I tried using the HF AMI as described here https://www.philschmid.de/setup-aws-trainium and then the `optimum-cli` works as expected. Again, when I install the development version inside an instance created on that AMI `optimum-cli` simply stops working.

Here is a dump of pip freeze on that AMI (hope it helps):
```
absl-py==1.4.0
accelerate==0.20.3
aiohttp==3.8.4
aiosignal==1.3.1
anyio==3.7.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asttokens==2.2.1
async-timeout==4.0.2
attrs==21.2.0
Automat==20.2.0
aws-neuronx-runtime-discovery==2.9
Babel==2.8.0
backcall==0.2.0
bcrypt==3.2.0
beautifulsoup4==4.12.2
bleach==6.0.0
blinker==1.4
boto3==1.27.0
botocore==1.30.0
cachetools==5.3.1
certifi==2020.6.20
cffi==1.15.1
chardet==4.0.0
charset-normalizer==3.1.0
click==8.0.3
cloud-init==23.1.2
cloud-tpu-client==0.10
colorama==0.4.4
coloredlogs==15.0.1
comm==0.1.3
command-not-found==0.3
configobj==5.0.6
constantly==15.1.0
cryptography==3.4.8
datasets==2.13.0
dbus-python==1.2.18
debugpy==1.6.7
decorator==5.1.1
defusedxml==0.7.1
dill==0.3.6
distlib==0.3.4
distro==1.7.0
distro-info===1.1build1
docutils==0.20.1
[...]

PR #533, opened 2024-03-27
upgrade optimum and then install optimum-neuron

shub-kris:
# What does this PR do?
This PR fixes issue #532 and applies the fix suggested by @dacorvo here: https://github.com/huggingface/optimum-neuron/issues/142#issuecomment-2022448325

Fixes # (issue)

## Before submitting


infrastructure/ami/scripts/install-huggingface-libraries.sh
```diff
@@ -20,6 +20,8 @@ pip install --upgrade --no-cache-dir \
     "jinja2==3.1.2" \
     "attrs==23.1.0"
 
+# Temporary fix for the issue: https://github.com/huggingface/optimum-neuron/issues/142
+pip install -U optimum
```
dacorvo, line 24: Looking tat the logs, it installs 1.18, and then when installing optimum-neuron, it seems to use 1.17. I am completely lost with this installation process.
Anyway, it seems to work but we need to clean this up.

dacorvo (approved):
LGTM, but still a workaround.

dacorvo:
Let's improve tests to make sure it really works.

infrastructure/ami/scripts/install-huggingface-libraries.sh
JingyaHuang: I think something that we should do is that we always keep optimum-neuron setup compatible with the latest optimum, in this way we are more comfortable to be able to install optimum neuron via either `optimum[neuronx]` or `optimum[neuron]` extra...

shub-kris:
> Let's improve tests to make sure it really works.

It doesn't work, when I try 
image

JingyaHuang:
My recent commits shall solve the installation issue, could you review it? Tested with `ami-0b40b2acbce648147`, the cli works as expected and the issue reported #545 is also solved. @dacorvo @philschmid @shub-kris

infrastructure/ami/scripts/install-huggingface-libraries.sh
```diff
@@ -9,24 +9,22 @@ echo "TRANSFORMERS_VERSION: $TRANSFORMERS_VERSION"
 echo "OPTIMUM_VERSION: $OPTIMUM_VERSION"
 
 pip install --upgrade --no-cache-dir \
-    "transformers[sklearn,sentencepiece,vision]==$TRANSFORMERS_VERSION" \
-    "datasets==2.16.1" \
-    "accelerate==0.23.0" \
-    "diffusers==0.25.0" \
```
philschmid, line 15: are those all controlled by optimum?

infrastructure/ami/scripts/install-huggingface-libraries.sh
```diff
@@ -9,24 +9,22 @@ echo "TRANSFORMERS_VERSION: $TRANSFORMERS_VERSION"
...
     "evaluate==0.4.1" \
     "requests==2.31.0" \
     "notebook==7.0.6" \
     "markupsafe==2.1.1" \
     "jinja2==3.1.2" \
     "attrs==23.1.0"
 
+# Temporary fix for the issue: https://github.com/huggingface/optimum-neuron/issues/142
+pip install -U optimum
```
philschmid, line 20: lets pin the version?

philschmid (commented):
LGTM

infrastructure/ami/scripts/install-huggingface-libraries.sh
JingyaHuang: All those are now controlled by extras of optimum-neuron, hard-coded this way could lead to dependencies conflicts and it's a bit difficult to maintain.

philschmid: 👍🏻

infrastructure/ami/scripts/install-huggingface-libraries.sh
JingyaHuang: I think from now on `optimum-neuron` will be always compatible with the latest optimum since the installation of the latest optimum-neuron via optimum's neuronx extra won't work otherwise

PR #539, opened 2024-03-28
Disable logging during precompilation

michaelbenayoun:
# What does this PR do?

- Disables logging during precompilation. This is because `neuron_parallel_compile` reads the logs to extract the HLO graphs to compile. For some reason, strange characters are logged which makes `neuron_parallel_compile` fail when it tries to read the file, resulting in only a small portion of the total number of graphs being compiled.  This in turn results in compilation during actual training which makes things prohibitively long. To avoid all of this we simply do not log.
-  Sets the main process fo callbacks to be `dp_rank=tp_rank=0 and pp_rank = pp_size - 1` under PP setting so that Wandb and other callback can actually track the loss. While this is not ideal, and will be improved in the coming days, it works for now.

### DIFF
--- a/examples/language-modeling/run_clm.py
+++ b/examples/language-modeling/run_clm.py
@@ -466,9 +466,9 @@ def main():
 
     # We resize the embeddings only when necessary to avoid index errors. If you are creating a model from scratch
     # on a small vocab and want a smaller embedding size, remove this test.
-    embedding_size = model.get_input_embeddings().weight.shape[0]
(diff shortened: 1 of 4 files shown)


JingyaHuang (approved):

dacorvo (approved):
LGTM, thanks

dacorvo:
The inf2 builders have changed, and for the TGI workflow that uses the Makefile, it is missing the gawk package that was previously there.

PR #541, opened 2024-03-29
Add setup runtime step for K8S

glegendre01:
# What does this PR do?

After moving the Inferentia2 builder to Kubernetes, only the AWS Neuron dkms driver is installed on each builder.
This adds an installation step to add other tools.

Fixes # (issue)

## Before submitting

### DIFF
--- a/.github/workflows/test_inf2.yml
+++ b/.github/workflows/test_inf2.yml
@@ -31,10 +31,8 @@ jobs:
           EOF
           wget -qO - https://apt.repos.neuron.amazonaws.com/GPG-PUB-KEY-AMAZON-AWS-NEURON.PUB | sudo apt-key add -
           sudo apt-get update -y
-          sudo apt-get install aws-neuronx-tools aws-neuronx-oci-hook aws-neuronx-runtime-lib aws-neuronx-collectives -y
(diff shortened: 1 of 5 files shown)


.github/workflows/test_inf2.yml
```diff
@@ -23,6 +23,16 @@ jobs:
...
+        run: |
+          . /etc/os-release
+          sudo tee /etc/apt/sources.list.d/neuron.list > /dev/null <<EOF
+          deb https://apt.repos.neuron.amazonaws.com ${VERSION_CODENAME} main
+          EOF
+          wget -qO - https://apt.repos.neuron.amazonaws.com/GPG-PUB-KEY-AMAZON-AWS-NEURON.PUB | sudo apt-key add -
+          sudo apt-get update -y
+          sudo apt-get install aws-neuronx-tools aws-neuronx-oci-hook aws-neuronx-runtime-lib aws-neuronx-collectives -y
+          export PATH=/opt/aws/neuron/bin:$PATH
       - name: Check AMI
```
dacorvo, line 36: You can remove the step where I displayed the AMI package versions then.
Also, we will always get the latest version of the packages:
- on one hand, this is good because we won't have to update the builders when a new SDK is released,
- on another hand it might break if the python packages are not compatible.

I think it is worth the risk, but would like to head what @JingyaHuang @michaelbenayoun think.

glegendre01: We can install specific version if you prefer . as you want

dacorvo: After discussion, since the dkms comes preinstalled in the builder with a pinned version, then yes.

dacorvo (approved):

Issue #544, opened 2024-03-29
optimum.neuron.modeling_decoder.get_available_cores returns a wrong result


oOraph:
### System Info

```shell
optimum-neuron main
```

### Who can help?

I think I can help :), I'll provide a fix proposal asap

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

Reproduce:

Spawn ec2 instance inf2.24xlarge

Spawn a container and give it only access to parts of the neuron devices, for example 1/6  (the container should not have the NEURON_RT_NUM_CORES env variable specified otherwise this will just mask the issue)

```
from optimum.neuron.modeling_decoder import get_available_cores 
print(get_available_cores())
```

prints 12, not 2

reason, from the code and what I observe:
https://github.com/huggingface/optimum-neuron/blob/e83dc20025065148a75192d6b999d52d3c6f2c0e/optimum/neuron/modeling_decoder.py#L60

looking here is apparently not accurate better to look directly in /dev

```
root@test-raph:/# ls -alh /sys/class/neuron_device/
total 0
drwxr-xr-x  2 root root 0 Mar 29 17:34 .
drwxr-xr-x 33 root root 0 Mar 29 17:34 ..
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron0 -> ../../devices/virtual/neuron_device/neuron0
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron1 -> ../../devices/virtual/neuron_device/neuron1
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron2 -> ../../devices/virtual/neuron_device/neuron2
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron3 -> ../../devices/virtual/neuron_device/neuron3
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron4 -> ../../devices/virtual/neuron_device/neuron4
lrwxrwxrwx  1 root root 0 Mar 29 17:43 neuron5 -> ../../devices/virtual/neuron_device/neuron5
```

```
root@test-raph:/# ls /dev/
core  fd  full  mqueue  neuron0  null  ptmx  pts  random  shm  stderr  stdin  stdout  termination-log  tty  urandom  zero
```

[...]

PR #546, opened 2024-04-02
fix: bug in get_available_cores within container

oOraph:
fixes https://github.com/huggingface/optimum-neuron/issues/544


oOraph:
@dacorvo, @JingyaHuang some tests fail but it does not look like it's related to any change made in this pr: is looks like some missing token or auth issues (at least for Optimum Neuron - Common tests on Trainium and Optimum neuron / Test INF1 inference / Run INF1 tests)

can you confirm ?

dacorvo:
> @dacorvo, @JingyaHuang some tests fail but it does not look like it's related to any change made in this pr: it looks like some missing token or auth issue (at least for both tests "Optimum Neuron - Common tests on Trainium" and "Optimum neuron / Test INF1 inference / Run INF1")
> 
> can you confirm ?

Yes, builds from forks cannot use the CI token, leading to unauthorized access errors.

dacorvo (approved):
LGTM, thanks !

PR #556, opened 2024-04-08
Add step closure instead of regular log

michaelbenayoun:
# What does this PR do?

As per title.


michaelbenayoun:
Handled in #602

PR #561, opened 2024-04-10
Extend TGI integration tests

dacorvo:
# What does this PR do?

This extends the TGI integration tests by running all tests on not only llama but also gpt2 model configurations.


text-generation-inference/integration-tests/model_fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+        logger.error(f"Failed to export model: {e}")
+        return
+
+
+@pytest.fixture(scope="session", params=MODEL_CONFIGURATIONS.keys())
+def neuron_model_config(request):
+    """Expose a pre-trained neuron model
+
+    The fixture first makes sure the following model artifacts are present on the hub:
+    - exported neuron model under optimum/neuron-testing-<version>-<name>,
```
JingyaHuang, line 61: Why do you want to both cache it under `optimum/neuron-testing-cache` repo and store in a repo `optimum/neuron-testing-<version>-<name>` instead of just the cache repo?

text-generation-inference/integration-tests/model_fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+    - a configuration name,
+    - the original model id,
+    - the export parameters,
+    - the neuron model id,
+    - the neuron model local path.
+
+    For each exposed model, the local directory is maintained for the duration of the
+    test session and cleaned up afterwards.
+    The hub model artifacts are never cleaned up and persist accross sessions.
+    They must be cleaned up manually when the optimum-neuron version changes.
```
JingyaHuang, line 75: btw do we also need to clean the cache in the official neuron cache repo?

text-generation-inference/integration-tests/model_fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+
+    """
+    config_name = request.param
```
JingyaHuang, line 78: it's "model_name" / "model_arch"?

JingyaHuang (approved):
LGTM, just left some minor questions to better understand. Thanks for enhancing the tests!

text-generation-inference/integration-tests/model_fixtures.py
dacorvo: Because I want to test both fetching a neuron model from the hub and exporting it "on-the_fly" from the cached artifacts.

text-generation-inference/integration-tests/model_fixtures.py
dacorvo: The official cache is not affected by these files. It is however sometimes affected when running tests locally: the registry is updated for dev versions for instance.
I usually remove the registry files afterwards.
We could also remove older compiler trees.

text-generation-inference/integration-tests/model_fixtures.py
dacorvo: It is a free-form string used as the key in the model_configurations dictionary (here gpt2, llama).

PR #566, opened 2024-04-13
Allow download subfolder for caching models with subfolder

JingyaHuang:
# What does this PR do?

The stable diffusion cache actually doesn't work when they have subfolder structure. This PR fixes this.

## Before submitting

### DIFF
--- a/optimum/neuron/utils/hub_neuronx_cache.py
+++ b/optimum/neuron/utils/hub_neuronx_cache.py
@@ -25,6 +25,7 @@ from tempfile import TemporaryDirectory
 from typing import Any, Dict, List, Literal, Optional, Union
 
 from huggingface_hub import HfApi, get_token
+from huggingface_hub.hf_api import RepoFile, RepoFolder
 from transformers import AutoConfig, PretrainedConfig
 
 from ..version import __version__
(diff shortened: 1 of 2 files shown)


optimum/neuron/utils/hub_neuronx_cache.py
```diff
@@ -189,12 +190,26 @@ def download_folder(self, folder_path: str, dst_path: str):
                 try:
                     # cached remotely
                     for repo_content in folder_info:
-                        # TODO: this works for `RepoFile` but not `RepoFolder`
-                        local_path = self.api.hf_hub_download(self.repo_id, repo_content.path)
-                        filename = Path(local_path).name
-                        dst_path = Path(dst_path)
-                        dst_path.mkdir(parents=True, exist_ok=True)
-                        os.symlink(local_path, dst_path / filename)
+                        if isinstance(repo_content, RepoFile):
```
dacorvo, line 193: Wasn't it possible then to use snapshot_download instead ?

michaelbenayoun (commented):
Same question as @dacorvo

optimum/neuron/utils/hub_neuronx_cache.py
JingyaHuang: Is snapshot_download usable for a subfolder under a repo instead of the whole repo? I tried it by setting `allow_patterns` as the hash key but there was no file matched...

JingyaHuang:
Hey @dacorvo and @michaelbenayoun, thanks for the comments, I just adjust a bit the solution but not using `snapshot_download`, it will just list all repo files in the subfolder(`list_repo_tree(..., recursive=True)`) and download.

dacorvo (approved):

PR #572, opened 2024-04-19
Use AWS 2.18.0 AMI as base

dacorvo:
# What does this PR do?

Unfortunately, the 0.0.21 HFDLAMI was generated from an outdated AWS DLAMI, leading to system packages out-of-sync with python packages. This bumps the base AWS DLAMI version to address this and allow the regeneration of a working 0.0.21 HFDLAMI.

### DIFF
--- a/infrastructure/ami/hcl2-files/variables.pkr.hcl
+++ b/infrastructure/ami/hcl2-files/variables.pkr.hcl
@@ -10,7 +10,7 @@ variable "instance_type" {
 }
 
 variable "source_ami" {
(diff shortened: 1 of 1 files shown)


michaelbenayoun (approved):

JingyaHuang (approved):
So the [`20240409 DLAMI`](https://aws.amazon.com/marketplace/pp/prodview-gr3e6yiscria2) is using outdated Neuron SDK? Could you add the step to the notion guide as well?

Issue #571, opened 2024-04-18
Optimum neuron inference test results with NotImplementedError.


musunita:
### System Info

```shell
Hugging Face Neuron Deep Learning AMI (Ubuntu 22.04)[ami=ami-073e0687022c65b38 ]

Instance type: Inf2.48xlarge

pre-installed package:

aws_neuron_venv_pytorch) ubuntu@ip-172-31-9-88:~$ pip list | grep "neuron"
aws-neuronx-runtime-discovery 2.9
libneuronxla                  0.5.971
neuronx-cc                    2.13.66.0+6dfecc895
neuronx-distributed           0.7.0
neuronx-hwm                   2.12.0.0+422c9037c
optimum-neuron                0.0.21
tensorboard-plugin-neuronx    2.6.7.0
torch-neuronx                 1.13.1.1.14.0
torch-xla                     1.13.1+torchneurone
transformers-neuronx          0.10.0.21
```

Using 0.0.21 version of optimum-neuron
aws_neuron_venv_pytorch) ubuntu@ip-172-31-9-88:~/optimum-neuron$ git branch
* (HEAD detached at v0.0.21)
```


### Who can help?

@dacorvo, @JingyaHuang,
Inference test details is enclosed in tests in optimum-neuron.pdf.
[Tests in Optimum Neuron (3) (4).pdf](https://github.com/huggingface/optimum-neuron/files/15029385/Tests.in.Optimum.Neuron.3.4.pdf)

While running inference tests on INf2.48xlarge, encounter following errors.
```
(aws_neuron_venv_pytorch) ubuntu@ip-172-31-9-88:~/optimum-neuron$ pytest -m is_inferentia_test tests
=================================================================================== test session starts ===================================================================================
platform linux -- Python 3.8.10, pytest-8.0.0, pluggy-1.4.0
rootdir: /home/ubuntu/optimum-neuron
configfile: pyproject.toml
plugins: anyio-4.2.0
collected 1134 items / 1 error / 665 deselected / 469 selected                                                                                                                            

ERRORS ==========================================================================================
_________________________________________________________________________ ERROR collecting tests/test_examples.py _________________________________________________________________________
tests/test_examples.py:514: in <module>
    class CausalLMExampleTester(ExampleTesterBase, metaclass=ExampleTestMeta, example_name="run_clm"):
tests/test_examples.py:288: in __new__
[...]

PR #578, opened 2024-04-22
Fixes `test_examples.py` tests collection

michaelbenayoun:
# What does this PR do?

As per title.
Fixes: #571

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -282,6 +282,9 @@ class ExampleTestMeta(type):
 
             tensor_parallel_size = 2 if tp_support is not TPSupport.NONE else 1
 
+            if not ParallelizersManager.is_model_supported(model_type):
+                continue
+
             if not is_neuronx_distributed_available():
                 pp_support = False
             else:
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -264,6 +264,8 @@ class ExampleTestMeta(type):
 
     def __new__(cls, name, bases, attrs, example_name=None):
         models_to_test = []
+        if not is_neuronx_distributed_available():
+            return
         if example_name is not None:
             models_to_test = _SCRIPT_TO_MODEL_MAPPING.get(example_name)
             if models_to_test is None:


musunita:
Thank for the fix. I dont see above error anymore. But I encounter a new error now as below.
```
No input shapes provided, using default shapes, {'batch_size': 1, 'sequence_length': 128}
_______________________________________________ test_export_no_parameters[feature-extraction-hf-internal-testing/tiny-random-XLMModel] _______________________________________________

self = <optimum.neuron.utils.hub_neuronx_cache.CompileCacheHfProxy object at 0x7f0d8009f850>, repo_id = 'optimum-internal-testing/optimum-neuron-cache-for-testing'
default_cache = <libneuronxla.neuron_cc_cache.CompileCacheFs object at 0x7f0f393d7c70>, endpoint = None, token = None

    def __init__(
        self, repo_id: str, default_cache: CompileCache, endpoint: Optional[str] = None, token: Optional[str] = None
    ):
        # Initialize the proxy cache as expected by the parent class
        super().__init__(default_cache.cache_url)
        self.cache_path = default_cache.cache_path
        # Initialize specific members
        self.default_cache = default_cache
        self.api = HfApi(endpoint=endpoint, token=token, library_name="optimum-neuron", library_version=__version__)
        # Check if the HF cache id is valid
        try:
            if not self.api.repo_exists(repo_id):
>               raise ValueError(f"The {repo_id} repository does not exist or you don't have access to it.")
E               ValueError: The optimum-internal-testing/optimum-neuron-cache-for-testing repository does not exist or you don't have access to it.

optimum/neuron/utils/hub_neuronx_cache.py:116: ValueError
```

michaelbenayoun:
That is because you do not have access to this repo, and this is the one used by default for the test. 
To be able to run the test on your side:

- Create a model repo on the Hub
- Set the environment variable `CUSTOM_CACHE_REPO=the-name-of-your-model-repo`
- Set the environment variable `HF_TOKEN=your-hf-token-that-provides-access-to-the-repo`

Issue #371, opened 2023-12-08
SDXL: `neuron_config` of `text_encoder_2` is `CLIPTextNeuronConfig` instead of `CLIPTextWithProjectionNeuronConfig`


neo:
I tried to search in the codebase and am not sure how they got chosen, but it appears that `pipeline.text_encoder_2.neuron_config` is `optimum.exporters.neuron.model_configs.CLIPTextNeuronConfig` while I suspect it should be `CLIPTextWithProjectionNeuronConfig`

neo:
wondering is this at least verified that it exists?

neo:
not stale? @JingyaHuang

JingyaHuang:
Will check! Sorry for the late reply.

JingyaHuang:
I kinda got where this question comes from, there is a bug on sdxl when we reconstruct neuron config of text encoder 2 from a json file 😅. The test I did above was during the export, but not the neuron config we reconstructed during the inference:

https://github.com/huggingface/optimum-neuron/blob/3359516a49e2c0c68e3db28d35ece53f38193439/optimum/neuron/modeling_diffusion.py#L524

neo:
hi! yes!

I was going to reply similar _just now_ 😂 I was looking at during loading/inference, and here's what I got:

![image](https://github.com/huggingface/optimum-neuron/assets/9028430/74a09157-56fc-4ab9-b111-777d1dd390d6)

neo:
I also have a feeling that the json file of text encoder 2 might be mistaken too:

https://huggingface.co/aws-neuron/stable-diffusion-xl-base-1-0-1024x1024/blob/main/text_encoder_2/config.json#L30-L34

the text encoder 2 output should include `pooler_output` based on my shallow understanding?

JingyaHuang:
From transformers' clip modeling code, text encoder (`CLIPTextModel`) shall be the one which [has `pooler_output` as output](https://github.com/huggingface/transformers/blob/37fa1f654f17b68bbe30440c64e611f1a4d55bc7/src/transformers/models/clip/modeling_clip.py#L753) while text encoder 2 (`CLIPTextModelWithProjection`) [outputs `text_embeds`](https://github.com/huggingface/transformers/blob/37fa1f654f17b68bbe30440c64e611f1a4d55bc7/src/transformers/models/clip/modeling_clip.py#L1238) which is the text emb obtained by applying the projection layer to the pooled output.

neo:
[...]

PR #581, opened 2024-04-25
Make stable diffusion pipelines compatible with compel

### LINKED ISSUE #357: SDXL with Compel: unexpected keyword argument `output_hidden_states`
From looking at the code, I could tell that `output_hidden_states` is always true for SDXL's text encoders; however in compel, it's always explicitly passed (even though it's also True for SDXL): https://github.com/damian0815/compel/blob/v2.0.2/src/compel/embeddings_provider.py#L392. And it's causing the `NeuronModelTextEncoder` to error out `unexpected keyword argument 'output_hidden_states'`.

So I'm wondering, what's the best way to approach this as well as possibly a similar issue with `return_dict`? Thanks!

JingyaHuang:
# What does this PR do?

- [x] Make `NeuronStableDiffusionXLPipeline` compel compatible
- [x] Make `NeuronStableDiffusionPipeline` compel compatible
- [x] Tests
- [x] Fix the neuron config wrongly reconstructed for sdxl's text encoder 2

Fixes #357 

@neo @Suprhimp
## Before submitting

### DIFF
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -867,7 +867,7 @@ class NeuronModelTextEncoder(_NeuronDiffusionModelPart):
         inputs = (input_ids,)
         outputs = self.model(*inputs)
 
-        if kwargs.pop("return_dict", False):
+        if return_dict:
             outputs = ModelOutput(dict(zip(self.neuron_config.outputs, outputs)))
 
         return outputs
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -27,6 +27,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 import torch
 from huggingface_hub import snapshot_download
 from transformers import CLIPFeatureExtractor, CLIPTokenizer, PretrainedConfig
+from transformers.modeling_outputs import ModelOutput
 
 from ..exporters.neuron import (
     load_models_and_neuron_configs,
@@ -521,7 +522,8 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
             if file_paths[1].is_file():
                 model_config = DiffusersPretrainedConfig.from_json_file(file_paths[1])
                 configs[name] = model_config
-                neuron_configs[name] = cls._neuron_config_init(model_config)
+                model_type = "clip-text-with-projection" if name == "text_encoder_2" else None
+                neuron_configs[name] = cls._neuron_config_init(model_config, model_type=model_type)
 
         if data_parallel_mode is None:
             data_parallel_mode = cls.set_default_dp_mode(configs["unet"])
@@ -585,6 +587,7 @@ class NeuronStableDiffusionPipelineBase(NeuronBaseModel):
         auto_cast: Optional[str] = "matmul",
         auto_cast_type: Optional[str] = "bf16",
         dynamic_batch_size: bool = False,
+        output_hidden_states: bool = False,
         data_parallel_mode: Optional[str] = None,
         lora_model_ids: Optional[Union[str, List[str]]] = None,
         lora_weight_names: Optional[Union[str, List[str]]] = None,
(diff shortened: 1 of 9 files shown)


dacorvo (approved):
LGTM, thanks ! And my deepest apologies for the wrongly formatted TGI test from my last pull-request you had to fix (looks like style-check is not triggered on TGI changes).

optimum/neuron/modeling_diffusion.py
```diff
@@ -842,9 +850,20 @@ def __init__(
     ):
         super().__init__(model, parent_model, config, neuron_config, DIFFUSION_MODEL_TEXT_ENCODER_NAME)
 
-    def forward(self, input_ids: torch.Tensor):
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        *args,
+        **kwargs,
+    ):
```
neo, line 858: I personally wouldn't just open up the door for any kwargs because there are actually other additional kwargs that are not being handled, and also I think we should assert the new args to make sure they are consistent with how it was compiled during tracing.
```suggestion
        output_attentions: Optional[bool] = None,
        output_hidden_states: Optional[bool] = None,
        return_dict: Optional[bool] = None,
    ):
        if attention_mask is not None:
            assert torch.equal(torch.ones_like(attention_mask), attention_mask), "attention_mask is expected to be only all ones"
        if output_hidden_states is not None:
            assert bool(self.text_encoder.config.output_hidden_states) == bool(output_hidden_states), "output_hidden_states is expected to be consistent with how it was compiled"

```

and this is like how I was going to PR it

optimum/neuron/modeling_diffusion.py
```diff
@@ -842,9 +850,20 @@ def __init__(
...
+        input_ids = input_ids.to(torch.long)  # dummy generator uses long int for tracing
+
         inputs = (input_ids,)
         outputs = self.model(*inputs)
+
+        if kwargs.pop("return_dict", False):
```
neo, line 864: supplementing the suggestion above so that you can directly apply and commit
```suggestion
        if return_dict:
```

neo (commented):
Thank you for inviting me to review! I left some minor comments, and I also tested it and it seemed like this would fix #371 as well 😆

Suprhimp:
I tested with my own model even with lora, and it works pretty well.

I compiled with `
output_hidden_states
` params `true`

And I tested with this example code 😃 

```

from optimum.neuron import NeuronStableDiffusionXLPipeline
from compel import Compel, ReturnedEmbeddingsType

pipe = NeuronStableDiffusionXLPipeline.from_pretrained("sdxl_turbo_neuron/", data_parallel_mode="all")
prompt = "Self-portrait oil painting++++++++++, a beautiful cyborg with golden hair, 8k"
negative_prompt = "worst quality"

compel = Compel(
        tokenizer=[pipe.tokenizer, pipe.tokenizer_2],
        text_encoder=[pipe.text_encoder, pipe.text_encoder_2],
        returned_embeddings_type=ReturnedEmbeddingsType.PENULTIMATE_HIDDEN_STATES_NON_NORMALIZED,
        requires_pooled=[False, True],
    )
prompt_embeds, pooled = compel(prompt)
neg_prompt_embeds, neg_pooled = compel(negative_prompt)

images = pipe(
        prompt_embeds=prompt_embeds,
        negative_prompt_embeds=neg_prompt_embeds,
        pooled_prompt_embeds=pooled,
        negative_pooled_prompt_embeds=neg_pooled, 
        guidance_scale=0.0, num_inference_steps=1).images

```

Thanks for the update!

optimum/neuron/modeling_diffusion.py
```diff
@@ -521,7 +522,8 @@ def _from_pretrained(
             if file_paths[1].is_file():
                 model_config = DiffusersPretrainedConfig.from_json_file(file_paths[1])
                 configs[name] = model_config
-                neuron_configs[name] = cls._neuron_config_init(model_config)
+                model_type = "clip-text-with-projection" if name == "text_encoder_2" else None
+                neuron_configs[name] = cls._neuron_config_init(model_config, model_type=model_type)
```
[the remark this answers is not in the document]
JingyaHuang: Thanks @neo, good catch. I actually put the wrong model type for text encoder 2 when I added the sdxl support, and then just patch the model type when reconstructing the neuron config. The right way to do it is just fix the model type! 

Thxxx that's super super helpful as suggestion! 💯

optimum/neuron/modeling_diffusion.py
neo: sorry for the wrong arg name 😅

optimum/neuron/modeling_diffusion.py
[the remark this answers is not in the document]
neo: glad it works this way 😆 I didn't really know if it was also coupled/connected with other things

because I saw [here](https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0/blob/main/text_encoder_2/config.json#L16) has `clip_text_model` and wasn't sure if it's the same thing

neo (approved):
glad you like my suggestions! 😆 

and thank you for making this happen ❤️

JingyaHuang:
Thanks @neo and @Suprhimp for the help, the PR is merged. I will include it in our next release!

PR #583, opened 2024-04-30
[Inference] Add `SentenceTransformers` support to `pipeline` for `feature-extration`

philschmid:
# What does this PR do?

This PR adds a new, slightly modified `FeatureExtractionPipeline` from Transformers that allows us to use it with `sentence-transformers` models. When using the `pipeline` object from `optimum,,` the library checks if the requested model for `feature-extraction` is a `sentence-transformers` model and if so, it would return the `sentence_embeddings` instead of the first hidden state. 

Thats is done by adding a new `is_sentence_transformer_model` that checks if the requested model is a `transformers` or `sentence-transformers` model. If it is a `sentence-transformers` model, it uses `NeuronModelForSentenceTransformers` and the `FeatureExtractionPipeline` returns `_model_outputs.sentence_embedding[0]` instead of `model_outputs[0]`

Example:
```python
from optimum.neuron import pipeline

input_shapes = {"batch_size": 1, "sequence_length": 64} 
p = pipeline("feature-extraction","sentence-transformers/all-MiniLM-L6-v2",export=True, input_shapes=input_shapes)
> # Using Sentence Transformers compatible Feature extraction pipeline

p("test")
> [0.06765521317720413,
> 0.06349243223667145,
> 0.04871273413300514,
> 0.0793028473854065,
```

Validated with `torch.allclose`

Implications: 
* `sentence-transformers` models will now always return the `sentence_embeddings` when initialized with the `FeatureExtractionPipeline` pipeline. 

Alternatives options: 
* Instead of modifying the `feature-extraction` pipeline, we could also introduce a new task `sentence-embeddings` to optimum, but that might hinder more general adoption since it is unique to optimum-neuron.


philschmid:
@tomaarsen can you also do a review?

optimum/neuron/pipelines/transformers/sentence_transformers.py
```diff
@@ -0,0 +1,99 @@
...
+config_name = "config_sentence_transformers.json"
+
+
+def is_sentence_transformer_model(model: str, token: str = None, cache_folder: str = None, revision: str = None):
+    """Checks if the model is a sentence transformer model based on provided model id"""
+    if is_sentence_transformers_available():
+        from sentence_transformers.util import load_file_path
+
+        try:
+            load_file_path(
```
JingyaHuang, line 17: This method shall also be applied to other non sentence transformers models isn't it? I think if we keep with the task name "feature-extraction" then we shall let the users tell or infer ourselves the library name infer_library_from_model which class to use with `TasksManager.infer_library_from_model`: https://github.com/huggingface/optimum/blob/e3fd2776a318a3a7b9d33315cc42c04c181f6d2f/optimum/exporters/tasks.py#L1690

dacorvo (commented):
Several pipeline tests are failing. It is unclear to me if they should be triggered or not.

optimum/neuron/pipelines/transformers/sentence_transformers.py
philschmid: Thank you for sharing the TaskManager. I wasn't aware of that. I ll update the method

optimum/neuron/pipelines/transformers/sentence_transformers.py
```diff
@@ -0,0 +1,90 @@
...
+
+        preprocess_params = tokenize_kwargs
+
+        postprocess_params = {}
+        if return_tensors is not None:
+            postprocess_params["return_tensors"] = return_tensors
+
+        return preprocess_params, {}, postprocess_params
+
+    def preprocess(self, inputs, **tokenize_kwargs) -> Dict[str, GenericTensor]:
```
tomaarsen, line 59: My primary concern at this time is that the Sentence Transformer tokenizer uses [this `max_seq_length`](https://huggingface.co/sentence-transformers/all-mpnet-base-v2/blob/main/sentence_bert_config.json#L2) as the "correct" maximum length as opposed to the value defined in the `tokenizer_config.json`. 

Here, we are relying on the tokenizer defined in the `Pipeline`, which won't use the max_seq_length. As a result, I think this ST pipeline component will perform differently (worse, to be precise) for longer input texts. A solution is to use `model_inputs = self.model.tokenize(inputs)` instead. 
Do note that the ST tokenize method does not allow for extra tokenize kwargs such as truncation, return_tensors, or padding. These are unfortunately hardcoded at the moment.

tomaarsen: Yet another solution is to rely exclusively on `self.model.encode(...)` in `def _forward`, but I recognize that this might clash with some requirements of the `Pipeline`.

philschmid: @tomaarsen for inferentia models need to be traced to a sequence length before running inference since we have static shapes. You always need to specify a `sequence_length` and `batch_size` before you can compile a model which is then used. 
This abstracted away by the user in the NeuronModelForSentenceTransformers class.

philschmid: There is no "sentence-transformers" used at all at the end. since the model is traced and it happens with transformers. We should be good here.

dacorvo (approved):
Disclaimer: I don't know sentence transformers, so this review is on the general outlook of the code and test, which looks good to me. @JingyaHuang and @tomaarsen reviews will be more relevant.

optimum/neuron/pipelines/transformers/sentence_transformers.py
philschmid: See here: https://github.com/huggingface/optimum-neuron/blob/18460aa22407586e82b345745a9fc55d9f4828b3/optimum/neuron/modeling.py#L201

tomaarsen: Ahh, I see! Thanks for the heads up. I figured it was more like Intel Gaudi, which does just work with regular Sentence Transformers (as long as the padding is "max_length" to also get static shapes & the device is "hpu"). 

Then my concern still stands: I think the [`max_seq_length`](https://huggingface.co/sentence-transformers/all-mpnet-base-v2/blob/main/sentence_bert_config.json#L2) might not be taken into account correctly.

optimum/neuron/pipelines/transformers/sentence_transformers.py
```diff
@@ -0,0 +1,90 @@
...
+
+from transformers.pipelines.base import GenericTensor, Pipeline
+
+from optimum.utils import is_sentence_transformers_available
+
+
+def is_sentence_transformer_model(model: str, token: str = None, revision: str = None):
+    """Checks if the model is a sentence transformer model based on provided model id"""
+    if is_sentence_transformers_available():
+        from optimum.exporters.tasks import TasksManager
```
JingyaHuang, line 11: We could put the import at the top of the file.

JingyaHuang (commented):
Thanks @philschmid, left some small nots.

optimum/neuron/pipelines/transformers/sentence_transformers.py
JingyaHuang: fyi, `max_seq_length` is not taken into account at all by the export of Neuron model. Shall we prevent users from setting static seq len higher than this value?

tomaarsen: For Sentence Transformer models, `max_seq_length` should have priority. In 99% of cases, this is stored in the `sentence_bert_config.json` file. You might indeed want to store it in a `config.json` when exporting, or override `model_max_length` in `tokenizer_config.json` as that should work in a more "expected" fashion.

optimum/neuron/pipelines/transformers/sentence_transformers.py
philschmid: done

JingyaHuang (approved):
LGTM, thanks @philschmid for adding it!

And thanks @tomaarsen for raising the concern of `max_seq_length`, it's on the backlog, will improve it in a coming PR.

Issue #576, opened 2024-04-22
Poor performance to generate images with NeuronStableDiffusionPipeline 


yahavb:
### System Info

```shell
I used https://huggingface.co/docs/optimum-neuron/tutorials/stable_diffusion to build and deploy inference endpoint. I used the optimum version https://github.com/yahavb/edge_diffusion_on_eks/blob/master/app/run-sd2.py against https://github.com/yahavb/edge_diffusion_on_eks/blob/master/app/run.py.

Inference of a single image with num_inference_steps=1 produced 824.9ms with NeuronStableDiffusionPipeline and 198.5ms for StableDiffusionPipeline.
```

### Who can help?

@JingyaHuang

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

https://github.com/aws-samples/edge_diffusion_on_eks

### Expected behavior

Comparable performance of NeuronStableDiffusionPipeline and StableDiffusionPipeline

JingyaHuang:
Hi @yahavb, thanks for opening the issue. Let me check if I can reproduce it.

JingyaHuang:
Hey @yahavb, have you tried setting `inline_weights_to_neff=True`? It's an arg that I recently set to `False` by default (since we would like to leverage it for caching), and according to my experiment, it seems to slow down quite heavily the inference...

yahavb:
setting `inline_weights_to_neff` to `True` improved the latency performance. thanks!

```
from optimum.neuron import NeuronStableDiffusionPipeline

compiler_args = {"auto_cast": "matmul", "auto_cast_type": "bf16","inline_weights_to_neff": "True"}
input_shapes = {"batch_size": batch_size, "height": height, "width": width}
stable_diffusion = NeuronStableDiffusionPipeline.from_pretrained(model_id, export=True, **compiler_args, **input_shapes)
```

PR #584, opened 2024-04-30
[Inference] Fix inference latency issue when weights/neff are separated

JingyaHuang:
# What does this PR do?

As reported in #576, the inference latency is heavily impacted when the weights and neff are not inlined. This is because the weights are not automatically loaded to neuron devices, and if we don't do that we suffer from huge host-device communication overhead.

This PR is supposed to patch this.

Caveat: Current data parallel API doesn't consider the case when weights and neff are not inlined. Here we use the class `WeightSeparatedDataParallel` as a temporary workaround. This will be included in Neuron SDK 2.20, and by then time this class will be removed from Optimum Neuron. And the current non-inlined models still show 1.5X latency compared to inlined models according to several small quick experiments. 

## Before submitting


optimum/neuron/modeling_base.py
```diff
@@ -100,25 +102,32 @@ def __init__(
...
+    ) -> torch.jit._script.ScriptModule:
         """
         Loads a TorchScript module compiled by neuron(x)-cc compiler. It will be first loaded onto CPU and then moved to
         one or multiple [NeuronCore](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/arch/neuron-hardware/neuroncores-arch.html).
 
         Args:
             path (`Union[str, Path]`):
                 Path of the compiled model.
+            to_neuron (`bool`, defaults to `False`):
+                Whether to move manually the traced model to NeuronCore. It's only needed when `inline_weights_to_neff=False`, otherwise it is loaded automatically to a Neuron device.
```
michaelbenayoun, line 121: You forgot `device_id`.

JingyaHuang: Added!

optimum/exporters/neuron/convert.py
```diff
@@ -31,7 +31,7 @@
     store_compilation_config,
 )
 from ...neuron.utils.cache_utils import get_model_name_or_path
-from ...neuron.utils.hub_neuronx_cache import (
+from ...neuron.utils.hub_cache_utils import (
```
dacorvo, line 34: So you did not like the name, or there is another reason ?

optimum/neuron/modeling_decoder.py
```diff
@@ -492,3 +492,10 @@ def push_to_hub(
             revision=revision,
             ignore_patterns=ignore_patterns,
         )
+
+    @staticmethod
```
dacorvo, line 496: We need to discuss this, as it is a bit hackish.

optimum/exporters/neuron/convert.py
JingyaHuang: With the lazy loading, I got an import error since the context manager is named the same as the script:
https://github.com/huggingface/optimum-neuron/blob/7439a2d32808ce19ddece2d9a34d047af46dd1b1/optimum/neuron/utils/hub_neuronx_cache.py#L307

optimum/neuron/modeling_decoder.py
JingyaHuang: Sure, it comes from this line in transformers: https://github.com/huggingface/transformers/blob/1518508467d96b3866fc4ebcb7a5b3a2e0df2aa4/src/transformers/pipelines/base.py#L883, I did not see an easy workaround other than that, do not want to override the pipeline in transformers neither...

dacorvo: I think that's because we don't handle the device correctly

dacorvo: Actually we handle it correctly, but no matter what, the pipeline code will call this.
Two options:
- bump transformers to 4.41.0,
- at least check that the device is identical to the hard-coded device we use ("cpu"), and raise an error otherwise.

dacorvo: This is fixed in transformers as a side-effect of this pull-request:
https://github.com/huggingface/transformers/pull/30534
The issue actually comes from `optimum` that reimplements the `PretrainedModel` class without providing all its methods (including `to()`).

dacorvo: I would suggest the following refactoring to address that issue and potential future issues with missing methods.:
https://github.com/huggingface/optimum-neuron/pull/608
This introduces a common `NeuronModel` class from which all neuron models should inherit.

dacorvo (approved):
LGTM, thanks !

PR #587, opened 2024-05-03
Improve llama models performance

dacorvo:
# What does this PR do?

This modifies the default Neuron configuration when exporting Llama models for inference, setting the attention layout to "BSH" instead of "HSB".

This configuration has almost no impact on the token generation time (a.k.a `decode`), and significantly reduces the context encoding time (a.k.a `prefill`).

In the process, the TGI router version is bumped to 2.0.2.


docs/source/_toctree.yml
```diff
@@ -50,6 +50,8 @@
       title: Llama2 13b on AWS Inferentia2
     - local: benchmarks/inferentia-mistral-v2
       title: Mistral v0.2 7b on AWS Inferentia2
+    - local: benchmarks/inferentia-llama3-8b
+      title: Llama3 8b on AWS Inferentia2
```
michaelbenayoun, line 54: nit: to stick with the same naming everywhere.
```suggestion
      title: Llama-3 8B on AWS Inferentia2
```

michaelbenayoun (approved):
LGTM
That looks awesome @dacorvo !

PR #589, opened 2024-05-03
Improve tgi env wrapper for neuron

oOraph:
when several compatible entries are found, take the one that uses the most cores and then has the biggest batch size


dacorvo (commented):
@philschmid did a similar pull-request on moon-landing for the snippets.
His code is much simpler, can you check it out ?
https://github.com/huggingface/moon-landing-aws-integrations/pull/10

dacorvo (approved):
LGTM, thanks !

Issue #593, opened 2024-05-10
caching diffusion models does not work


oOraph:
### System Info

```
optimum neuron 0.21.0
```

```
ii  aws-neuronx-collectives           2.20.22.0-c101c322e                     amd64        neuron_ccom built using CMake
ii  aws-neuronx-dkms                  2.16.7.0                                amd64        aws-neuronx driver in DKMS format.
ii  aws-neuronx-runtime-lib           2.20.22.0-1b3ca6425                     amd64        neuron_runtime built using CMake
ii  aws-neuronx-tools                 2.17.1.0                                amd64        Neuron profile and debug tools
```

### Who can help?

@JingyaHuang @oOraph 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

run the minimal script twice:
```
from optimum.neuron import NeuronStableDiffusionPipeline
model_id = "runwayml/stable-diffusion-v1-5"
compiler_args = {"auto_cast": "matmul", "auto_cast_type": "bf16"}
input_shapes = {"batch_size": 1, "height": 1024, "width": 1024}
stable_diffusion = NeuronStableDiffusionPipeline.from_pretrained(model_id, export=True, **compiler_args, **input_shapes)
```

The model get recompiled and exported twice

Logging details:

Lookup:
```
neuronxcc-2.13.66.0+6dfecc895/MODULE_500c5bd266330f2a15fa
```
Export:
```
neuronxcc-2.13.66.0+6dfecc895/MODULE_323b66a01f7f04552356
```

Troubleshoot:

there is a mismatch between the hash checked when looking for cached exports and the hash computed after building an export. Consequently, the model is never found in cache

the problem arise because there is a diff between cache configs here (lookup):

https://github.com/huggingface/optimum-neuron/blob/7e21931b6d74d21997ee5f5ea8742ab7da977e29/optimum/neuron/modeling_diffusion.py#L752

and here (export):

https://github.com/huggingface/optimum-neuron/blob/7e21931b6d74d21997ee5f5ea8742ab7da977e29/optimum/exporters/neuron/convert.py#L410

The diff is really minimal:

```
85c85
          "static_height" : 128,
---
          "static_height" : 16,
88c88
          "static_width" : 128
---
          "static_width" : 16
```

in unet config

[...]

PR #594, opened 2024-05-13
Fix diffusion caching

oOraph:
fixes https://github.com/huggingface/optimum-neuron/issues/593

### DIFF
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 """NeuroStableDiffusionPipeline class for inference of diffusion models on neuron devices."""
 
+import copy
 import importlib
 import logging
 import os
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling_diffusion.py
```diff
@@ -698,6 +699,7 @@ def _export(
             save_dir = TemporaryDirectory()
             save_dir_path = Path(save_dir.name)
             # 1. Fetch all model configs
+            input_shapes_copy = copy.deepcopy(input_shapes)
```
michaelbenayoun, line 702: nit: maybe just shadow `input_shapes` and do:
```suggestion
            input_shapes = copy.deepcopy(input_shapes)
```

oOraph: this will shadow it but before it's actually updated in the call below, so the problem won't be fixed, no ?

michaelbenayoun: Ok, no worries then, let's keep it like that.

michaelbenayoun (approved):
Gently pinging @dacorvo since INF2 test do not pass. It seems to be unrelated but double checking.

dacorvo:
I need to update the TGI neuron test repos as it seems that the CI user does not have the rights to create them.

PR #595, opened 2024-05-15
Fix excessive CPU memory consumption on TGI startup

dacorvo:
# What does this PR do?

When launching a TGI instance with a non-neuron model as parameter, the model needs to be exported from cached neuron artifacts during the container startup.

Before this change, the export was done without minimizing the CPU memory, which made it impossible to use this kind of "on-the-fly" export on the smaller `ml.inf2.xlarge` instances.

### DIFF
--- a/text-generation-inference/server/text_generation_server/model.py
+++ b/text-generation-inference/server/text_generation_server/model.py
@@ -6,7 +6,7 @@ from typing import Optional
 from huggingface_hub import snapshot_download
 from huggingface_hub.constants import HF_HUB_CACHE
 from loguru import logger
-from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, GenerationConfig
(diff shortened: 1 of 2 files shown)


text-generation-inference/server/text_generation_server/model.py
```diff
@@ -91,7 +91,7 @@ def fetch_model(
         # Prefetch the neuron model from the Hub
         logger.info(f"Fetching revision [{revision}] for neuron model {model_id} under {HF_HUB_CACHE}")
         log_cache_size()
-        return snapshot_download(model_id, revision=revision)
+        return snapshot_download(model_id, revision=revision, ignore_patterns="*.bin")
```
michaelbenayoun, line 94: Nice!

michaelbenayoun (approved):

PR #608, opened 2024-05-23
Fix pipelines

dacorvo:
# What does this PR do?

This fixes a regression in Neuron pipelines with `transformers == 4.40.2`.

The issue is actually fixed in transformers as a side-effect of this pull-request:
https://github.com/huggingface/transformers/pull/30534
But the issue in `optimum-neuron` actually comes from the fact that all neuron models are based on `optimum.OptimizedModel` that reimplements `transformers.PretrainedModel` without providing all its methods (including `to()`).

This pull-request therefore introduces a new top-level `NeuronModel` class that inherits from `optimum.OptimizedModel` and provides the missing methods.

All neuron model classes should now inherit from `NeuronModel`.

In the process, `NeuronBaseModel` is renamed to `NeuronTracedModel` to remove any ambiguity.


optimum/neuron/modeling_base.py
```diff
@@ -12,600 +12,29 @@
...
-        self,
-        model: torch.jit._script.ScriptModule,
-        config: "PretrainedConfig",
-        model_save_dir: Optional[Union[str, Path, TemporaryDirectory]] = None,
-        model_file_name: Optional[str] = None,
-        preprocessors: Optional[List] = None,
-        neuron_config: Optional["NeuronDefaultConfig"] = None,
-        **kwargs,
-    ):
+    def __init__(self, model: "PreTrainedModel", config: "PretrainedConfig"):
```
JingyaHuang, line 29: For traced models, we are passing `torch.jit._script.ScriptModule` as model here.

optimum/neuron/modeling_base.py
```diff
@@ -12,600 +12,29 @@
...
-        private: Optional[bool] = None,
-        revision: Optional[str] = None,
-        use_auth_token: Union[bool, str] = True,
-        endpoint: Optional[str] = None,
-    ) -> str:
-        if isinstance(use_auth_token, str):
-            huggingface_token = use_auth_token
-        elif use_auth_token:
-            huggingface_token = HfFolder.get_token()
+        if hasattr(model, "device"):
```
JingyaHuang, line 31: Is it necessary, for traced model setting all dummy device as CPU would be fine I think.

optimum/neuron/modeling_base.py
```diff
@@ -12,600 +12,29 @@
...
-        return outputs
+            self.device = torch.device("cpu")
 
-    @property
-    def is_weights_neff_separated(self) -> bool:
-        """
-        Whether the Neuron model has separated weights and neff graph (by setting `inline_weights_to_neff=False` during the compilation).
-        """
-        return not self.config.neuron.get("inline_weights_to_neff", True)
+    def to(self, device: Union[str, torch.device]):
```
JingyaHuang, line 36: Do we need to add content if this is actually a no-op function.

optimum/neuron/modeling_traced.py
```diff
@@ -0,0 +1,611 @@
+# coding=utf-8
```
JingyaHuang, line 1: Maybe just keep it in `modeling_base.py`?

JingyaHuang (commented):
Thx David for opening the PR! Left some nits!

optimum/neuron/modeling_base.py
dacorvo: As long as it exist, someone might want to call it, so I added a check.

optimum/neuron/modeling_base.py
dacorvo: I realized that when the CI failed ... 🫣

JingyaHuang (approved):
LGTM!

PR #609, opened 2024-05-27
Tgi correct clear implementation

dacorvo:
# What does this PR do?

This pull-request fixes an issue when clearing TGI requests.

When a client cancels a TGI request, two different methods can be called on the TGI server:
- if the request is cancelled after prefill, then the router asks the server to "filter" the decoding batch from the corresponding request. This is correctly implemented,
- if the request is cancelled during prefill, then the router asks the server to clear the whole prefill batch. This was not correctly implemented because in that configuration we cleared all requests, even those not included in that prefill batch.

We now remember to which prefill batch each request belongs to be able to clear only the relevant requests.

Note that this pull-request also fixes a bug with models deployed on TGI that do not support continuous batching (i.e. gpt2 and older models).


JingyaHuang (approved):
LGTM, thanks for the fix!

PR #610, opened 2024-05-27
Refactor documentation and improve tgi deployment

dacorvo:
# What does this PR do?

This first refactors the documentation to merge two similar pages related to models export and inference.

This then improves the TGI deployment user experience in several ways:

- also export tokenizer when exporting LLM models,
- remove redundant HF_BATCH_SIZE and HF_SEQUENCE_LENGTH environment variables,
- reduce CPU usage when launching the service on a local model,
- add a dedicated TGI documentation page with simplified instructions,
- add a reference to the export documentation when no cached configuration is found.


docs/source/guides/export_model.mdx
```diff
@@ -40,20 +40,12 @@ AWS provides two generations of the Inferentia accelerator built for machine lea
...
-  (*But these limitations could be bypass with [dynamic batching](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/frameworks/torch/torch-neuronx/api-reference-guide/inference/api-torch-neuronx-trace.html#dynamic-batching) and [bucketing](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/appnotes/torch-neuron/bucketing-app-note.html#bucketing-app-note)*).
-* Neuron models are hardware-specialized, which means:
-  *  Models traced with Neuron can no longer be executed in non-Neuron environment.
-  *  Models compiled for inf1 (NeuronCore-v1) are not compatible with inf2 (NeuronCore-v2), and vice versa.
+Although pre-compilation avoids overhead during the inference, a compiled Neuron model has some limitations:
+* the input shapes and data types used during the compilation cannot be changed.
+* Neuron models are specialized for each hardware and SDK version, which means:
+  * Models compiled with Neuron can no longer be executed in non-Neuron environment.
```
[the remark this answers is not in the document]
dacorvo: That was in the original file. I think it means you cannot run them on CPU.

docs/source/guides/export_model.mdx
```diff
@@ -196,23 +197,20 @@ optimum-cli export neuron --model local_path --task question-answering --batch_s
...
-...   "distilbert-base-uncased-finetuned-sst-2-english", export=True, **input_shapes
-... )
+>>> from optimum.exporters.tasks import TasksManager
+>>> from optimum.exporters.neuron.model_configs import *  # Register neuron specific configs to the TasksManager
 
-# Save the model
->>> model.save_pretrained("./distilbert-base-uncased-finetuned-sst-2-english_neuron/")
+>>> distilbert_tasks = list(TasksManager.get_supported_tasks_for_model_type("distilbert", "neuron").keys())
+>>> print(distilbert_tasks)
+['feature-extraction', 'fill-mask', 'multiple-choice', 'question-answering', 'text-classification', 'token-classification']
```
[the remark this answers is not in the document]
dacorvo: Yes, why not.

text-generation-inference/tgi_env.py
```diff
@@ -184,6 +176,7 @@ def main():
     work properly
     :return:
     """
+    logging.basicConfig(level=logging.DEBUG, force=True)
```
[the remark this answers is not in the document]
dacorvo: Good catch. I need to revert this.

docs/source/guides/export_model.mdx
```diff
@@ -167,9 +159,18 @@ Input shapes:
 
 ```
 
-In the last section, you can see some input shape options to pass for exporting static neuron model, meaning that exact shape inputs should be used during the inference as given during compilation. If you are going to use variable-size inputs, you can pad your inputs to the shape used for compilation as a workaround. If you want the batch size to be dynamic, you can pass `--dynamic-batch-size` to enable dynamic batching, which means that you will be able to use inputs with difference batch size during inference, but it comes with a potential tradeoff in terms of latency.
+### Exporting standard (non-LLM) NLP models
```
JingyaHuang, line 162:
```suggestion
### Exporting non-LLM models
```
The traced model could be applied to ViT, Timm models and audio models as well.

docs/source/guides/models.mdx
```diff
@@ -1,217 +0,0 @@
...
-```
-
-And the next time when you want to run inference, just load your compiled model which will save you the compilation time:
-
-```python
->>> from optimum.neuron import NeuronModelForSequenceClassification
->>> model = NeuronModelForSequenceClassification.from_pretrained("my-neuron-repo")
-```
-
-As you see, there is no need to pass the neuron arguments used during the export as they are
```
JingyaHuang, line 67: I think that we should add the snippet of loading pre-compiled models somewhere (don't feel like it should be in `export_model` neither), it might be obvious for us but better explain properly to users, export is necessary each time, and where compilation args are stored.

docs/source/guides/models.mdx
```diff
@@ -1,217 +0,0 @@
...
-
-tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased-finetuned-sst-2-english")
-inputs = tokenizer("Hamilton is considered to be the best musical of human history.", return_tensors="pt")
-
-logits = model(**inputs).logits
-print(model.config.id2label[logits.argmax().item()])
-# 'POSITIVE'
-```
-
-`compiler_args` are optional arguments for the compiler, these arguments usually control how the compiler makes tradeoff between the inference performance (latency and throughput) and the accuracy. Here we cast FP32 operations to BF16 using the Neuron matrix-multiplication engine.
```
JingyaHuang, line 103: Explaining `compiler_args` is important as well.

docs/source/guides/models.mdx
```diff
@@ -1,217 +0,0 @@
...
->>> neuron_config_constructor = TasksManager.get_exporter_config_constructor(
-...     model=model, exporter="neuron", task='text-classification'
-... )
->>> print(neuron_config_constructor.func.get_mandatory_axes_for_task(task))
-# ('batch_size', 'sequence_length')
-```
-
-<Tip>
-
-Be careful, the input shapes used for compilation should be inferior than the size of inputs that you will feed into the model during the inference.
```
JingyaHuang, line 126: I don't feel like this tip should be removed. We should remind users that padding are put in place, and compile with large static shapes means wasted compute.

docs/source/guides/neuronx_tgi.mdx
```diff
@@ -0,0 +1,183 @@
...
+The service is launched simply by running the neuronx-tgi container with two sets of parameters:
+
+```
+docker run <system_parameters> ghcr.io/huggingface/neuronx-tgi:latest <service_parameters>
+```
+
+- system parameters are used to map ports, volumes and devices between the host and the service,
+- service parameters are forwarded to the `text-generation-launcher`.
+
+When deploying a service, you will need a working Neuron model. The NeuronX TGI service supports two main modes of operation:
```
JingyaHuang, line 41:
```suggestion
When deploying a service, you will need a pre-compiled Neuron model. The NeuronX TGI service supports two main modes of operation:
```

JingyaHuang (commented):
Thanks for reorganising the doc. For the export / inference part, I feel like there are some some tips/reminders of inference shouldn't be removed.

docs/source/guides/models.mdx
dacorvo: What you suggest is a bit different than what was written originally. Can you suggest a new tip and indicate where I should insert it ?

JingyaHuang: something like:
```
Be careful, we pad the inputs to the shapes used for the compilation. But the inputs that you will feed into the model during the inference should have shapes inferior to the static shapes for compilation. And the padding brings computation overhead.
```

under the snippet of re-loading pre-compiled model I think (under line 247)

docs/source/guides/models.mdx
JingyaHuang: sorry, missed that part.

PR #612, opened 2024-05-28
Initial PEFT support

michaelbenayoun:
# What does this PR do?

Adds support for PEFT, tested for LoRA. For now, support is only for DDP. Support for TP will come in another PR.

### DIFF
--- a/optimum/neuron/utils/peft_utils.py
+++ b/optimum/neuron/utils/peft_utils.py
@@ -15,7 +15,7 @@
 """Utilities related to the PEFT library and support."""
 import functools
 import gc
-from typing import Any, List, Optional, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Union
 
 from transformers.utils import is_peft_available
 
@@ -41,6 +41,10 @@ else:
 
     def set_peft_model_state_dict(*args, **kwargs):
         pass
+
+
+if TYPE_CHECKING:
+    pass
 
 
 class NeuronPeftModel(PeftModel):
--- a/optimum/neuron/utils/peft_utils.py
+++ b/optimum/neuron/utils/peft_utils.py
@@ -13,18 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utilities related to the PEFT library and support."""
-import gc
 import functools
+import gc
 from typing import Any, List, Optional, Union
 
 from transformers.utils import is_peft_available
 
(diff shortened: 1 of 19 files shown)


tests/test_peft.py
```diff
@@ -0,0 +1,115 @@
...
+        seed_patcher = create_static_seed_patcher(LlamaForCausalLM, 42)
+        with seed_patcher:
+            _, model = get_tokenizer_and_tiny_llama_model()
+            orig_model_path = output_dir / "orig_peft"
+            orig_peft_model = orig_get_peft_model(model, peft_config)
+
+        orig_peft_model.save_pretrained(orig_model_path.as_posix())
+
+        # PEFT model saved using `NeuronPeftModel`.
+        seed_patcher = create_static_seed_patcher(LlamaForCausalLM, 42)
```
dacorvo, line 91: Do you really need to recreate this ?

optimum/neuron/utils/peft_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        pass
+
+
+if TYPE_CHECKING:
+    pass
+
+
+class NeuronPeftModel(PeftModel):
+    @requires_neuronx_distributed
+    def save_pretrained(
```
dacorvo, line 52: So this is the only method that needs to be implemented differently for Neuron models, right ?
Is it because everything else is already handled by the generic code in accelerator.prepare_model ?

tests/test_peft.py
michaelbenayoun: Yes, otherwise it fails (tried that). Maybe we can make it "reusable" because it's annoying but did not want to spend too much time on that.

dacorvo (commented):
Thank you for the pull-request. It looks good to me, however unless I am mistaken there is no test verifying that the training works, only tests verifying that the model can be peft-ed, prepared, and saved correctly (but maybe I missed something).

JingyaHuang (approved):
LGTM ! Just as David mentioned, a training test would be necessary, and if possible maybe move all peft tests under a tests/peft folder? Then I could put sd peft related stuff under it as well.

michaelbenayoun:
I moved the tests to `tests/peft`.

I wrote a "basic" test that runs training. It does not check about the loss on anything because it's not trivial to overfit with a random model only by finetuning the LoRA adapters. For now this test guarantees that the code runs at least.

michaelbenayoun:
Merging, the issue on TGI is going to be handled in #620

PR #615, opened 2024-05-30
Bump transformers and optimum version

dacorvo:
# What does this PR do?

This bumps `transformers` to `4.41.1` and optimum to `1.20.0`.

This also adds the new `transformers` "stop strings" generation feature to `NeuronModelForCausalLM`, allowing to support the "stop_sequences" parameter in NeuronX TGI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,9 +13,9 @@ except Exception as error:
 
 
 INSTALL_REQUIRES = [
-    "transformers == 4.40.2",
+    "transformers == 4.41.1",
     "accelerate == 0.29.2",
-    "optimum ~= 1.19.1",
+    "optimum ~= 1.20.0",
     "huggingface_hub >= 0.20.1",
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -13,9 +13,9 @@
 
 
 INSTALL_REQUIRES = [
-    "transformers == 4.40.2",
+    "transformers == 4.41.1",
```
JingyaHuang, line 16: could you test before the release (with perhaps test pypi) that `pip install optimum[neuronx]` installs the correct version of optimum and optimum-neuron? thx

JingyaHuang (approved):
LGTM , tho we might need @michaelbenayoun to confirm that bumping trfrs version won't break the training.

dacorvo:
All tests are green except distributed training which was already failing anyway (latest builds were cancelled after several hours).

PR #618, opened 2024-06-01
Skip invalid gen config

dacorvo:
# What does this PR do?

The latest version of `transformers` refuses to save invalid generation config. This triggers an exception when exporting a model to Neuron since we must first save that model to a local directory.

This modifies the export code to check the generation config and recreate it from the model config if it is invalid.


JingyaHuang (approved):
LGTM, thanks!

PR #620, opened 2024-06-03
PEFT + TP support

michaelbenayoun:
# What does this PR do?


optimum/neuron/distributed/utils.py
```diff
@@ -294,15 +293,43 @@ def load_tensor_for_weight(
...
+                # "U32": torch.uint32,
+                "F32": torch.float32,
+                "F64": torch.float64,
+                "I64": torch.int64,
+                # "U64": torch.uint64,
+            }
+            dtype = dtype_str_to_torch_dtype[tensor_slice.get_dtype()]
+            tensor = torch.empty(shape, dtype=dtype)
+            if tensor_slices is not None:
+                tensor = tensor[slices]
```
michaelbenayoun, line 325: This basically allows us to only load the metadata of a given tensor from the safetensors file and not the data itself.
We used that during precompilation to save time as the actual value of weights is not important.

tests/distributed/utils.py
```diff
@@ -14,128 +14,15 @@
...
-        if device is not None:
-            random_labels = random_labels.to(device)
-        labels["labels"] = random_labels
-        if seed is not None:
-            torch.manual_seed(orig_seed)
-    elif model_class_name in [*get_values(MODEL_FOR_CTC_MAPPING_NAMES)]:
-        labels["labels"] = torch.zeros(shape, dtype=torch.float32, device=device)
-    else:
-        raise NotImplementedError(f"Generating the dummy input named for {model_class_name} is not supported yet.")
-    return labels
```
michaelbenayoun, line 138: It was only moved.

tests/distributed/utils.py
```diff
@@ -205,54 +92,3 @@ def gather_along_dim(input_: torch.Tensor, dim: int) -> torch.Tensor:
...
-                    pad_value = pad_token_id
-                else:
-                    pad_value = 1
-                tensor = torch.nn.functional.pad(
-                    tensor,
-                    pad=(0, pad_to_multiple_of - tensor.shape[1] % pad_to_multiple_of),
-                    value=pad_value,
-                )
-                inputs[name] = tensor
-    return inputs
```
michaelbenayoun, line 258: Same, it was moved.

tests/utils.py
```diff
@@ -155,37 +174,40 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-@contextlib.contextmanager
-def create_static_seed_patcher(model_class: Type["PreTrainedModel"], seed: int):
+class StaticSeedPatcher:
```
michaelbenayoun, line 177: @dacorvo I made the context manager reusable!

optimum/neuron/distributed/base.py
```diff
@@ -622,15 +625,39 @@ def should_parallelize_layer_predicate_func(layer):
...
+                    # In this case it means that `param` was parallelized but not `root_tied_param`.
+                    # It will be overwritten by root_tied_param when tiying the weights if we do not do anything.
+                    # We tie `root_tied_param` to `param`.
+                    # What will happen is as follows:
+                    #   - If weight_map contains a checkpoint for `param_name`, it has already been initialized or will
+                    #   be initialized with `cls._maybe_load_weights_to_parallel_linear`.
+                    #   - Otherwise, it has not been initialized and `cls._initialize_or_load_weights` will take care
+                    #   of it.
+                    root_parent_base_mod, _ = get_base_model_and_peft_prefix(root_parent_mod)
+                    setattr(root_parent_base_mod, root_attr_name, param)
```
dacorvo, line 659: It is a bit unclear why setting the param in the base model (and not only in the peft model) will ultimately triggers its parallelization.

optimum/neuron/distributed/utils.py
```diff
@@ -1089,8 +1342,36 @@ def initialize_parallel_linear(mod: "layers.BaseParallelLinear", parameter_names
...
+    for name, param in module.named_parameters():
+        clone.register_parameter(name, torch.nn.Parameter(torch.empty_like(param, device="cpu")))
+
+    clone.__class__ = module.__class__
+    return clone
+
+
 def parameter_can_be_initialized(model: torch.nn.Module, parent_module: torch.nn.Module, parameter_name: str) -> bool:
-    clone = copy.deepcopy(parent_module)
+    # TODO: cannot alwats print the duplicated clone because it does not have all the same attributes.
```
dacorvo, line 1372:
```suggestion
    # TODO: cannot always print the duplicated clone because it does not have all the same attributes.
```

optimum/neuron/utils/peft_utils.py
```diff
@@ -57,57 +92,193 @@ def save_pretrained(
...
+
+                dummy_mod = DummyModule()
+                neuronx_distributed.trainer.save_checkpoint(
+                    output_dir,
+                    tag="adapter_shards",
+                    model=dummy_mod,
+                    async_save=async_save,
+                )
+
+                # Importing here to avoid ciruclar imports.
```
dacorvo, line 185:
```suggestion
                # Importing here to avoid circular imports.
```

dacorvo (commented):
I am missing the knowledge of the peft internals to truly grasp the pull-request, but I think I understood that it includes:
- propagating the neuron parallelization behaviour to the base model instead of the peft model (or to both ?),
- explicitly adding parallelization for the extra lora layers in the peft model,
- adjusting parameter count based on the extra lora layers, taking parameter size into account.
Detected a couple of nits.
It looks like the CI did not run the PEFT tests, perhaps because it failed before that ? Can you check ?

optimum/neuron/distributed/base.py
michaelbenayoun: Basically the peft model contains the base model. They have the same exactly the same parameters.
Here what we are trying to solve is this:

1. We have parameters that are tied before parallelization, meaning they share the same weight. 
2. One of these weight is parallelized during parallelization, meaning that a new parameter is created to replace it.
3. The other parameter has not been replaced so we end-up with two parameters: one sharded and one full weight. 
4. We have a method to re-tie the weights after parallelization / moving the weights to the XLA device. If this method uses the full weight as the "reference" instead of the parallelized one it will overwrite it even though that is the one we actually wanted to keep. 
5. To prevent that, we check all the tied weights right after parallelization. If one of them is left unchanged while the other has been sharded, we updated it to the sharded version.

michaelbenayoun:
> I am missing the knowledge of the peft internals to truly grasp the pull-request, but I think I understood that it includes:
> 
> * propagating the neuron parallelization behaviour to the base model instead of the peft model (or to both ?),
> * explicitly adding parallelization for the extra lora layers in the peft model,
> * adjusting parameter count based on the extra lora layers, taking parameter size into account.
> 
> Detected a couple of nits. It looks like the CI did not run the PEFT tests, perhaps because it failed before that ? Can you check ?

To answer your questions:

- The peft model is just a wrapper around the base model. But yes since the `ParallelizerManagers` is able to only deal with `PreTrainedModel`s to find which subclass to use, we need a bit of work to make it work with `NeuronPeftModel`.
- Indeed, we find the base model from the peft model and parallelize it as before
- On top of that, we also need to shard the LoRA adapters which we do as you understood
- We use a way of counting the parameters that works with both TP and PP and quantization (for QLoRA in the future?) for peft models

optimum/neuron/distributed/base.py
michaelbenayoun: The peft model does not contain other parameters than the base model. So when I overwrite the parameter, I need to parent module of the parameter. It might be not useful here to `get_base_model_and_peft_prefix` here (I do not remember which occasion brought me to do that) but it's a no-op if the model is already a base model.

dacorvo (approved):
Thank you for the pull-request. We need to do something for the distributed tests though ...

michaelbenayoun:
Yes, I will work on it, it is annoying.

PR #630, opened 2024-06-13
Do not copy local checkpoint

dacorvo:
# What does this PR do?

This avoids a redundant copy when a model is exported from a local path. This is made possible by the fact that we don;t need to split the model weights anymore.

### DIFF
--- a/tests/generation/test_export.py
+++ b/tests/generation/test_export.py
@@ -13,8 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from tempfile import TemporaryDirectory
+
 import pytest
 from generation_utils import check_neuron_model
+from transformers import AutoModelForCausalLM
 
 from optimum.neuron import NeuronModelForCausalLM, NeuronModelForSeq2SeqLM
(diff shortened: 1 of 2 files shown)


tests/generation/test_export.py
```diff
@@ -30,15 +33,21 @@
...
+@pytest.mark.parametrize("local", [True, False], ids=["local", "from_hub"])
+def test_decoder_export(local, export_decoder_id, batch_size, sequence_length, num_cores, auto_cast_type):
+    export_kwargs = {
+        "batch_size": batch_size,
+        "sequence_length": sequence_length,
+        "num_cores": num_cores,
+        "auto_cast_type": auto_cast_type,
+    }
+    if local:
+        with TemporaryDirectory() as model_path:
```
michaelbenayoun, line 45: I recently discovered the tmpdir pytest that can be useful as well.

michaelbenayoun (approved):

tests/generation/test_export.py
dacorvo: Indeed, thanks ! I can't use it here because I only need it if local=true

PR #632, opened 2024-06-14
Make neuron_cc_optlevel `None` by default

michaelbenayoun:
# What does this PR do?

As per title. If the user does not specify an optimization level explicitly for the compiler, none is set. Before it was set to `2` by default which can slow the compilation.


dacorvo (approved):
LGTM, thanks !

PR #634, opened 2024-06-14
Fix gradient checkpointing with PEFT

michaelbenayoun:
# What does this PR do?

As per title.
Without this fix gradient checkpointing will fail.


JingyaHuang (approved):
LGTM, thanks for fixing! Don't we have a test that is supposed to catch this?

michaelbenayoun:
It's not that easy to test.

PR #635, opened 2024-06-14
Set bf16 to true when needed

michaelbenayoun:
# What does this PR do?

Sets the `bf16` flag when its relevant.

### DIFF
--- a/optimum/neuron/training_args.py
+++ b/optimum/neuron/training_args.py
@@ -179,7 +179,7 @@ class NeuronTrainingArgumentsMixin:
         )
 
         # If the user did not specify bf16=True but the flags are set, we set bf16=True.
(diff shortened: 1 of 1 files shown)


optimum/neuron/training_args.py
```diff
@@ -177,6 +178,25 @@ def __post_init__(self):
             async_save=self.async_save,
         )
 
+        # If the user did not specify bf16=True but the flags are set, we set bf16=True.
+        # Without this we can fall in the case where XLA will compile the graph in bf16 with torch.finfo unpacthed,
```
dacorvo, line 182:
```suggestion
        # Without this we can fall in the case where XLA will compile the graph in bf16 with torch.finfo unpatched,
```

dacorvo (approved):
LGTM. Just a small nit.

PR #636, opened 2024-06-17
Add in option to pass in num_proc parameter for data packing

BaiqingL:
# What does this PR do?

Add in the option to pass in a `num_proc` parameter to the `pack_dataset` function, which can potentially allow for more throughput when packing input tokens.


BaiqingL:
Hi @philschmid , are there any concerns or blockers regarding this option? The default value is still supplied as 1 so it shouldn't interfere with anything current.

philschmid:
Hey @BaiqingL, 

Thank you for adding this. We want to keep the examples simple and avoid too many optional parameters for advanced usage patterns. Feel free to add those to a fork or the example you use.

PR #642, opened 2024-06-26
CI cache builder

dacorvo:
# What does this PR do?

This adds a CI workflow to build the inference cache nightly or on-demand.

### DIFF
--- a/.github/workflows/inference_cache.yml
+++ b/.github/workflows/inference_cache.yml
@@ -1,0 +1,61 @@
+name: Optimum neuron inference cache builder
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Schedule the workflow to run every day at midnight UTC
+    - cron: '0 0 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+
+jobs:
+  cache:
+    name: Create optimum-neuron inference cache
+    runs-on: [self-hosted, 12-aws-inf2, 192-cpu, ci] # run the job on the newly created runner
+    env:
+      AWS_REGION: us-east-1
+    strategy:
+      fail-fast: false
+      matrix:
+        config: [
+          gpt2,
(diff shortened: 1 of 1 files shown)


dacorvo:
Example workflow:
https://github.com/huggingface/optimum-neuron/actions/runs/9691995166
Some of the jobs are failing because of a lack of disk size (only 200 G). This is being fixed by @glegendre01

.github/workflows/inference_cache.yml
```diff
@@ -0,0 +1,61 @@
...
+          python -m pip config set global.extra-index-url https://pip.repos.neuron.amazonaws.com
+      - name: Install optimum neuron
+        run: |
+          source aws_neuron_venv_pytorch/bin/activate
+          python -m pip install .[neuronx]
+      - name: Create cache for ${{matrix.config}} models
+        run: |
+          source aws_neuron_venv_pytorch/bin/activate
+          config_prefix_url=https://huggingface.co/aws-neuron/optimum-neuron-cache/raw/main/inference-cache-config
+          HF_TOKEN=${{secrets.HF_TOKEN_OPTIMUM_NEURON_CACHE}} \
```
JingyaHuang, line 60: This token has writing access to `aws-neuron/optimum-neuron-cache`?

JingyaHuang (approved):
LGTM ! Thanks David, I shall add it for SD models as well.

.github/workflows/inference_cache.yml
michaelbenayoun: Good question!

michaelbenayoun (approved):
LGTM, that's awesome!

.github/workflows/inference_cache.yml
dacorvo: Yes, and it is fine-grained, so this is the only repo it can write to. I generated it from my own account because it needs to be able to read weights from gated repos.

PR #646, opened 2024-06-28
Restore optimized attention score for sd15 & fix the generated images quality issue

JingyaHuang:
# What does this PR do?

We have previously [overridden the attention processor](https://github.com/huggingface/optimum-neuron/blob/2c524dfa392cef756f3625688f45a00299c424db/optimum/exporters/neuron/utils.py#L391), this was ok but unnecessary(could be useful for torch 2.0 but we skipped it) until we bumped to torch 2.1 where `AttnProcessor2_0` with scaled dot-product attention shall be applied.

This PR restore the optimized attention score and remove the codes overriding the attention processor of the unet to fix the quality issue.

## Before submitting


dacorvo (approved):
Thank you for the pull-request ! LGTM

PR #648, opened 2024-07-03
Fix MPMD detected error during training with TP

michaelbenayoun:
# What does this PR do?

This PR tries to fix the issue reported here https://github.com/aws-neuron/neuronx-distributed/issues/24.

It seems to be linked to `torch.autocast`:

- Before this PR, when it was supposed to be disabled we would do: `with torch.autocast(enabled=False):`
- Now we do: `with contexlib.nullcontex():`. 

For some reason the first approach was leading to different cores trying to execute different graphs, leading to the MPMD detected error.


dacorvo (approved):
LGTM, thanks !

PR #649, opened 2024-07-05
Fix consolidation for TP

michaelbenayoun:
# What does this PR do?

Consolidation was happening on the XLA device if the checkpoint was saved with `xser`. This could lead to OOM if the checkpoint data was bigger than the device memory, which is often the case. This PR fixes that.

cc @philschmid


philschmid (approved):
LGTM!

PR #651, opened 2024-07-11
Add more audio models: ast, hubert, unispeech, unispeech-sat, wavlm

JingyaHuang:
# What does this PR do?

Add more audio models based on our support for wav2vec2.

## Before submitting


optimum/exporters/neuron/model_configs.py
```diff
@@ -423,11 +428,144 @@ def inputs(self) -> List[str]:
...
+#         "audio-classification",
+#         "audio-frame-classification",
+#         "audio-xvector",
+#     ],
+# )
+# class Wav2Vec2BertNeuronConfig(Wav2Vec2NeuronConfig):
+#     pass
+
+
+# TODO: compilation failed due to a bug in xla, https://github.com/huggingface/transformers/pull/31912 might be a workaround.
```
dacorvo, line 540: I think the workaround has been rejected.

dacorvo (approved):
Nice, thanks !

optimum/exporters/neuron/model_configs.py
JingyaHuang: yeah it is, we got to wait for the fix first on PyTorch XLA, then on Neuron SDK, it will be a long wait to go I guess... Considering if I should remove them...

dacorvo (approved):

PR #654, opened 2024-07-16
Various fixes for training

michaelbenayoun:
# What does this PR do?

- Fixes the MPMD that happens when `logging_steps < num_step_in_epoch` at the end of the epoch.
- Fixes `has_write_access_to_repo`

### DIFF
--- a/optimum/neuron/utils/cache_utils.py
+++ b/optimum/neuron/utils/cache_utils.py
@@ -12,7 +12,6 @@
 # limitations under the License.
 """Utilities for caching."""
 
-import functools
 import os
 import re
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


optimum/neuron/utils/cache_utils.py
```diff
@@ -127,7 +129,59 @@ def is_private_repo(repo_id: str) -> bool:
...
+        return True  # has write access, otherwise would have been 403 forbidden.
+    except HfHubHTTPError as e:
+        if e.response.status_code == 403:
+            return False
+        else:
+            raise ValueError(f"Cannot determine write access to {repo_id}") from e
+
+
+@cache_has_write_access
+def has_write_access_to_repo_without_api_action(repo_id: str) -> bool:
```
JingyaHuang, line 177: where do you plan to use it?

JingyaHuang (approved):
Not familiar with the MPMD issue, the changes look good to me in general.

optimum/neuron/utils/cache_utils.py
michaelbenayoun: I actually do not plan to use it, it is more like if we need that in the future. We can remove that as well.

michaelbenayoun: Removed it!

PR #656, opened 2024-07-17
Fix spelling in error message

jimburtoft:
This PR changes the spelling of Talse to False.

I THINK that Talse should be False here.  The boolean flips depending on whether we are looking at inline_weights_to_neff or is_weights_neff_separated.  

It looks like the only time you would get to this is if inline_weights_to_neff is set to True, so the recommendation would be to switch it to False.

Fixes # (issue)
N/A

## Before submitting
- [X ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

### DIFF
--- a/optimum/neuron/utils/misc.py
+++ b/optimum/neuron/utils/misc.py
@@ -549,7 +549,7 @@ def replace_weights(
     prefix: str = "model",
 ):
     """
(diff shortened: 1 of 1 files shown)


JingyaHuang (approved):
Thanks for correcting it!

dacorvo (approved):
Fhank you for this pull-requesf !

PR #659, opened 2024-07-18
Update batch_size error message.

jimburtoft:
The previous message mentions max-input-length, even though we don't do any checks for that.  

I only see this message when I compiled the model with a different batch size (or set the wrong batch size in MAX_BATCH_SIZE), so I want to add specific instructions to fix that failure case.  I can't think of any other way it could happen.

I think this message also make sense in the SageMaker environment.

Things to think about:
1. Any length limits on the error messages?  I figure more descriptive is better.  If it does get truncated it still makes sense.
2. Does it make sense to be more strict with the batch size check?   Currently, if you compiled with a smaller batch size, it would still pass this check.  That is OK for a compiled batch size of one, but does it need to be mod configured batch size or anything?
3. Did you really want to add some kind of a check on max-input-length / model.max_length and this was a forward looking message?

Fixes # (issue)

## Before submitting

### DIFF
--- a/text-generation-inference/server/text_generation_server/generator.py
+++ b/text-generation-inference/server/text_generation_server/generator.py
@@ -351,7 +351,7 @@ class NeuronGenerator(Generator):
         batch_size = self.model.batch_size
         if len(batch.requests) > batch_size:
             raise ValueError(
-                f"Inconsistent server configuration: please make sure max-prefill-tokens does not exceed {batch_size} x max-input-length."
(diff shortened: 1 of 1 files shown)


jimburtoft:
@dacorvo for review.

dacorvo (approved):
LGTM, thanks !

dacorvo:
This message comes from the old days when TGI did not accept the max_batch_size parameter and we need to use indirect means to control it.
I think all parameters could be checked when we load the model: maybe in another pull-request.

jimburtoft:
Hey @dacorvo , looks like this never got merged.  Maybe add "checking all parameters" to a todo.  The goal of this PR is to just give a specific helpful error message in this particular case.

jimburtoft:
Hey @dacorvo, just pinging no this error message update.  Looks like you approved it but it is blocked by a failing check?  I double checked my quotes.

dacorvo:
@jimburtoft the error is because your credentials are not sufficient for some of the tests.

PR #670, opened 2024-07-25
Support resize embeddings

michaelbenayoun:
# What does this PR do?

This PR adds support for resizing embedding when using TP. The new size needs to be divisible by the TP size.


JingyaHuang (approved):
LGTM. Thanks for adding the feature!

dacorvo (approved):
One small nit but otherwise LGTM, thanks !

PR #673, opened 2024-07-30
Examples for E2E Mistral training & deployment

jonetiz:
# What does this PR do?
Adds examples which are used for an upcoming AWS Machine Learning blog.


examples/mistral-e2e/examples/run_clm.py
michaelbenayoun, line 1: Is it the `run_clm.py` script we already provide?

michaelbenayoun (commented):
Overall it is quite different than our examples which are the equivalent of the examples in Transformers. 
Should we create a directory like "aws-examples" or something where we would put them?

cc @philschmid @pagezyhf

examples/mistral-e2e/examples/run_clm.py
jonetiz: Distinction is in dataset loading, it uses load_from_disk to load a saved DatasetDict, before downloading from HF hub.

michaelbenayoun: Alright, I do not think this is really needed as a distinction. But that's fair.

michaelbenayoun:
@pagezyhf @philschmid wdyt?

examples/mistral-e2e/examples/chat.py
```diff
@@ -0,0 +1,60 @@
...
+    chat.append({'role': 'user', 'content': message})
+
+    for i in range(0, len(chat), 2):
+        prompt = tokenizer.apply_chat_template(chat[i:], tokenize=False)
+
+        tokens = tokenizer(prompt)
+        if len(tokens.input_ids) <= max_tokens:
+            return prompt
+
+    raise SystemError
```
philschmid, line 29: we should use the `tokenizer.apply_chat_template` method.

examples/mistral-e2e/examples/chat.py
```diff
@@ -0,0 +1,60 @@
+from transformers import AutoTokenizer
+from optimum.neuron import NeuronModelForCausalLM
+
+model = NeuronModelForCausalLM.from_pretrained("./mistral_neuron", local_files_only=True)
```
philschmid, line 4: How `./mistral_neuron` is downloaded?

examples/mistral-e2e/examples/compile.py
```diff
@@ -0,0 +1,15 @@
+from optimum.neuron import NeuronModelForCausalLM
+
+# num_cores is the number of neuron cores. Find this with the command neuron-ls
+compiler_args = {"num_cores": 12, "auto_cast_type": 'bf16'}
+input_shapes = {"batch_size": 1, "sequence_length": 4096}
+
+# Compiles an Optimum Neuron model from the previously trained (uncompiled) model
+model = NeuronModelForCausalLM.from_pretrained(
+    "mistral",
```
philschmid, line 9: What is in this directory?

examples/mistral-e2e/examples/dataset.py
```diff
@@ -0,0 +1,23 @@
+from datasets import DatasetDict, load_dataset
+
+def format(sample):
+    sample['text'] = f"<s>[INST] {sample['question']} [/INST]\n\n{sample['answer']}</s>"
+    return sample
```
philschmid, line 5: similar to the top, we should use `apply_chat_template`

examples/mistral-e2e/examples/run_clm.py
philschmid: yes we should arther update the general on to make it support loading datasets from disk.

examples/mistral-e2e/examples/test.py
philschmid, line 1: isn't that the same as chat?

philschmid (commented):
Thank you for opening the PR. I missing some context. Whats is the goal of that? I cannot see instruction what to do with the yaml file or how the scripts are used. If want to make an official example we should make sure that's understandable and useable for every user.

jonetiz:
> Thank you for opening the PR. I missing some context. Whats is the goal of that? I cannot see instruction what to do with the yaml file or how the scripts are used. If want to make an official example we should make sure that's understandable and useable for every user. 
> 
> 
> 
> 

So this is for an AWS Machine Learning blog, to give beginners to intermediates an idea of how to use AWS Trainium and AWS Inferentia2 for simple end-to-end training and inferencing. 

The YAML file is an Amazon CloudFormation template - it doesn't _necessarily_ need to be included, so I can take that out.

examples/mistral-e2e/examples/chat.py
jonetiz: We wanted to show how formatting works to the reader, hence not using apply_chat_template here.

examples/mistral-e2e/examples/chat.py
jonetiz: mistral_neuron is outputted from the compiler, the idea is to transfer the fine-tuned model from the Trainium instance to Inferentia via S3.

examples/mistral-e2e/examples/compile.py
jonetiz: The fine tuned model downloaded from S3

examples/mistral-e2e/examples/test.py
jonetiz: We wanted to walk readers of the blog through making a simple test to show the fine tuning works - chat.py is a larger example to be a proper chat application including context memory

michaelbenayoun:
I would say let's me the tutorial as minimalist as possible and put it in an `aws-examples` folder. We can still update that afterwards.

examples/mistral-e2e/examples/run_clm.py
michaelbenayoun: Honestly I think we should keep examples easy. Saving the dataset on disk is not really needed. There is a caching mechanism, we should just push them to use the dataset id and let the Datasets library handle the rest.

aws-examples/mistral-e2e/chat.py
```diff
@@ -0,0 +1,62 @@
+# This is a simple command-line chat application that has contextual memory.
+
```
michaelbenayoun, line 2: Can you add a licence here (as well as the other python files) please? You can check our examples.

aws-examples/mistral-e2e/chat.py
```diff
@@ -0,0 +1,62 @@
...
+    chat.append({'role': 'user', 'content': message})
+
+    for i in range(0, len(chat), 2):
+        prompt = tokenizer.apply_chat_template(chat[i:], tokenize=False)
+
+        tokens = tokenizer(prompt)
+        if len(tokens.input_ids) <= max_tokens:
+            return prompt
+
+    raise SystemError
```
michaelbenayoun, line 31: Why?

aws-examples/README.md
michaelbenayoun, line 1: Nice!

aws-examples/mistral-e2e/chat.py
```diff
@@ -0,0 +1,62 @@
...
+
+    print(answer)
+
+    history.append([message, answer])
+
+    chat(history, max_tokens)
+
+if __name__ == "__main__":
+    history = []
+    max_tokens=4096
```
michaelbenayoun, line 61:
```suggestion
    max_tokens = 4096
```

aws-examples/mistral-e2e/compile.py
```diff
@@ -0,0 +1,15 @@
...
+compiler_args = {"num_cores": 12, "auto_cast_type": 'bf16'}
+input_shapes = {"batch_size": 1, "sequence_length": 4096}
+
+# Compiles an Optimum Neuron model from the previously trained (uncompiled) model
+model = NeuronModelForCausalLM.from_pretrained(
+    "mistral_trained",
+    export=True,
+    **compiler_args,
+    **input_shapes)
+
```
michaelbenayoun, line 13:
```suggestion
    **input_shapes
)

```

examples/mistral-e2e/examples/run_clm.py
jonetiz: @michaelbenayoun 

In our example, gsm8k is not pre-formatted for run_clm.py, which uses a single text column for training data, so the dataset needs to be formatted.

The reason we're loading from disk is to show customers (in the blog post) how the dataset needs to be formatted by leading them through it step-by-step (which ends up being dataset.py). While we can modify the training script to shape the dataset as necessary, we thought it'd be easier to show it in a standalone file - for the purposes of our blog post, run_clm.py isn't discussed in detail (the idea being it's plug and play with any LLM or dataset) so simplicity isn't necessarily our concern there.

Technically speaking, we could modify run_clm.py to shape the data itself, but I'd also argue it's better to have a training script that can take in any pre-formatted dataset, rather than having to modify it every time for new datasets. Hope this helps you see my point of view.

aws-examples/mistral-e2e/chat.py
jonetiz: Will add additional comments for explanation in upcoming commit.

PR #661, opened 2024-07-22
Use AWS Neuron SDK 2.19.1

dacorvo:
# What does this PR do?

This bumps the AWS Neuron SDK to 2.19.1.

This new SDK mainly improves the parallelization of context encoding when using continuous batching for LLM.

PR #685, opened 2024-08-27
Add workflow for caching stable diffusion models

JingyaHuang:
# What does this PR do?

* Add a workflow independent to the llm one (sd models don't need 48xlarge instance since the compilation is not leveraging neuron devices, but the validation does, for which 8xlarge is enough.)
* The workflow for SD is triggered automatically weekly on Saturday.

## Before submitting


.github/workflows/inference_cache_stable_diffusion.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Install Neuron runtime
+        run: |
+          . /etc/os-release
+          sudo tee /etc/apt/sources.list.d/neuron.list > /dev/null <<EOF
+          deb https://apt.repos.neuron.amazonaws.com ${VERSION_CODENAME} main
+          EOF
+          wget -qO - https://apt.repos.neuron.amazonaws.com/GPG-PUB-KEY-AMAZON-AWS-NEURON.PUB | sudo apt-key add -
+          sudo apt-get update -y
+          sudo apt-get install aws-neuronx-tools=2.18.3.0 aws-neuronx-runtime-lib=2.21.41.0-fb1705f5f aws-neuronx-collectives=2.21.46.0-69b77134b  -y
```
dacorvo, line 32: These packages are obsolete: you should use the ones from 2.19.1 (this is true also for the llm cache workflow)

.github/workflows/inference_cache_stable_diffusion.yml
```diff
@@ -0,0 +1,52 @@
...
+    - cron: '0 0 * * 6'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+
+jobs:
+  cache:
+    name: Create optimum-neuron inference cache
+    runs-on:
+      group: aws-inf2-8xlarge
```
dacorvo, line 16: Is this group using the latest driver corresponding to 2.19.1 ?

JingyaHuang: I think so, we have bump to 2.19.1 for a while(since #661 ), and the CIs for inference have been using the group for inference tests, so shall be ok right?

.github/workflows/inference_cache_stable_diffusion.yml
JingyaHuang: How are these obsolete? The versions in `inference_cache_stable_diffusion.yml` are version numbers I copied from my local 2.19.1 env, do you mean that I shall use the[ ones in AWS Neuron doc](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/release-notes/index.html#trn1-packages), those without the suffix?

JingyaHuang: The versions in the LLM workflow files seem to come from [Neuron 2.18.2](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/release-notes/prev/content.html#neuron-2-18-2-04-25-2024).

dacorvo: You're right: the versions you used are correct. It is only those of the LLM workflow that have not been updated.

.github/workflows/inference_cache_stable_diffusion.yml
dacorvo: I am not sure they are not up to date, because we did not request the infra team to bump the driver version, so unless they do it automatically, it is still using 2.18 driver.

dacorvo (approved):

.github/workflows/inference_cache_stable_diffusion.yml
JingyaHuang: will check with the infra team.

PR #686, opened 2024-08-28
Bump Optimum to 1.22 + Adapt to the SD task refactoring in Optimum main

JingyaHuang:
# What does this PR do?

Make optimum-neuron compatible with changes in https://github.com/huggingface/optimum/pull/1947. Besides, with the changes, the `InstructPix2Pix` support in Optimum Neuron will fail, a PR is opened in Optimum repo to patch this: https://github.com/huggingface/optimum/pull/1999. 

Merge this only when we bump to a stable Optimum release including the two PRs mentioned above.

## Before submitting


tests/inference/test_modeling.py
```diff
@@ -13,1884 +13,1512 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import gc
-import os
```
dacorvo, line 16: Why so many lines were removed and other commented out ?

dacorvo (commented):
Thanks for the pull-request. It looks good to me except for the changes in test_modleing.py that I did not understand.

tests/inference/test_modeling.py
JingyaHuang: Good catch, I accidentally remove them for debugging a specific test.... Putting it back.

dacorvo (approved):
LGTM, thank you for the pull-request !

Issue #698, opened 2024-09-20
optimum neuron optimum-cli compile fails in newest TGI Neuron


jimburtoft:
### System Info

```shell
!docker run -p 8080:80 \
-v $(pwd):/data \
--device=/dev/neuron0 \
--device=/dev/neuron1 \
--device=/dev/neuron2 \
--device=/dev/neuron3 \
--device=/dev/neuron4 \
--device=/dev/neuron5 \
-ti \
--entrypoint "optimum-cli" neuronx-tgi:latest \
env


Platform:

- Platform: Linux-5.15.0-1056-aws-x86_64-with-glibc2.35
- Python version: 3.10.12


Python packages:

- `optimum-neuron` version: 0.0.25.dev0
- `neuron-sdk` version: 2.20.0
- `optimum` version: 1.21.4
- `transformers` version: 4.43.2
- `huggingface_hub` version: 0.25.0
- `torch` version: 2.1.2+cu121
- `aws-neuronx-runtime-discovery` version: 2.9
- `libneuronxla` version: 2.0.4115.0
- `neuronx-cc` version: 2.15.128.0+56dc5a86
- `neuronx-distributed` version: NA
- `neuronx-hwm` version: NA
- `torch-neuronx` version: 2.1.2.2.3.0
- `torch-xla` version: 2.1.4
- `transformers-neuronx` version: 0.12.313


Neuron Driver:


WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

aws-neuronx-collectives/now 2.22.26.0-17a033bc8 amd64 [installed,local]
aws-neuronx-dkms/now 2.18.12.0 amd64 [installed,local]
aws-neuronx-runtime-lib/now 2.22.14.0-6e27b8d5b amd64 [installed,local]
aws-neuronx-tools/now 2.19.0.0 amd64 [installed,local]
```

### Who can help?

@dacorvo 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [X] My own task or dataset (give details below)

### Reproduction (minimal, reproducible, runnable)

When I use the newly merged TGI image to compile with optimum-cli, I get an error message.

I haven't been able to test it without TGI because I am having trouble upgrading my image to 2.20.

```!git clone https://github.com/huggingface/optimum-neuron.git && cd optimum-neuron && make neuronx-tgi```

[...]

PR #700, opened 2024-09-23
Fix compilation inside tgi

dacorvo:
# What does this PR do?

This addresses #698 .

The root cause is that the `libxml2` package is missing from the TGI image, while it is required by the neuronx compiler (but not pulled automatically).


JingyaHuang (approved):
LGTM, thx David!

PR #702, opened 2024-09-24
Fix PP issue

michaelbenayoun:
# What does this PR do?

As per title.

### DIFF
--- a/tests/test_trainers.py
+++ b/tests/test_trainers.py
@@ -27,6 +27,7 @@ from transformers import (
     AutoConfig,
     AutoModelForSequenceClassification,
 )
+from transformers.testing_utils import is_staging_test
 
 from optimum.neuron import NeuronSFTConfig, NeuronSFTTrainer, NeuronTrainer, NeuronTrainingArguments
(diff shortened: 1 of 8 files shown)


tests/test_trainers.py
```diff
@@ -462,3 +463,9 @@ def test_without_packing(self, parallel_sizes, tmpdir):
 
     def test_with_packing(self, parallel_sizes, tmpdir):
         return self._test_sft_trainer(parallel_sizes, tmpdir, True)
+
+
+@is_trainium_test
```
dacorvo, line 468: I think this is a WIP, right ?

dacorvo (approved):
LGTM, although this does not seem to fix the main issue with the neverending workflows.

tests/test_trainers.py
michaelbenayoun: No it's just that pytest fails with error code 5 when it does not collect staging and trainium tests. So I just created a dummy test.

michaelbenayoun:
Yes very hard to debug the neverending issue.

Issue #456, opened 2024-02-01
export_model guide disagrees with codebase


timgautier:
The guide shows the usage statement here: https://github.com/huggingface/optimum-neuron/blob/eb2a93fe89721a3977957cb3748832bf9d48e382/docs/source/guides/export_model.mdx#L90

usage: optimum-cli export neuron [-h] -m MODEL [--task TASK] [--atol ATOL] [--cache_dir CACHE_DIR] [--trust-remote-code]
                                 [--compiler_workdir COMPILER_WORKDIR] [--disable-validation] [--auto_cast {none,matmul,all}]
                                 [--auto_cast_type {bf16,fp16,tf32}] [--dynamic-batch-size] [--num_cores NUM_CORES] [--unet UNET]
                                 [--output_hidden_states] [--output_attentions] [--batch_size BATCH_SIZE]
                                 [--sequence_length SEQUENCE_LENGTH] [--num_beams NUM_BEAMS] [--num_choices NUM_CHOICES]
                                 [--num_channels NUM_CHANNELS] [--width WIDTH] [--height HEIGHT]
                                 [--num_images_per_prompt NUM_IMAGES_PER_PROMPT] [-O1 | -O2 | -O3]
                                 output

However, running `optimum-cli export neuron --help` yields this usage statement:

usage: optimum-cli export neuron [-h] -m MODEL [--task TASK] [--library-name {transformers,diffusers,sentence_transformers}] [--subfolder SUBFOLDER] [--atol ATOL]
                                 [--cache_dir CACHE_DIR] [--trust-remote-code] [--compiler_workdir COMPILER_WORKDIR] [--disable-validation] [--auto_cast {none,matmul,all}]
                                 [--auto_cast_type {bf16,fp16,tf32}] [--dynamic-batch-size] [--unet UNET] [--output_hidden_states] [--output_attentions] [--batch_size BATCH_SIZE]
                                 [--sequence_length SEQUENCE_LENGTH] [--num_beams NUM_BEAMS] [--num_choices NUM_CHOICES] [--num_channels NUM_CHANNELS] [--width WIDTH] [--height HEIGHT]
                                 [--num_images_per_prompt NUM_IMAGES_PER_PROMPT] [-O1 | -O2 | -O3]
                                 output

[...]

PR #703, opened 2024-09-24
fixes --library-name parameter in example for sentence_transformers.mdx

jimburtoft:
Looks like we stopped letting users pass in --library-name at some point.  This is the only place I could find it in the docs.  We infer it from the model type.  

If I run the example:
```
optimum-cli export neuron -m Snowflake/snowflake-arctic-embed-l --library-name sentence_transformers --sequence_length 1024 --batch_size 1 --task feature-extraction snowflake-arctic-embed-l-inf2
/opt/aws_neuron_venv_pytorch/lib/python3.8/site-packages/diffusers/models/transformers/transformer_2d.py:34: FutureWarning: `Transformer2DModelOutput` is deprecated and will be removed in version 1.0.0. Importing `Transformer2DModelOutput` from `diffusers.models.transformer_2d` is deprecated and this will be removed in a future version. Please use `from diffusers.models.modeling_outputs import Transformer2DModelOutput`, instead.
  deprecate("Transformer2DModelOutput", "1.0.0", deprecation_message)
usage: optimum-cli
Optimum CLI tool: error: unrecognized arguments: --library-name snowflake-arctic-embed-l-inf2

```

If I remove --library-name, the example compiles.

In the code example, they show it being passed in.  I assume that still works because I see it being used in the tests but I didn't check.

I assume this will change will eventually fix https://huggingface.co/docs/optimum-neuron/en/inference_tutorials/sentence_transformers 

# What does this PR do?

Fixes # (issue)
Maybe fixes https://github.com/huggingface/optimum-neuron/issues/456

## Before submitting
- [X ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

### DIFF
--- a/docs/source/inference_tutorials/sentence_transformers.mdx
+++ b/docs/source/inference_tutorials/sentence_transformers.mdx
@@ -49,7 +49,7 @@ model.save_pretrained("bge_emb_inf2/")
 Here we will use the `optimum-cli` to convert the model. Similar to the `NeuronModelForSentenceTransformers` we need to define our input shape and batch size. The input shape is defined by the `sequence_length` and the batch size by `batch_size`. The `optimum-cli` will automatically convert the model to a format compatible with AWS Inferentia2 and save it to the specified output directory.
 
 ```bash
-optimum-cli export neuron -m BAAI/bge-small-en-v1.5 --library-name sentence_transformers --sequence_length 384 --batch_size 1 --task feature-extraction bge_emb_inf2/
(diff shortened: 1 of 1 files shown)


jimburtoft:
@dacorvo

JingyaHuang (approved):
LGTM, thanks @jimburtoft for the fix. Yes, the library is inferred by the Taskmanager now, and there is no need to pass the library name anymore.

PR #705, opened 2024-09-26
Update cached models and benchmarks

dacorvo:
# What does this PR do?

This modifies the cache population workflow to use the files corresponding to new models.

This also updates the benchmark in the documentation.


benchmark/text-generation/mistral_small.py
```diff
@@ -27,7 +25,7 @@ def main():
             export=True,
             batch_size=batch_size,
             sequence_length=seq_length,
-            auto_cast_type="fp16",
+            auto_cast_type="bf16",
```
JingyaHuang, line 28: So it's now proven that bf16 is better?

JingyaHuang (approved):
LGTM, thanks for the updates!

benchmark/text-generation/mistral_small.py
dacorvo: It actually does not make any difference as far as I can tell, and this seems to allow to load weights faster (because they are not converted).

michaelbenayoun (approved):
LGTM

dacorvo:
I have made a small modification in the export code to always set `fuse_qkv` to True. This ruined my cached test models, so I had to reupload them with that change. I have some errors on docker tests that I don't understand. Let's wait for this new CI run ...

PR #707, opened 2024-09-30
Add tool to list top models

dacorvo:
I wrote a small script to list the top-k Hub models (downloads or trending).

Models can also be filtered by arch and/or task.

This is useful to periodically evaluate the list of models to add to the neuron cache.


JingyaHuang (approved):
LGTM, thanks for adding the tool!

PR #715, opened 2024-10-10
Make the Parallelizer class work for inference

michaelbenayoun:
# What does this PR do?

As per title.

### DIFF
--- a/optimum/neuron/distributed/encoder_decoder_models.py
+++ b/optimum/neuron/distributed/encoder_decoder_models.py
@@ -111,6 +111,7 @@ class T5ParallelMLP(ParallelMLP):
         sequence_parallel_enabled: bool = False,
         device: Optional[torch.device] = None,
         should_parallelize_layer_predicate_func: Optional[Callable[[torch.nn.Module], bool]] = None,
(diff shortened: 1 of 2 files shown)


optimum/neuron/distributed/encoder_decoder_models.py
```diff
@@ -111,6 +111,7 @@ def transform(
         sequence_parallel_enabled: bool = False,
         device: Optional[torch.device] = None,
         should_parallelize_layer_predicate_func: Optional[Callable[[torch.nn.Module], bool]] = None,
+        **parallel_layer_specific_kwargs,
```
JingyaHuang, line 114: as per my experience with T5, the majority of args in `parallel_layer_specific_kwargs` sent to the ``transform()` functions in T5's parallel modules raised errors, eg:
```
TypeError: transform() got an unexpected keyword argument 'skip_linear_weight_load'
```

michaelbenayoun: I think it's just a matter of adding a missing argument here.

dacorvo:
I think we can close this one @michaelbenayoun

PR #717, opened 2024-10-14
Update stale bot

dacorvo:
# What does this PR do?

This updates the stale bot to use an existing github action instead of the custom python script.

This also adds a workflow_dispatch to allow to run the stale bot on demand.


michaelbenayoun (approved):
Thanks for taking care of this!!

PR #725, opened 2024-10-25
Fix wrong inputs/model placement when using a single core

JingyaHuang:
# What does this PR do?

It's a recent find that I did not experience before. Previously, when only a single neuron core was used, input tensors were allocated to the same core that the model has been loaded with `torch.jit.load`. But it seems that now the input tensors are always allocated to nc:0, whereas the model is loaded to a core with lower usage, so we could bump into unexpected flaky situations where they are in different cores... 

```
2024-Oct-25 15:40:20.728276 13974:13974 ERROR  NMGR:dlr_check_tensor_set_on_same_vtpb_v2    Tensor input249 allocated on vnc 0 must be allocated on same vnc as model (1)
```

## Before submitting

### DIFF
--- a/optimum/neuron/modeling_traced.py
+++ b/optimum/neuron/modeling_traced.py
@@ -53,6 +53,7 @@ if is_neuron_available():
     NEURON_COMPILER_VERSION = get_neuroncc_version()
 
 if is_neuronx_available():
+    import torch_neuronx
     from torch_neuronx import move_trace_to_device
 
     NEURON_COMPILER_TYPE = "neuronx-cc"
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling_traced.py
```diff
@@ -127,8 +128,12 @@ def load_model(
         if path.is_file():
             model = torch.jit.load(path)
             # For non-inlined models, send the module manually to device. This is important for weights/neff non-inlined module since when loading the module, the neff is automatically moved to Neuron but not the weights. We need to move the weights to Neuron as well manually to avoid great host to device IO penalty.
-            if is_neuronx_available() and to_neuron:
-                move_trace_to_device(model, device_id)
+            if is_neuronx_available():
+                torch_neuronx.experimental.set_neuron_cores(
+                    model, start_nc=0, nc_count=1
```
dacorvo, line 133: What happens if the model has tp_degree > 1 ?

PR #729, opened 2024-10-29
Add HF_TRANSFER to AMI 

philschmid:
# What does this PR do?

Adds HF_TRANSFER to AMI by default for faster model loading.


dacorvo (approved):
LGTM, thanks !

PR #737, opened 2024-11-17
Added num_cores argument to neuron

gaarutyunov:
# What does this PR do?

Added num_cores argument to optimum-cli export neuron

Fixes # (issue)

Was getting AttributeError: 'Namespace' object has no attribute 'num_cores' on optimum-cli export neuron

## Before submitting

### DIFF
--- a/optimum/commands/export/neuron.py
+++ b/optimum/commands/export/neuron.py
@@ -60,6 +60,12 @@ def parse_args_neuron(parser: "ArgumentParser"):
         default=None,
         help="If specified, the absolute difference tolerance when validating the model. Otherwise, the default atol for the model will be used.",
     )
+    optional_group.add_argument(
+        "--num_cores",
+        type=int,
+        default=1,
+        help="Number of neuron cores used when instantiating the model. Each neuron core has 16 Gb of memory, which means that bigger models need to be split on multiple cores.",
+    )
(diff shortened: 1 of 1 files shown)


dacorvo (commented):
This parameter is not supported for neuron, only for neuronx.

gaarutyunov:
> This parameter is not supported for neuron, only for neuronx.

It is used [here](https://github.com/gaarutyunov/optimum-neuron/blob/main/optimum/exporters/neuron/__main__.py#L709) and results in an AttributeError I've mentioned in the issue.

Should I stick to export neuronx? Is export neuron still supported?

gaarutyunov:
It is also mentioned in the docs [here](https://huggingface.co/docs/optimum-neuron/en/guides/export_model) in the section Exporting LLMs to Neuron.

dacorvo:
@gaarutyunov I agree this is confuzing but really this all applies to neuronx export, even if the name is neuron.

JingyaHuang:
Hi @gaarutyunov, are you using inf1 instances? The export of LLMs is not supported by inferentia 1, you might want to use inferentia 2 instances as an alternative.

gaarutyunov:
> Hi @gaarutyunov, are you using inf1 instances? The export of LLMs is not supported by inferentia 1, you might want to use inferentia 2 instances as an alternative. 

Ok, thank you for clarifying this, it wasn't clear from the documentation. I will try inf2.

Issue #738, opened 2024-11-18
[Diffusers] Regression of CPU memory usage


JingyaHuang:
### Issue

We were able to run SDXL artifacts compiled with inf2.8xlarge on inf2.xLarge (as stated in the [blog](https://www.philschmid.de/inferentia2-stable-diffusion-xl)). However, we recently found that SDXL's CPU memory usage increased, leading to OOM during the inference on inf2.xlarge. In this issue, we will note down some experiment results to trace where the regression was introduced.

### Tasks

- [ ] Latest Optimum Neuron (0.0.26) on Neuron SDK 2.15.0
- [ ] Other Neuron SDK versions
- [ ] Pytorch version 1.13.1 v.s 2.1? 

### Reproduction (minimal, reproducible, runnable)

```bash
optimum-cli export neuron --model stabilityai/stable-diffusion-xl-base-1.0  --batch_size 1 --height 1024 --width 1024 --num_images_per_prompt 4 --auto_cast matmul --auto_cast_type bf16 sd_neuron_xl/
```

### Expected behavior

Being able to fit into inf2.xlarge after being compiled with inf2.8xlarge.

JingyaHuang:
Experiment 2: Neuron sdk 2.16.1 + Optimum Neuron 0.0.18

* Compiled artifacts
* Inference failed with

JingyaHuang:
Experiment 2: Neuron sdk 2.16.1 + Optimum Neuron 0.0.13

* Compiled artifacts: https://huggingface.co/Jingya/sdxl_on_0.0.13_neuron_2.16.1/tree/main
* Inference: succeeded
 -> the regression was introduced between Optimum Neuron 0.0.13 - 0.0.18

JingyaHuang:
Optimum Neuron v0.0.14 + Neuron SDK 2.16.1 ✔️

JingyaHuang:
Optimum Neuron v0.0.15 + Neuron SDK 2.16.1 👎 

[Jingya/sd_neuron_xl_2.16.1_0.0.15](https://huggingface.co/Jingya/sd_neuron_xl_2.16.1_0.0.15)

```
Loading only U-Net into both Neuron Cores...
Killed
```

The regression shall has been introduced between v0.0.14 -> v0.0.15

JingyaHuang:
[...]

PR #742, opened 2024-11-25
Emphasize the order of loading models in diffusion pipelines

JingyaHuang:
# What does this PR do?

as explained in #738


dacorvo (approved):
LGTM, thanks. I think the failing tests are related to the cache bug in 2.20.

PR #743, opened 2024-11-26
chore: bump Neuron SDK to 2.20.2

dacorvo:
# What does this PR do?

This updates the Neuron SDK version used by optimum-neuron to 2.20.2. This will solve the cache issues for models compiled in different virtual environments.

I had CI failures for:
- distributed training test (hangs), but I am pretty sure it is unrelated,
- T5 parallel test, but I ran the test locally and it worked.


JingyaHuang (approved):
LGTM id all CIs are green. Thanks for taking care of it!

PR #748, opened 2024-12-11
Bump TGI router version

dacorvo:
# What does this PR do?

This bumps the TGI router version to v3.0.0.

The migration has been done in several steps to track a regression in performance that was introduced between v2.4.0 and v2.4.1.

The root cause of the regression is the new calculation of the `max_batch_prefill_tokens` that defaults to `4096` for neuron while it was previously equal to `max_batch_size * max_input_tokens`. This pull-request includes a workaround to set it to the correct value if it is not specified on the command line or in the env.

### DIFF
--- a/text-generation-inference/Cargo.toml
+++ b/text-generation-inference/Cargo.toml
@@ -14,7 +14,7 @@ default-members = [
 resolver = "2"
 
 [workspace.package]
-version = "2.1.1"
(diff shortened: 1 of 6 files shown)


michaelbenayoun (approved):
Not very familiar with the whole thing but lgtm

PR #749, opened 2024-12-17
Qwen 2.5 tutorial

pagezyhf:
Trying to build a tutorial using Qwen 2.5 with Inferentia

A few blockers: 
- I don't think we have a compiled version of this model in the cache. At least I couldn't find using optimum-cli neuron lookup. 
- I'd like to have a similar tutorial deploying an endpoint in Sagemaker but I don't know where to put it in the repo. Any ideas? So far all Sagemaker examples live in Philipp's blog, linked via https://huggingface.co/docs/optimum-neuron/en/guides/sagemaker

### DIFF
--- a/docs/source/inference_tutorials/qwen2-5-7b-chatbot.mdx
+++ b/docs/source/inference_tutorials/qwen2-5-7b-chatbot.mdx
@@ -1,0 +1,226 @@
+<!---
+Copyright 2024 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


docs/source/inference_tutorials/qwen2-5-7b-chatbot.mdx
```diff
@@ -0,0 +1,226 @@
...
+
+Depending on your choice of parameters and inferentia host, this may take from a few minutes to more than an hour.
+
+For your convenience, we host a pre-compiled version of that model on the Hugging Face hub, so you can skip the export and start using the model immediately in section 2.
+
+```python
+from optimum.neuron import NeuronModelForCausalLM
+
+
+compiler_args = {"num_cores": 24, "auto_cast_type": 'bf16'}
```
dacorvo, line 52: This configuration is not cached by default. Here is the list of cached configs:
https://huggingface.co/aws-neuron/optimum-neuron-cache/blob/main/inference-cache-config/qwen2.5.json

PR #755, opened 2025-01-06
Decoder: fix performance regression

dacorvo:
# What does this PR do?

With `transformers >= 4.46.0` the top-k, top-p and temperature logits processors have been included in the common logits processors lists, where they were previously included in a logits warpers list used only when sampling.

This defeated the optimization previously introduced in `optimum-neuron` to replace the logits warpers by a single fused warper. This led to a huge increase of the sampling latency.

This pull-request removes the superfluous `transformers` sampling warpers and restores the optimized processing.


JingyaHuang (approved):
LGTM!

PR #759, opened 2025-01-10
Trainium spelling nits

jimburtoft:
# What does this PR do?
This PR fulfills my late night need to not go to bed, while also assuaging some minor OCD tendencies.   

I came across a misspelling while working on an error with Swin models and felt the need to fix it everywhere, because how can I rest when something somewhere on the Internet is WRONG!

Other than a change to a keyword in setup.py (which _probably_ won't affect anything?), everything else is in comments or error messages.

Fixes # (issue)
If only this could fix some of my issues...

## Before submitting
- [ x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

### DIFF
--- a/optimum/neuron/trainers.py
+++ b/optimum/neuron/trainers.py
@@ -1510,13 +1510,13 @@ class _TrainerForNeuron:
 
 class NeuronTrainer(_TrainerForNeuron, Trainer):
     """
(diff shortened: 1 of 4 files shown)


dacorvo (approved):

dacorvo (approved):
Thanks for the pull-request 😉 !

PR #764, opened 2025-01-18
feat: Add support for phi4

jlonge4:
This PR adds support for Meta's Phi-4 model by adapting the existing LLaMA implementation. 

The Phi-4 architecture follows the LLaMA architecture closely, with the main difference being in how the weights are stored (fused `qkv_proj` and `gate_up` vs separate projections).

### DIFF
--- a/benchmark/text-generation-inference/performance/phi4-14b/docker-compose.yaml
+++ b/benchmark/text-generation-inference/performance/phi4-14b/docker-compose.yaml
@@ -1,0 +1,76 @@
+version: '3.7'
+
+services:
+  tgi-1:
+    image: neuronx-tgi:latest
+    ports:
+      - "8081:8081"
+    environment:
+      - PORT=8081
+      - MODEL_ID=${MODEL_ID}
+      - HF_AUTO_CAST_TYPE=${HF_AUTO_CAST_TYPE}
+      - HF_NUM_CORES=8
+      - MAX_BATCH_SIZE=${MAX_BATCH_SIZE}
+      - MAX_INPUT_TOKENS=${MAX_INPUT_TOKENS}
+      - MAX_TOTAL_TOKENS=${MAX_TOTAL_TOKENS}
+      - MAX_CONCURRENT_REQUESTS=512
+      - HF_TOKEN=${HF_TOKEN}
+    devices:
+      - "/dev/neuron0"
+      - "/dev/neuron1"
+      - "/dev/neuron2"
+      - "/dev/neuron3"
+
+  tgi-2:
+    image: neuronx-tgi:latest
+    ports:
+      - "8082:8082"
+    environment:
+      - PORT=8082
+      - MODEL_ID=${MODEL_ID}
+      - HF_AUTO_CAST_TYPE=${HF_AUTO_CAST_TYPE}
+      - HF_NUM_CORES=8
+      - MAX_BATCH_SIZE=${MAX_BATCH_SIZE}
+      - MAX_INPUT_TOKENS=${MAX_INPUT_TOKENS}
+      - MAX_TOTAL_TOKENS=${MAX_TOTAL_TOKENS}
+      - MAX_CONCURRENT_REQUESTS=512
+      - HF_TOKEN=${HF_TOKEN}
+    devices:
+      - "/dev/neuron4"
+      - "/dev/neuron5"
+      - "/dev/neuron6"
+      - "/dev/neuron7"
+
+  tgi-3:
+    image: neuronx-tgi:latest
+    ports:
+      - "8083:8083"
+    environment:
+      - PORT=8083
+      - MODEL_ID=${MODEL_ID}
+      - HF_AUTO_CAST_TYPE=${HF_AUTO_CAST_TYPE}
+      - HF_NUM_CORES=8
+      - MAX_BATCH_SIZE=${MAX_BATCH_SIZE}
+      - MAX_INPUT_TOKENS=${MAX_INPUT_TOKENS}
+      - MAX_TOTAL_TOKENS=${MAX_TOTAL_TOKENS}
+      - MAX_CONCURRENT_REQUESTS=512
+      - HF_TOKEN=${HF_TOKEN}
+    devices:
+      - "/dev/neuron8"
+      - "/dev/neuron9"
+      - "/dev/neuron10"
+      - "/dev/neuron11"
+
+  loadbalancer:
+    image: nginx:alpine
+    ports:
+      - "8080:80"
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf:ro
+    depends_on:
+      - tgi-1
+      - tgi-2
+      - tgi-3
+    deploy:
+      placement:
+        constraints: [node.role == manager]
--- a/benchmark/text-generation-inference/performance/phi4-14b/nginx.conf
+++ b/benchmark/text-generation-inference/performance/phi4-14b/nginx.conf
@@ -1,0 +1,15 @@
+### Nginx TGI Load Balancer
+events {}
+http {
+    upstream tgicluster {
+        server tgi-1:8081;
+        server tgi-2:8082;
+        server tgi-3:8083;
+    }
+    server {
+        listen 80;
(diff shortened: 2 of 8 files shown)


dacorvo:
@jlonge4 thank you very much for this pull-request: adding support for phi4 would be awesome.

We are however heavily refactoring the export mechanism to remove the dependency to `transformers-neuronx` and simplify the contribution of new models.

Can you take a look at that [pull-request](https://github.com/huggingface/optimum-neuron/pull/775) and see if it would make it easier for you to add support for phi4 based on the new HLO backend ?

jlonge4:
Hi there @dacorvo , just took a look at the difference and it certainly seems a lot slimmer! I think my effort would be the same in regard to the most important part for this which is the `load_weights` function. However it would obviously get rid of a lot of boiler plate. I am down to rework this PR and merge into add_hlo branch if you prefer.

dacorvo:
@jlonge4 the pull-request has been merged. Please let me know if you need any help rebasing your branch.

jlonge4:
@dacorvo Hopefully the last commit is pretty close

dacorvo:
@jlonge4 I rebased and squashed your branch into a new [phi4 branch](https://github.com/huggingface/optimum-neuron/tree/phi4), then did a few tests. The `load_weights` method required some important changes, but eventually I managed to export and run `microsoft/phi4`.

I think the most efficient is that you reset your branch using mine locally (assuming here you have an "upstream" repo pointing to the main `optimum-neuron` repository:

```
git fetch upstream
git reset --hard upstream/phi4
```

Then you can resume the work on `phi4` to add more tests (I only kept the export unit tests since TGI has now moved to the upstream repository).

jlonge4:
Just for my knowledge, what all specifically had to be done further to the load weights func? Any specific tests you'd like added from here?

dacorvo:
> Just for my knowledge, what all specifically had to be done further to the load weights func? Any specific tests you'd like added from here?

Main changes:
- unfused weights had to be transposed (like other weights),
- q,k,v splits are not even because of MQA (q gets a bigger chunk),
- many many other changes related to my refactoring (config parameters and attributes removed)

dacorvo:
@jlonge4 I think it would be good to add a phi3 config to the generation tests here: https://github.com/huggingface/optimum-neuron/blob/c1cf0f0ef22ea0fb5aeb9c6c178ccf27c4a95278/tests/decoder/conftest.py#L26

jlonge4:
@dacorvo got it! Btw so that we can effectively handle both, do you think we should add a check for 'attn_heads == kv_heads' since phi3/3.5 mini models don't use GQA?

dacorvo:
@jlonge4 I realize you already added a `phi4` test configuration with `microsoft/phi-4`. This model is too big for our unit tests IMHO so I tried to use `microsoft/Phi-3-mini-4k-instruct' instead but the results are garbage.

As a sanity test I checked the results of `microsoft/phi-4` on `gsm8k` on CUDA an neuron, and there is a significant drop in accuracy.

Here is the CUDA result:

image

And the neuron one:

image

jlonge4:
@dacorvo hmm, so strange. I was able to make `phi3.5-mini-instruct` and `phi-4` work with my NXDI integration, and the only difference was the `load_state_dict` [here](https://github.com/jlonge4/neuronx-distributed-inference/blob/e6dd6efff9fa88040ebe709208e72a9734a97074/src/neuronx_distributed_inference/models/phi3/modeling_phi3.py#L125). 

NXDI logit matching results: 

```Expected Output:  [", that is the question:\nWhether 'tis nobler in the mind to suffer\nThe slings and arrows of outrageous fortune,\nOr to take arms against a sea of troubles\nAnd by opposing end them. To die, to sleep;\nNo"] tensor([[29892,   393,   338,   278,  1139, 29901,    13,  8809,  1979,   525,
         28898, 22182,  1358,   297,   278,  3458,   304,  8812,    13,  1576,
          2243,   886,   322,   564,  5727,   310,   714,  6617,   681, 19717,
         29892,    13,  2816,   304,  2125, 10188,  2750,   263,  7205,   310,
         18835,    13,  2855,   491,  9209,   292,  1095,   963, 29889,  1763,
           762, 29892,   304,  8709, 29936,    13,  3782]])
Expected Logits Shape:  torch.Size([57, 1, 32064])
2025-Jan-10 01:04:05.0778 8296:9764 [0] nccl_net_ofi_rdma_init:7734 CCOM WARN NET/OFI OFI fi_getinfo() call failed: No data available
2025-Jan-10 01:04:05.0829 8296:9764 [0] nccl_net_ofi_create_plugin:251 CCOM WARN NET/OFI Unable to find a protocol that worked.  Failing initialization.
2025-Jan-10 01:04:05.0880 8296:9764 [0] nccl_net_ofi_create_plugin:316 CCOM WARN NET/OFI aws-ofi-nccl initialization failed
2025-Jan-10 01:04:05.0922 8296:9764 [0] nccl_net_ofi_init:139 CCOM WARN NET/OFI Initializing plugin failed
2025-Jan-10 01:04:05.0959 8296:9764 [0] net_plugin.cc:94 CCOM WARN OFI plugin initNet() failed is EFA enabled?
Actual Output:  [", that is the question:\nWhether 'tis nobler in the mind to suffer\nThe slings and arrows of outrageous fortune,\nOr to take arms against a sea of troubles\nAnd by opposing end them. To die, to sleep;\nNo"] tensor([[29892,   393,   338,   278,  1139, 29901,    13,  8809,  1979,   525,
         28898, 22182,  1358,   297,   278,  3458,   304,  8812,    13,  1576,
          2243,   886,   322,   564,  5727,   310,   714,  6617,   681, 19717,
         29892,    13,  2816,   304,  2125, 10188,  2750,   263,  7205,   310,
         18835,    13,  2855,   491,  9209,   292,  1095,   963, 29889,  1763,
           762, 29892,   304,  8709, 29936,    13,  3782]])
Actual Logits Shape:  torch.Size([57, 1, 32064])
Passed logits validation
```

dacorvo:
@jlonge4 I did a lot of tests comapring with GPU versions and:
- the outputs of `microsoft/phi4` are almost identical when doing greedy (at least the 256 first tokens), so the degradation of the `gsm8k` results might be due to some changes in the way I tested (it was through TGI on neuron, and HF directly on GPU, so maybe I am missing a system prompt somewhere). I nevertheless have suspicions that the rope code is not entirely identical, which would explain why both results on gpu and neuron deviate eventually,
- the outputs of `microsoft/Phi-3-mini-4k-instruct` are not complete garbage: they just don't make sense. This is probably related to the fact that it uses a standard MQA (that's the only difference), which means I might have a bug in the modeling code for that configuration.

dacorvo:
Hey @jlonge4 thank you for trying to fix the issue I raised, but:
- your change will give the exact same result as the current code,
- I realized that the issue is not related to MQA (see below).

So after countless attempts at finding a difference between the two models, I realized that I had always tested `phi4` with a batch size of 1 and `phi3-mini` with a batch size of 4.

I did a quick test with inverted batch sizes, and the issue is indeed there: with a batch_size of 1 both models are correct, but for some reason when the batch_size is higher it is as if the context was completely ignored, as it generates some random text that looks like training material.

I quickly checked for a more general regression and it looks like I don't have that issue with llama. I will do more tests to try to figure out when this issue appears. Thank you for your patience.

jlonge4:
@dacorvo No worries at all. Thanks so much for all your work to get this PR the rest of the way!

dacorvo:
@jlonge4 I finally identified the root cause for the errors, and it is not related to the `phi3` architecture. The private `neuronxcc` flash attention kernel does not seem to provide the correct results in the specific input configuration that is triggered for `phi` models.
This happens here: https://github.com/huggingface/optimum-neuron/blob/9fbc3ae928afe01f63d869a54f86a9216cf181a1/optimum/neuron/backends/hlo/layers/attention.py#L664.

I tried different combination of parameters to identify a pattern without success so far.

jlonge4:
@dacorvo well that's a nice sanity check at least! I suppose nothing we can do from there. Thanks so much for all of your help / education on this!

PR #771, opened 2025-01-29
Bump transformers version to 4.48.1

dacorvo:
# What does this PR do?

This bumps the transformers version to 4.48.1 to allow the deployment of images containing `optimum-neuron` in AWS services (all previous transformers versions are now rejected because they include some scripts identified as potentially harmful).

This also modifies the documentation to change the installation instructions: installing as an optimum extra is no longer recommended and will be soon deprecated.

There were some changes in transformers modeling for LLama and Mistral models attention forward that needed to be reflected in the distributed parallelization modeling code.

There were also some changes in the attributes of the transformers Attention classes for the same models that broke the parallelization dynamic transformations: the changes in this pull-request are just a workaround, and eventually these dynamic transformations should be replaced by explicit modeling code (the forward methods are already overloaded, so there is no point in patching dynamically the layers where it would be simpler to recreate them directly from the config).

### DIFF
--- a/optimum/neuron/distributed/decoder_models.py
+++ b/optimum/neuron/distributed/decoder_models.py
@@ -834,6 +834,10 @@ class MistralParallelizer(Parallelizer):
                 **parallel_layer_specific_kwargs,
             )
         for layer in model.model.layers:
+            # FIXME: temporary workaround to avoid too many changes in the transformation code
+            layer.self_attn.num_heads = layer.self_attn.config.num_attention_heads
+            layer.self_attn.num_key_value_heads = layer.self_attn.config.num_key_value_heads
+            layer.self_attn.hidden_size = layer.self_attn.config.hidden_size
             layer.self_attn = MistralParallelSelfAttention.transform(
                 model,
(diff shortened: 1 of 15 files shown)


tengomucho (approved):

JingyaHuang (approved):
lgtm, thanks.

optimum/neuron/distributed/decoder_models.py
```diff
@@ -588,6 +587,10 @@ def _parallelize(
             layers = model.model.layers
 
         for layer in layers:
+            # FIXME: temporary workaround to avoid too many changes in the transformation code
+            layer.self_attn.num_heads = layer.self_attn.config.num_attention_heads
+            layer.self_attn.num_key_value_heads = layer.self_attn.config.num_key_value_heads
+            layer.self_attn.hidden_size = layer.self_attn.config.hidden_size
```
michaelbenayoun, line 593: Thanks!

michaelbenayoun (approved):
LGTM!

PR #773, opened 2025-01-29
chore(ci): remove black from CI, we already use ruff

tengomucho:
# What does this PR do?

CI was failing due to black errors, but we are already using ruff to check style and lint, so removing black seems like a fair choice.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,11 +15,6 @@
 requires = ["setuptools==69.5.1"]
 build-backend = "setuptools.build_meta"
 
-[tool.black]
-line-length = 119
-target-version = ['py38']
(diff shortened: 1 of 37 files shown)


pyproject.toml
```diff
@@ -15,11 +15,6 @@
 requires = ["setuptools==69.5.1"]
 build-backend = "setuptools.build_meta"
 
-[tool.black]
-line-length = 119
-target-version = ['py38']
-extend-exclude = '.ipynb'
-
```
dacorvo, line 22: We need to find a way to keep the exclusion of the notebooks, as it is too painful.

dacorvo (approved):

PR #776, opened 2025-02-06
display available cached versions in TGI server error message

jimburtoft:
If a model is cached with a different configuration, I want to display alternative options to the user.  

If someone copies from the deploy code on Hugging Face and changes something (e.g. sequence length), it is not obvious why it isn't working from this code. (especially if they don't understand compiling because they are referencing the original model)

Based on a true story!

added some carriage returns to make it more readable 

get_hub_cached_entries does generate an error if it is fed a model that doesn't have a model_type.  For example:  (randomly selected) model_id = "hexgrad/Kokoro-82M"

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/aws_neuronx_venv_pytorch_2_1/lib/python3.10/site-packages/optimum/neuron/utils/hub_cache_utils.py", line 431, in get_hub_cached_entries
    model_type = target_entry.config["model_type"]
KeyError: 'model_type'

However, we already call that function inside of is_cached at the top of this block, so I don't know if we are filtering for certain types of models before we get to this point or not.  If not, the existing code would generate that error before it ever gets here.

## Before submitting

### DIFF
--- a/text-generation-inference/server/text_generation_server/model.py
+++ b/text-generation-inference/server/text_generation_server/model.py
@@ -99,10 +99,28 @@ def fetch_model(
     if not is_cached(model_id, neuron_config):
         hub_cache_url = "https://huggingface.co/aws-neuron/optimum-neuron-cache"
         neuron_export_url = "https://huggingface.co/docs/optimum-neuron/main/en/guides/export_model#exporting-neuron-models-using-neuronx-tgi"
+            
+        entries = get_hub_cached_entries(model_id, "inference")
+        available_configs = ""
+        if entries:
+            config_list = []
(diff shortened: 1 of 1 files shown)


jimburtoft:
@dacorvo Not urgent.  Do carriage returns break logging in any way?

dacorvo:
@jimburtoft thnak you for this pull-request. The neuron tgi code has now been upstreamed to the main TGI repository: could you resubmit your pull-request there ? Sorry for the inconvenience ...

jimburtoft:
Moved to TGI repository as https://github.com/huggingface/text-generation-inference/pull/3063

PR #777, opened 2025-02-06
chore(ami): set env variable to enalbe hf transfer

tengomucho:
This triggers faster download/upload speeds when dealing with the hub on the AMI image.


JingyaHuang (approved):
LGTM, thx

dacorvo (approved):

PR #786, opened 2025-02-17
Bump diffusers dependency

dacorvo:
# What does this PR do?

This bumps the minimum `diffusers` version, since the `0.28.x` versions leads to an import error with recent `huggingface_hub` versions.


setup.py
```diff
@@ -29,7 +29,7 @@
     "sentencepiece",
     "datasets",
     "sacremoses",
-    "diffusers>=0.28.0, <=0.30.3",
+    "diffusers>=0.29.0, <=0.30.3",
```
JingyaHuang, line 32: where does it comes from? optimum?

dacorvo: I added it in the commit [message](https://github.com/huggingface/optimum-neuron/pull/786/commits/baa33d3e26f77512b31f0b2518a44170995624dd): it comes from huggingface_hub.

michaelbenayoun (approved):

PR #789, opened 2025-02-19
Add Whisper for the task "automatic-speech-recognition" w/o. KV cache

JingyaHuang:
# What does this PR do?

### Export

* With CLI

```bash
optimum-cli export neuron --model openai/whisper-tiny --task automatic-speech-recognition --batch_size 1 --sequence_length 128 --auto_cast matmul --auto_cast_type bf16 whisper_tiny_neuronx_bf16/
```

* With API

```python
from optimum.neuron import NeuronWhisperForConditionalGeneration
save_directory = "whisper_tiny_neuronx/"

# 1. Export
compiler_args = {"auto_cast": "matmul", "auto_cast_type": "bf16"}
input_shapes = {"batch_size": 1, "sequence_length": 128}
neuron_model = NeuronWhisperForConditionalGeneration.from_pretrained(
    model_id,
    export=True,
    **compiler_args,
    **input_shapes,
)
# Save locally or upload to the HuggingFace Hub
neuron_model.save_pretrained(save_directory)
```

### Inference

```python
neuron_model = NeuronWhisperForConditionalGeneration.from_pretrained(save_directory)
predicted_ids = neuron_model.generate(input_features)
transcription = processor.batch_decode(predicted_ids, skip_special_tokens=True)
print(transcription[0])
```

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you make sure to update the documentation with your changes?


optimum/exporters/neuron/model_wrappers.py
```diff
@@ -31,10 +34,11 @@
 
 
 class UnetNeuronWrapper(torch.nn.Module):
-    def __init__(self, model, input_names: List[str]):
+    def __init__(self, model, input_names: List[str], device: str = None):
```
tengomucho, line 37: why do you need to specify the device here? It is not really used in most cases isn't it? It seems to me it's a way to make these wrappers' init functions coherent, but maybe you do not need to set it as member if you do not plan to use it.

JingyaHuang: It's truly used for cases that we create tensors inside the wrapper (xla for the tracing, whereas we reuse the wrapper on CPU during the validation). I think it's nice to have for the consistency.

optimum/exporters/neuron/base.py
```diff
@@ -379,6 +386,7 @@ def patch_model_for_export(
         dummy_inputs: Optional[Dict[str, torch.Tensor]] = None,
         forward_with_tuple: bool = False,
         eligible_outputs: Optional[List[Union[str, int]]] = None,
+        device: str = None,
```
michaelbenayoun, line 389:
```suggestion
        device: Optional[str] = None,
```

optimum/exporters/neuron/convert.py
```diff
@@ -533,7 +532,7 @@ def export_neuronx(
     # Prepare the model / function(tp) to trace
     aliases = {}
     tensor_parallel_size = config.tensor_parallel_size
-    if isinstance(config, TextSeq2SeqNeuronConfig):
+    if hasattr(config, "is_encoder_decoder") and config.is_encoder_decoder:
```
michaelbenayoun, line 535:
```suggestion
    if getattr(config, "is_encoder_decoder", False):
```

optimum/exporters/neuron/model_wrappers.py
michaelbenayoun, line 37:
```suggestion
    def __init__(self, model, input_names: List[str], device: Optional[str] = None):
```

optimum/neuron/modeling_seq2seq.py
```diff
@@ -89,9 +91,100 @@
...
+    """
+
+    def __init__(
+        self,
+        model: torch.jit._script.ScriptModule,
+        parent_model: NeuronTracedModel,
+        config: Optional["PretrainedConfig"] = None,
+        neuron_config: Optional["NeuronDefaultConfig"] = None,
+        model_type: str = "encoder",
+        device: Optional[int] = None,
```
michaelbenayoun, line 106: Is it really an `int`?

optimum/neuron/modeling_seq2seq.py
```diff
@@ -653,19 +744,32 @@ def __init__(
         neuron_config: Optional[Dict[str, str]] = None,
     ):
         super().__init__(model, parent_model, config, neuron_config, "encoder")
+        stride = getattr(self.config, "stride", [1, 2])
+        self.conv1 = DummyLayer(stride=[stride[0]])
+        self.conv2 = DummyLayer(stride=[stride[1]])
```
michaelbenayoun, line 749: What's the use of these dummy layers?

setup.py
```diff
@@ -13,7 +13,7 @@
 
 
 INSTALL_REQUIRES = [
-    "transformers ~= 4.48.1",
+    "transformers ~= 4.49.0",
```
michaelbenayoun, line 16: Are we sure we want to do that in this PR?

optimum/exporters/neuron/model_wrappers.py
tengomucho: yes, but why do you set the value in the object if you do not use it?

setup.py
JingyaHuang: There was a bug in the `WhisperGenerationMixin.detect_language` leading the compilation issue, the PR won't be valid with transformers lower than 4.49.0 unfortunately. Inference-wise, all CIs are green, would it be a problem for the training side?

optimum/neuron/modeling_seq2seq.py
JingyaHuang: Good catch, will fix!

setup.py
JingyaHuang: https://github.com/huggingface/transformers/blob/858545047c05a35fde437b2ada3a901844cd1e60/src/transformers/models/whisper/generation_whisper.py#L1574

(use_cache need to be set to False here)

optimum/exporters/neuron/model_wrappers.py
JingyaHuang: because custom wrappers are called here:
https://github.com/huggingface/optimum-neuron/blob/1b138c37e99fac4e73e77fe3cbf422e758886cb1/optimum/exporters/neuron/convert.py#L203
 so I preferred adding the unused(for now) device to **kwargs to every wrapper

dacorvo (changes requested):
LGTM, but two small changes required 🙏

.github/actions/prepare_venv/action.yml
```diff
@@ -7,6 +7,7 @@ runs:
         shell: bash
         run: |
           sudo apt install python3-venv python3-dev -y
+          sudo apt-get install python3-opencv -y
```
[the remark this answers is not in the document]
JingyaHuang: Actually, all inference CIs are impacted by this dependency as far as something in the `optimum/neuron/utils/misc.py` is imported since we imported `from transformers.modeling_utils import _add_variant` there.

```
optimum/neuron/generation/utils.py:29: in <module>
    from ..utils.misc import args_and_kwargs_to_kwargs_only
optimum/neuron/utils/misc.py:27: in <module>
    from transformers.modeling_utils import _add_variant
aws_neuron_venv_pytorch/lib/python3.10/site-packages/transformers/modeling_utils.py:53: in <module>
    from .loss.loss_utils import LOSS_MAPPING
aws_neuron_venv_pytorch/lib/python3.10/site-packages/transformers/loss/loss_utils.py:19: in <module>
    from .loss_deformable_detr import DeformableDetrForObjectDetectionLoss, DeformableDetrForSegmentationLoss
aws_neuron_venv_pytorch/lib/python3.10/site-packages/transformers/loss/loss_deformable_detr.py:4: in <module>
    from ..image_transforms import center_to_corners_format
aws_neuron_venv_pytorch/lib/python3.10/site-packages/transformers/image_transforms.py:[22](https://github.com/huggingface/optimum-neuron/actions/runs/13836914041/job/38714200984#step:6:23): in <module>
    from .image_utils import (
aws_neuron_venv_pytorch/lib/python3.10/site-packages/transformers/image_utils.py:84: in <module>
    import cv2
aws_neuron_venv_pytorch/lib/python3.10/site-packages/cv2/__init__.py:181: in <module>
    bootstrap()
aws_neuron_venv_pytorch/lib/python3.10/site-packages/cv2/__init__.py:153: in bootstrap
    native_module = importlib.import_module("cv2")
E   ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```

https://github.com/huggingface/optimum-neuron/actions/runs/13836914041/job/38714200984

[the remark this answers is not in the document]
JingyaHuang: It's brought by transformers 4.49.0, will check if there is a better way to tackle this.

.github/workflows/test_inf2_llm.yml
```diff
@@ -34,6 +34,9 @@ jobs:
         uses: ./.github/actions/prepare_venv
       - name: Install optimum-neuron
         uses: ./.github/actions/install_optimum_neuron
+      - name: Install cv2 dependencies
+        run: |
+          sudo apt-get install python3-opencv -y
```
dacorvo, line 39: I would like to avoid that if possible, as I will need to add this to the TGI images also, which is silly because it is never used.

JingyaHuang: this CI will fail as well without this dependency, it comes from these lines in transformers:
https://github.com/huggingface/transformers/blob/a22a4378d97d06b7a1d9abad6e0086d30fdea199/src/transformers/image_utils.py#L83-L84
which has been removed tho in main... but I don't want to wait for 4.49.1 release to merge this PR

JingyaHuang: We can bump to transformers 4.49.1 before you build TGI image next time perhaps

eercanayar:
As explicitly stated in the issue title, this has no KV cache support. I'm curious how does it impact the model performance compared to the KV-cached one? Is it in roadmap, or is there any blockers for that? Thanks in advance for your insights.

PR #799, opened 2025-02-27
ci: populate caches when tagging

dacorvo:
# What does this PR do?

The inference caches will now be updated when pushing a new release tag. This removes the need for running the workflow manually later on (although it is still possible).

### DIFF
--- a/.github/workflows/inference_cache_llm.yml
+++ b/.github/workflows/inference_cache_llm.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     # Schedule the workflow to run every day at midnight UTC
     - cron: '0 0 * * *'
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


tengomucho (approved):
LGTM

tengomucho:
is it still worth keeping the cron if you are populating when tagging?

dacorvo:
> is it still worth keeping the cron if you are populating when tagging?

Yes because it does the caching on main. Caching actually means two things:
- compile and cache the NEFFs files: these only depend on the compiler and graph,
- create a registry entry for the source model and configuration, so that we can lookup if it is cached: these are version/tag specific.

So basically, when tagging, only the registry entries will be created, because usually the NEFFs files will have been cached by the nightly.

PR #801, opened 2025-03-03
Custom modeling for training

michaelbenayoun:
# What does this PR do?

### DIFF
--- a/optimum/neuron/models/training/llama/modeling_llama.py
+++ b/optimum/neuron/models/training/llama/modeling_llama.py
@@ -1,0 +1,605 @@
+# coding=utf-8
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""LlaMa model implementation for Neuron"""
+
+import math
+import os
+from functools import partial
+from typing import List, Optional, Tuple, Union, Callable
+
+import torch
+import torch.nn.functional as F
+import torch.utils.checkpoint
+import torch_xla.core.xla_model as xm
+from torch_xla.utils.checkpoint import checkpoint
+from torch import nn
+
+from transformers.activations import ACT2FN
+from transformers.modeling_outputs import (
+    BaseModelOutputWithPast,
+    CausalLMOutputWithPast,
+)
+from transformers.models.llama.configuration_llama import LlamaConfig, KwargsForCausalLM
+from transformers.models.llama.modeling_llama import (
+    LLAMA_INPUTS_DOCSTRING,
+    LLAMA_START_DOCSTRING,
+)
+from transformers.models.llama.modeling_llama import LlamaAttention as LlamaAttentionHF
+from transformers.models.llama.modeling_llama import (
+    LlamaDecoderLayer as LlamaDecoderLayerHF,
+)
+from transformers.models.llama.modeling_llama import (
+    LlamaForCausalLM as LlamaForCausalLMHF,
+)
+from transformers.models.llama.modeling_llama import (
+    LlamaForSequenceClassification,
+    LlamaLinearScalingRotaryEmbedding,
+)
+from transformers.models.llama.modeling_llama import LlamaMLP as LlamaMLPHF
+from transformers.models.llama.modeling_llama import LlamaModel as LlamaModelHF
+from transformers.models.llama.modeling_llama import LlamaPreTrainedModel
+from transformers.models.llama.modeling_llama import LlamaRMSNorm as LlamaRMSNormHF
+from transformers.models.llama.modeling_llama import (
+    LlamaRotaryEmbedding,
+    apply_rotary_pos_emb,
+    repeat_kv,
+    rotate_half,
+)
+from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
+from transformers.processing_utils import Unpack
+from transformers.cache_utils import Cache, DynamicCache, StaticCache
+from transformers.utils import (
+    add_start_docstrings,
+    add_start_docstrings_to_model_forward,
+    logging,
+    replace_return_docstrings,
+)
+
+import neuronx_distributed.parallel_layers.utils as neuronx_dist_utils
+from neuronx_distributed.kernels.flash_attn import nki_flash_attn_func
+from neuronx_distributed.modules.qkv_linear import GQAQKVColumnParallelLinear
+from neuronx_distributed.parallel_layers import mappings
+from neuronx_distributed.parallel_layers.layers import (
+    ColumnParallelLinear,
+    ParallelEmbedding,
+    RowParallelLinear,
+)
+from neuronx_distributed.parallel_layers.loss_functions import parallel_cross_entropy
+from neuronx_distributed.parallel_layers.parallel_state import get_tensor_model_parallel_size
+from neuronx_distributed.utils.model_utils import move_model_to_device
+
+from ...loss_utils import ForCausalLMLoss
+
+
+logger = logging.get_logger(__name__)
+
+def _init_normal(std, w):
+    return nn.init.normal_(w, mean=0.0, std=std)
+
+
+class LlamaRMSNorm(LlamaRMSNormHF):
+    def __init__(self, hidden_size, eps=1e-6, sequence_parallel_enabled=False):
+        """
+        LlamaRMSNorm is equivalent to T5LayerNorm
+        """
+        super().__init__(hidden_size, eps=eps)
+        setattr(self.weight, "sequence_parallel_enabled", sequence_parallel_enabled)
+
+
+class LlamaMLP(LlamaMLPHF):
+    def __init__(self, config):
+        nn.Module.__init__(self)
+        self.config = config
+        self.pretraining_tp = config.pretraining_tp
+        self.hidden_size = config.hidden_size
+        self.intermediate_size = config.intermediate_size
+        self.act_fn = ACT2FN[config.hidden_act]
+
+        init_method = partial(_init_normal, config.initializer_range)
+        self.gate_up_proj = ColumnParallelLinear(
+            self.hidden_size,
+            2 * self.intermediate_size,
+            stride=2,
+            bias=False,
+            gather_output=False,
+            init_method=init_method,
+            sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+            dtype=self.config.torch_dtype,
+        )
+        self.down_proj = RowParallelLinear(
+            self.intermediate_size,
+            self.hidden_size,
+            bias=False,
+            input_is_parallel=True,
+            init_method=init_method,
+            sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+            dtype=self.config.torch_dtype,
+        )
+        self.split_size = self.intermediate_size // get_tensor_model_parallel_size()
+        if config.move_model_to_device:
+            move_model_to_device(self, xm.xla_device())
+
+    def forward(self, x):
+        gate_proj, up_proj = self.gate_up_proj(x).split(self.split_size, dim=2)
+
+        def activation_mlp(gate_proj, up_proj):
+            activation_output = self.act_fn(gate_proj)
+            return activation_output * up_proj
+
+        # We checkpoint the MLP compute too, since we see extra data movement which is more
+        # expensive than the recompute in this case.
+        if self.config.selective_checkpoint_enabled:
+            intermediate_states = checkpoint(activation_mlp, gate_proj, up_proj)
+        else:
+            intermediate_states = self.act_fn(gate_proj) * up_proj
+        down_proj = self.down_proj(intermediate_states)
+
+        return down_proj
+
+def eager_attention_forward(
+    module: nn.Module,
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    attention_mask: Optional[torch.Tensor],
+    scaling: float,
+    dropout: float = 0.0,
+    **kwargs,
+):
+    key_states = repeat_kv(key, module.num_key_value_groups)
+    value_states = repeat_kv(value, module.num_key_value_groups)
+
+    attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
+    if attention_mask is not None:
+        causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
+        attn_weights = attn_weights + causal_mask
+
+    attn_weights = nn.functional.softmax(attn_weights, dim=-1, dtype=torch.float32).to(query.dtype)
+    attn_weights = nn.functional.dropout(attn_weights, p=dropout, training=module.training)
+    attn_output = torch.matmul(attn_weights, value_states)
+    attn_output = attn_output.transpose(1, 2).contiguous()
+
+    return attn_output, attn_weights
+
+class LlamaAttention(LlamaAttentionHF):
+    """Multi-headed attention from 'Attention Is All You Need' paper"""
+
+    def __init__(self, config: LlamaConfig, layer_idx: int):
+        nn.Module.__init__(self)
+        self.config = config
+        self.layer_idx = layer_idx
+        self.hidden_size = config.hidden_size
+        self.num_heads = config.num_attention_heads
+        self.head_dim = getattr(config, "head_dim", self.hidden_size // self.num_heads)
+        self.num_key_value_heads = config.num_key_value_heads
+        self.num_key_value_groups = self.num_heads // self.num_key_value_heads
+        self.scaling = self.head_dim ** -0.5
+        self.attention_dropout = config.attention_dropout
+        self.is_causal = True
+
+        if not hasattr(config, "kv_shared_group_size"):
+            config.kv_shared_group_size = 1
+
+        if not hasattr(config, "qkv_linear"):
+            config.qkv_linear = False
+
+        if not hasattr(config, "fuse_qkv"):
+            config.fuse_qkv = False
+
+        if not hasattr(config, "use_flash_attention"):
+            self.use_flash_attention = False
+        else:
+            self.use_flash_attention = config.use_flash_attention
+
+        if (self.head_dim * self.num_heads) != self.hidden_size:
+            raise ValueError(
+                f"hidden_size must be divisible by num_heads (got `hidden_size`: {self.hidden_size}"
+                f" and `num_heads`: {self.num_heads})."
+            )
+        self._init_rope()
+
+        init_method = partial(_init_normal, config.initializer_range)
+        if self.config.qkv_linear:
+            self.qkv_proj = GQAQKVColumnParallelLinear(
+                self.hidden_size,
+                [self.num_heads * self.head_dim, self.num_key_value_heads * self.head_dim],
+                bias=False,
+                gather_output=False,
+                init_method=init_method,
+                sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+                kv_size_multiplier=self.config.kv_shared_group_size,
+                fuse_qkv=self.config.fuse_qkv,
+                dtype=self.config.torch_dtype,
+            )
+        elif self.config.fuse_qkv and self.num_heads == self.num_key_value_heads:
+            self.qkv_proj = ColumnParallelLinear(
+                self.hidden_size,
+                3 * self.num_heads * self.head_dim,
+                stride=3,
+                bias=False,
+                gather_output=False,
+                init_method=init_method,
+                sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+                dtype=self.config.torch_dtype,
+            )
+            self.split_size = self.num_heads * self.head_dim // get_tensor_model_parallel_size()
+        else:
+            self.q_proj = ColumnParallelLinear(
+                self.hidden_size,
+                self.num_heads * self.head_dim,
+                bias=False,
+                gather_output=False,
+                init_method=init_method,
+                sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+                dtype=self.config.torch_dtype,
+            )
+            self.k_proj = ColumnParallelLinear(
+                self.hidden_size,
+                self.num_key_value_heads * self.head_dim,
+                bias=False,
+                gather_output=False,
+                init_method=init_method,
+                sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+                dtype=self.config.torch_dtype,
+            )
+            self.v_proj = ColumnParallelLinear(
+                self.hidden_size,
+                self.num_key_value_heads * self.head_dim,
+                bias=False,
+                gather_output=False,
+                init_method=init_method,
+                sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+                dtype=self.config.torch_dtype,
+            )
+        self.o_proj = RowParallelLinear(
+            self.num_heads * self.head_dim,
+            self.hidden_size,
+            bias=False,
+            input_is_parallel=True,
+            init_method=init_method,
+            sequence_parallel_enabled=self.config.sequence_parallel_enabled,
+            dtype=self.config.torch_dtype,
+        )
+        self.num_heads = neuronx_dist_utils.divide(config.num_attention_heads, get_tensor_model_parallel_size())
+        self.num_key_value_heads = neuronx_dist_utils.divide(
+            config.num_key_value_heads * self.config.kv_shared_group_size, get_tensor_model_parallel_size()
+        )
+        self.num_key_value_groups = self.num_heads // self.num_key_value_heads
+
+        if config.move_model_to_device:
+            move_model_to_device(self, xm.xla_device())
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        position_embeddings: Tuple[torch.Tensor, torch.Tensor],
+        attention_mask: Optional[torch.Tensor],
+        past_key_value: Optional[Cache] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[FlashAttentionKwargs],
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+        bsz, q_len, _ = hidden_states.size()
+        if self.config.sequence_parallel_enabled:
+            q_len, bsz, _ = hidden_states.size()
+            q_len = q_len * get_tensor_model_parallel_size()
+
+        if (
+            self.config.fuse_qkv
+            and self.num_heads == self.num_key_value_heads
+            and self.config.kv_shared_group_size == 1
+        ):
+            qkv_states = self.qkv_proj(hidden_states)
+            query_states, key_states, value_states = qkv_states.split(self.split_size, dim=2)
+        elif self.config.qkv_linear:
+            query_states, key_states, value_states = self.qkv_proj(hidden_states)
+        else:
+            query_states = self.q_proj(hidden_states)
+            key_states = self.k_proj(hidden_states)
+            value_states = self.v_proj(hidden_states)
+
+        if self.config.sequence_parallel_enabled:
+            query_states = query_states.view(q_len, bsz, self.num_heads, self.head_dim).permute(1, 2, 0, 3)
+            key_states = key_states.view(q_len, bsz, self.num_key_value_heads, self.head_dim).permute(1, 2, 0, 3)
+            value_states = value_states.view(q_len, bsz, self.num_key_value_heads, self.head_dim).permute(1, 2, 0, 3)
+        else:
+            query_states = query_states.view(bsz, q_len, self.num_heads, self.head_dim).transpose(1, 2)
+            key_states = key_states.view(bsz, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+            value_states = value_states.view(bsz, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+
+        cos, sin = position_embeddings
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        if past_key_value is not None:
+            # sin and cos are specific to RoPE models; cache_position needed for the static cache
+            cache_kwargs = {"sin": sin, "cos": cos, "cache_position": cache_position}
+            key_states, value_states = past_key_value.update(key_states, value_states, self.layer_idx, cache_kwargs)
+
+        attention_interface: Callable = eager_attention_forward
+        if self.config._attn_implementation != "eager":
+            if self.config._attn_implementation != "flash_attention_2":
+                raise RuntimeError(
+                    f"Unsupported attention implementation for Neuron: {self.config._attn_implementation}"
+                )
+            else:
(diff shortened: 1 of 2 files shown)


optimum/neuron/models/training/transformations_utils.py
michaelbenayoun, line 1: This file defines the API on how we will represent model weights transformations that happen between the original Transformers implementation and the custom implementation we have.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        else:
+            attn_output = attn_output.transpose(1, 2).contiguous()
+            attn_output = attn_output.reshape(bsz, q_len, self.hidden_size // get_tensor_model_parallel_size())
+
+        attn_output = self.o_proj(attn_output)
+
+        return attn_output, attn_weights
+
+
+class LlamaDecoderLayer(LlamaDecoderLayerHF):
```
dacorvo, line 404: Okay here you can inherit because you reuse the forward.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        self.mlp = LlamaMLP(config, mp_config)
+        self.input_layernorm = LlamaRMSNorm(
+            config.hidden_size, eps=config.rms_norm_eps, sequence_parallel_enabled=mp_config.sequence_parallel_enabled
+        )
+        self.post_attention_layernorm = LlamaRMSNorm(
+            config.hidden_size, eps=config.rms_norm_eps, sequence_parallel_enabled=mp_config.sequence_parallel_enabled
+        )
+
+
+class LlamaModel(NeuronModelMixin, LlamaModelHF):
```
dacorvo, line 421: Here also I did not spot anything that is reused from LLamaModelHF, so I don't think the inheritance is required.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        output = BaseModelOutputWithPast(
+            last_hidden_state=hidden_states,
+            past_key_values=past_key_values if use_cache else None,
+            hidden_states=all_hidden_states,
+            attentions=all_self_attns,
+        )
+        return output if return_dict else output.to_tuple()
+
+
+class LlamaForCausalLM(NeuronModelMixin, LlamaForCausalLMHF):
```
dacorvo, line 614: I think the inheritance from LlamaForCausalLM is not required, and misleading since I don't think `generate` actually works, and you've overloaded `from_pretrained` and `save_pretrained`.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        output_hidden_states: Optional[bool] = None,
+        return_dict: Optional[bool] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **flash_attn_kwargs: Unpack[FlashAttentionKwargs],
+    ) -> Union[Tuple, BaseModelOutputWithPast]:
+        output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
+        output_hidden_states = (
+            output_hidden_states if output_hidden_states is not None else self.config.output_hidden_states
+        )
+        use_cache = use_cache if use_cache is not None else self.config.use_cache
```
dacorvo, line 518: The fact that you mention the cache is really misleading: is it really supported ? If you forced use_cache to be False it would simplify a lot the code below.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,714 @@
...
+
+            if not isinstance(config._attn_implementation, dict) and config._attn_implementation not in [
+                "eager"
+            ] + list(ALL_ATTENTION_FUNCTIONS.keys()):
+                message = f'Specified `attn_implementation="{config._attn_implementation}"` is not supported. The only possible arguments are `attn_implementation="eager"` (manual attention implementation)'
+                if cls._supports_flash_attn_2:
+                    message += (
+                        ', `"attn_implementation=flash_attention_2"` (implementation using nki flash attention 2)'
+                    )
+                # Keeping this if supported one day.
```
dacorvo, line 181: Since the `transformers` code will evolve when we decide to support this, I think it is better to avoid keeping these commented code snippets.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,714 @@
...
+
+        if config._attn_implementation == "flash_attention_2":
+            cls._check_and_enable_flash_attn_2(
+                config,
+                torch_dtype=torch_dtype,
+                device_map=device_map,
+                hard_check_only=False,
+                check_device_map=check_device_map,
+            )
+        # Keeping it if supported one day.
```
dacorvo, line 220: Same here: we can drop this.

tests/test_trainers.py
```diff
@@ -59,9 +59,9 @@
 class TestNeuronTrainingUtils(DistributedTest):
     @pytest.fixture(
         scope="class",
-        # TODO: enable dp + tp + pp, currently produces communication error between replicas.
-        params=[[2, 1, 1], [2, 2, 1], [2, 1, 2]],  # , [32, 2, 2]],
-        ids=["dp=2", "tp=2", "pp=2"],  # , "dp=8,tp=pp=2"],
+        # Anything related to PP seems broken because of the way the model is traced.
```
dacorvo, line 62: Does it mean that this test now uses the new custom modeling ? Or that it never worked ? Or that it is broken by your new changes ?

dacorvo (changes requested):
Thank you for this pull-request: massive contribution.
The general organization of the files makes sense and I think the `modelization_utils.py` and `transformation_utils.py` files are in particular good starting points.
I have requested a few changes that I think are important in this first iteration to:
- make it clearer what is expected to be provided by someone adding a new model,
- keep things simple in this first iteration (I personally think it is easier to add things we support later on instead of keeping placeholders).

Also, I think the basic features must be tested, and it was unclear to me by reading the pull-requests which parts are actually tested, apart from eager forward inference (and by the way we must include a flash_attention test since we support the option).

tests/distributed/test_common.py
```diff
@@ -108,10 +110,11 @@ def move_params_to_cpu(parameters):
 @is_trainium_test
 class TestCommonDistributed(DistributedTest):
     # TODO: enable dp=4,tp=pp=2 when working on the multi-node training PR.
+    # Anything related to PP seems broken because of the way the model is traced.
```
tengomucho, line 113: If PP is broken, I think it is better to raise an error. Remove the commented combination, and/or isolate it in another test that checks that en error is raised when PP is used.

tengomucho (changes requested):
A lot of work contributed here, thanks! This will help adding new models in a more efficient way.
I added few comments, but It mostly resumes to these requests:

- [ ] Remove the dependency from transformers modeling code
- [ ] Reuse and merge with existing code, avoid duplication for sharding and tests.
- [ ] Consider adding a complete test that shows training works (overfitting?)

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        if self.config._attn_implementation == "flash_attention_2":
+            if attention_mask is not None and (attention_mask == 0.0).any():
+                raise RuntimeError(f"Only a causal mask is supported with {self.config._attn_implementation}.")
+                # return attention_mask
+            return None
+
+        # For SDPA, when possible, we will rely on its `is_causal` argument instead of its `attn_mask` argument, in
+        # order to dispatch on Flash Attention 2. This feature is not compatible with static cache, as SDPA will fail
+        # to infer the attention mask.
+        past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
```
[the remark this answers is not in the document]
tengomucho: +1 to this, I think you can remove cache support as I did on granite [here](https://github.com/huggingface/optimum-neuron/commit/21de696a2c30925cced669fd80e3ab099ca0192d)

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -0,0 +1,692 @@
...
+                "partition_dim": param.partition_dim,
+                "partition_stride": param.partition_stride,
+                "num_partitions": param.num_partitions,
+            }
+        else:
+            metadata["parameters"][name] = {
+                "tensor_model_parallel": tensor_model_parallel,
+            }
+    for name, module in model.named_modules():
+        model_weight_transformation_specs = getattr(module, "specs", None)
```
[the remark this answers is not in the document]
michaelbenayoun: I thought about this. I just did not do it because I wanted to keep things "light" (I realize they are not light already...). 
I will do it since I agree it's a bit better. But we cannot do as how you suggest because it is an attribute that inner modules have, not only the main model instance. 

I can create a class `CustomModule` with this property. I can also help us find where the code is custom compared to Transformers.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+    return nn.init.normal_(w, mean=0.0, std=std)
+
+
+class LlamaRMSNorm(LlamaRMSNormHF):
+    def __init__(self, hidden_size, eps=1e-6, sequence_parallel_enabled=False):
+        """
+        LlamaRMSNorm is equivalent to T5LayerNorm
+        """
+        super().__init__(hidden_size, eps=eps)
+        setattr(self.weight, "sequence_parallel_enabled", sequence_parallel_enabled)
```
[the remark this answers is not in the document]
michaelbenayoun: I do it because they do it in `neuronx_distributed`: https://github.com/aws-neuron/neuronx-distributed/blob/main/examples/training/llama/modeling_llama_nxd.py#L144

That being said you raise a good question.

optimum/neuron/models/training/llama/modeling_llama.py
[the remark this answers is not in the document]
michaelbenayoun: Done.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+            self.intermediate_size,
+            self.hidden_size,
+            bias=False,
+            input_is_parallel=True,
+            init_method=init_method,
+            sequence_parallel_enabled=self.mp_config.sequence_parallel_enabled,
+            sequence_dimension=0,
+            dtype=self.config.torch_dtype,
+        )
+        self.split_size = self.intermediate_size // get_tensor_model_parallel_size()
```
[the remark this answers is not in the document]
michaelbenayoun: I will raise an error here.

optimum/neuron/models/training/llama/modeling_llama.py
michaelbenayoun: Removed the "fake" support for cache altogether.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+        self.config = config
+        self.mp_config = mp_config
+        self.hidden_size = config.hidden_size
+        self.intermediate_size = config.intermediate_size
+        self.act_fn = ACT2FN[config.hidden_act]
+
+        init_method = partial(_init_normal, config.initializer_range)
+
+        # Defines the MLP weight transformation specs
+        self.specs = ModelWeightTransformationSpecs(
```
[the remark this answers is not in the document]
michaelbenayoun: The first issue you mention is addressed. What do you mean by `fuse_axis` being redundant?

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Done.

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Done.

optimum/neuron/models/training/llama/modeling_llama.py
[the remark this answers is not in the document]
dacorvo: I thought maybe you could have deduced it by the type of Linear used, but after thinking about it a little bit more, being explicit is never a bad idea (versus sharing implicit knowledge).

tests/distributed/test_model_parallelization.py
```diff
@@ -736,3 +745,283 @@ def test_compute_query_indices_for_rank(
...
+
+def _custom_model_matches_original_model(
+    model_class_name,
+    model_name_or_path,
+    parallel_sizes,
+    sequence_parallel_enabled,
+    qkv_implementation,
+    attn_implementation,
+    monkeypatch,
+    torch_dtype=torch.float32,  # For now we do not match in bfloat16, so we test in float32.
```
michaelbenayoun, line 758: Investigate why it does not match with `bfloat16`.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -610,13 +738,123 @@ def forward(
...
+
+        return causal_mask
+
+
+class KwargsForCausalLM(FlashAttentionKwargs, LossKwargs): ...
+
+class LlamaForCausalLM(NeuronModelMixin, LlamaPreTrainedModel):
     _tied_weights_keys = ["lm_head.weight"]
     _tp_plan = {"lm_head": "colwise_rep"}
+    _pp_plan = {"lm_head": (["hidden_states"], ["logits"])}
```
tengomucho, line 794: So far we do not use the tp plan and the pp plan AFAIK. do you want to keep those anyway?

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -637,6 +875,24 @@ def __init__(self, config, mp_config: ModelParallelismConfig):
...
+
+    def set_output_embeddings(self, new_embeddings):
+        self.lm_head = new_embeddings
+
+    def set_decoder(self, decoder):
+        self.model = decoder
+
+    def get_decoder(self):
+        return self.model
+
```
tengomucho, line 835: I think you do not need these

optimum/neuron/models/training/transformations_utils.py
tengomucho: Add a comment on the file with this, so it will stay on the code 🙂

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -169,6 +169,26 @@ def to_original_weights(
...
 
+class CustomModule:
+    """
+    This class is used to mark a module as a custom module. It is used to identify the modules that contain weights
+    that need to transformed when loading and saving the model.
+    """
+    @property
+    def specs(self) -> ModelWeightTransformationSpecs:
+        if not hasattr(self, "_specs"):
+            self._specs = ModelWeightTransformationSpecs()
```
tengomucho, line 181: I think it would remove this. You can make `CustomModule` inherit from `ABC` to make clear it's an abstract class, and document what is required to implement it. If you want to make it not abstract instead, then add an `__init__` method and set `self._specs` there with a default value, so you do not have to have this check here.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -393,74 +838,530 @@ def from_pretrained(
...
-            local_files_only=local_files_only,
-            token=token,
-            revision=revision,
-            use_safetensors=use_safetensors,
-            use_safetensors_in_priority=True,
-            convert_to_safetensors=True,
-            **kwargs,
-        )
+        # ** Difference from original from_pretrained **
+        # In the original from_pretrained, here there some work done for quantization.
```
tengomucho, line 880:
```suggestion
        # In the original from_pretrained, here there was some initialization to support quantization.
```

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -128,10 +125,10 @@ def _load_state_dict_into_model(model_to_load, state_dict, start_prefix, assign_
     # We do not have the same weights as the state_dict since we use a custom model.
 
     # First we set the module names, this is required to have the specs properly defined.
-    set_module_names_in_transformation_specs(model_to_load)
+    # set_module_names_in_transformation_specs(model_to_load)
 
     # Then we adapt the state dict to the custom model.
-    state_dict = adapt_state_dict(model_to_load, state_dict, inplace=False)
+    # state_dict = adapt_state_dict(model_to_load, state_dict, inplace=False)
```
tengomucho: why are these commented?

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -1313,7 +1344,10 @@ def from_pretrained(
                         offload_folder=offload_folder,
                         offload_state_dict=offload_state_dict,
                         dtype=torch_dtype,
-                        hf_quantizer=hf_quantizer,
+                        # ** Difference from original from_pretrained **
+                        # We do not support hf_quantizer.
+                        hf_quantizer=None,
+                        # hf_quantizer=hf_quantizer,
```
tengomucho: remove the commented line

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: Done.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,714 @@
...
+        return config
+
+    # This method uses `torch.xla.utils.checkpoint.checkpoint` instead of the torch one.
+    def _set_gradient_checkpointing(self, enable: bool = True, gradient_checkpointing_func: Callable = checkpoint):
+        return super()._set_gradient_checkpointing(
+            enable=enable, gradient_checkpointing_func=gradient_checkpointing_func
+        )
+
+    @classmethod
+    def from_pretrained(
```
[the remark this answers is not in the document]
michaelbenayoun: I completely refactored this part. Now we own fully this function where I explain in detail where we diverge from what was done in Transformers.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,714 @@
...
+        # If weights are initially tied, we still copy the value but we do not tie them.
+        if should_fake_tie:
+            with torch.no_grad():
+                model.get_output_embeddings().weight.data.copy_(model.get_input_embeddings().weight)
+
+        xm.mark_step()
+
+        return model
+
+    def save_pretrained(
```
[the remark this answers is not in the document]
michaelbenayoun: Since it is a heavy work (+1000 line of diff just for this for `from_pretrained`) I suggest we create an item and handle this in a PR.

optimum/neuron/accelerate/accelerator.py
```diff
@@ -302,6 +303,14 @@ def _prepare_optimizer_for_zero_1(self, optimizer: torch.optim.Optimizer, device
 
     @patch_within_function(("accelerate.accelerator.AcceleratedOptimizer", NeuronAcceleratedOptimizer))
     def prepare_optimizer(self, optimizer: torch.optim.Optimizer, device_placement: Optional[bool] = None):
+        # If we use custom modeling, we do not have to do anything for now.
+        # We will have to do some work when supporting ZeRO-1.
+        model = self._models[0] if len(self._models) == 1 else None
+        if model is not None and inspect.getmodule(model.__class__).__name__.startswith(
+            "optimum.neuron.models.training"
+        ):
+            return super().prepare_optimizer(optimizer, device_placement=device_placement)
```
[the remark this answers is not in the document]
michaelbenayoun: It should since it is defined in `optimum/neuron/models/training`. 
The reason we skip the rest is because with custom modeling we do not need lazy loading or anything, so the optimizer is created with the proper parameters already.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+import torch.utils.checkpoint
+from torch import nn
+from transformers.activations import ACT2FN
+from transformers.cache_utils import Cache, DynamicCache, StaticCache
+from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
+from transformers.modeling_outputs import (
+    BaseModelOutputWithPast,
+    CausalLMOutputWithPast,
+)
+from transformers.models.llama.configuration_llama import LlamaConfig
```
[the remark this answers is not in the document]
michaelbenayoun: I made our `modeling_llama.py` completely independent from the one in Transformers now.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -0,0 +1,692 @@
...
+        get_tensor_model_parallel_rank,
+        get_tensor_model_parallel_size,
+    )
+else:
+    # We define this dummy function in case we do not have neuronx_distributed, for instance when building the docs.
+    def get_tensor_model_parallel_size():
+        return 0
+
+
+def create_local_fused_weight(tp_rank, tp_size, individual_weights, partition_dim, fuse_axis, out_weight=None):
```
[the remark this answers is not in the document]
michaelbenayoun: Done.
I have not removed the `sharding.py` file. Let's do it in the PR that changes Granite to use the overall API.

optimum/neuron/models/training/loss_utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+from ...distributed.utils import parallel_cross_entropy
+from ...utils import is_neuronx_distributed_available
+
+
+if is_neuronx_distributed_available():
+    from neuronx_distributed.parallel_layers.parallel_state import get_tensor_model_parallel_size
+
+
+_PARALLEL_CROSS_ENTROPY_SHOULD_PRESERVE_INPUT: bool = False
```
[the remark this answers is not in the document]
michaelbenayoun: Can we handle this later?
I feel it's pretty minor.

optimum/neuron/models/training/llama/modeling_llama.py
michaelbenayoun: Yes I do.
Because for instance it can be used in other methods such as `tie_weights`.

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: The `__init__` approach is annoying because in every module using it we will have to be careful about which `super().__init__` we are calling, that is the reason why I have done that.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+                            output_embeddings._is_hf_initialized = True
+            else:
+                not_initialized_submodules = dict(model.named_modules())
+
+            # ** Difference from original _load_pretrained_model **
+            # We initialize the parallel modules.
+            for name, mod in not_initialized_submodules.items():
+                if isinstance(mod, GQAQKVColumnParallelLinear):
+                    # There is a bug in initialization for this module.
+                    # In any case, we will always have weights for this in the case of `from_pretrained`.
```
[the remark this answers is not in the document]
michaelbenayoun: No because we wont call `from_pretrained` in this case.
I can investigate this issue but it does not seem top priority in this specific context.

tests/distributed/test_model_parallelization.py
```diff
@@ -736,3 +745,283 @@ def test_compute_query_indices_for_rank(
...
+
+
+def test_row_parallel_linear():
+    run_fn = partial(_test_parallel_linear, "row")
+    launch_procs(run_fn, 8, 8, 1)
+
+
+def test_column_parallel_linear():
+    run_fn = partial(_test_parallel_linear, "column")
+    launch_procs(run_fn, 8, 8, 1)
```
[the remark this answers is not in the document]
michaelbenayoun: I merged the two.

tests/distributed/test_model_parallelization.py
```diff
@@ -736,3 +745,283 @@ def test_compute_query_indices_for_rank(
...
+
+
+def test_flash_attention_v2():
```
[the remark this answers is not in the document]
michaelbenayoun: Done.

michaelbenayoun:
> Thank you for this pull-request: massive contribution. The general organization of the files makes sense and I think the `modelization_utils.py` and `transformation_utils.py` files are in particular good starting points. I have requested a few changes that I think are important in this first iteration to:
> 
> * make it clearer what is expected to be provided by someone adding a new model,
> * keep things simple in this first iteration (I personally think it is easier to add things we support later on instead of keeping placeholders).
> 
> Also, I think the basic features must be tested, and it was unclear to me by reading the pull-requests which parts are actually tested, apart from eager forward inference (and by the way we must include a flash_attention test since we support the option).

- What is expected is simply adding the proper transformation specs, and inheriting from `CustomModule`.
- In this PR we test:
   - The forward pass: we check the the forward pass from the original implementation and the one from the custom modeling produce the same outputs. We test different settings (eager attention, regular qkv, fused qkv, qkv gqa replication). The reason we do not test flash attention is because it does not match exactly for now. I was still able to train a model with flash attention though.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+    **kwargs,
+):
+    key_states = repeat_kv(key, module.num_key_value_groups)
+    value_states = repeat_kv(value, module.num_key_value_groups)
+
+    attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
+    if attention_mask is not None:
+        causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
+        attn_weights = attn_weights + causal_mask
+    elif causal:
```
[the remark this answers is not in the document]
michaelbenayoun: Done

optimum/neuron/models/training/modeling_utils.py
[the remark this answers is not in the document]
michaelbenayoun: Item added.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -0,0 +1,698 @@
...
+                else inputs_embeds.size(1)
+            )
+            cache_position = torch.arange(
+                past_seen_tokens, past_seen_tokens + current_length, device=inputs_embeds.device
+            )
+
+        if position_ids is None:
+            position_ids = cache_position.unsqueeze(0)
+
+        # This is not needed since we recompute a causal mask at each attention layer.
```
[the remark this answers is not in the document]
michaelbenayoun: I just tested overfitting and it works.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,714 @@
...
+            cls._check_and_enable_flash_attn_2(
+                config,
+                torch_dtype=torch_dtype,
+                device_map=device_map,
+                hard_check_only=False,
+                check_device_map=check_device_map,
+            )
+        # Keeping it if supported one day.
+        # elif requested_attn_implementation == "flex_attention":
+        #     config = cls._check_and_enable_flex_attn(config, hard_check_only=True)
```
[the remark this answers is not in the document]
michaelbenayoun: Done.

tests/test_trainers.py
michaelbenayoun: This does not use the custom modeling so this PR does not really interfere with this. It seems that it's more the new version of `neuronx_distributed` that is not able to trace the model with FX. 
It is most like linked to a combination of `neuronx_distributed` vs `transformers` version since they test on a very old version.

optimum/neuron/models/training/llama/modeling_llama.py
michaelbenayoun: Removed it since you are right we do not want to support this for now.

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Legacy from experimenting, removed these lines.

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Legacy from experimenting, removed these lines.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+            model_buffers = {".".join([prefix, key]) for key in model_buffers}
+        unexpected_keys = sorted(unexpected_keys - model_buffers)
+
+        # Clean up buffer for `inv-freq` because RoPE embedding moved under base model (https://github.com/huggingface/transformers/pull/34858)
+        has_inv_freq_buffers = any(buffer.endswith("rotary_emb.inv_freq") for buffer in model_buffers)
+        if has_inv_freq_buffers:
+            unexpected_keys = {k for k in unexpected_keys if "rotary_emb.inv_freq" not in k}
+
+        model.tie_weights()
+        # TODO: stopped here, start from here tomorrow.
```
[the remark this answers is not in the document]
michaelbenayoun: Done.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+                    # Load from a Flax checkpoint in priority if from_flax
+                    archive_file = os.path.join(pretrained_model_name_or_path, subfolder, FLAX_WEIGHTS_NAME)
+                elif use_safetensors is not False and os.path.isfile(
+                    os.path.join(pretrained_model_name_or_path, subfolder, _add_variant(SAFE_WEIGHTS_NAME, variant))
+                ):
+                    # Load from a safetensors checkpoint
+                    archive_file = os.path.join(
+                        pretrained_model_name_or_path, subfolder, _add_variant(SAFE_WEIGHTS_NAME, variant)
+                    )
+                elif use_safetensors is not False and os.path.isfile(
```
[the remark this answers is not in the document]
michaelbenayoun: It's not equivalent because `use_safetensors` can be `None`.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+                                    **has_file_kwargs,
+                                }
+                                if not has_file(pretrained_model_name_or_path, safe_weights_name, **has_file_kwargs):
+                                    Thread(
+                                        target=auto_conversion,
+                                        args=(pretrained_model_name_or_path,),
+                                        kwargs={"ignore_errors_during_conversion": True, **cached_file_kwargs},
+                                        name="Thread-auto_conversion",
+                                    ).start()
+                        else:
```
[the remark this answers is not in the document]
michaelbenayoun: I have not tested yet, but since it is supported in transformers, why not?

michaelbenayoun:
> A lot of work contributed here, thanks! This will help adding new models in a more efficient way. I added few comments, but It mostly resumes to these requests:
> 
> * [x]  Remove the dependency from transformers modeling code
> * [x]  Reuse and merge with existing code, avoid duplication for sharding and tests.
> * [ ]  Consider adding a complete test that shows training works (overfitting?)

I suggest we handle the last point in another PR to avoid making this PR bigger than it already is.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -0,0 +1,1549 @@
...
+        unexpected_keys = sorted(unexpected_keys - model_buffers)
+
+        # Clean up buffer for `inv-freq` because RoPE embedding moved under base model (https://github.com/huggingface/transformers/pull/34858)
+        has_inv_freq_buffers = any(buffer.endswith("rotary_emb.inv_freq") for buffer in model_buffers)
+        if has_inv_freq_buffers:
+            unexpected_keys = {k for k in unexpected_keys if "rotary_emb.inv_freq" not in k}
+
+        model.tie_weights()
+        # TODO: stopped here, start from here tomorrow.
+        if device_map is None:
```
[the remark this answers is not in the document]
michaelbenayoun: We support a subset of features: `device_map in [None, "xla", "cpu"]`.

tengomucho (changes requested):
Note that the linear tests now fail. I think that should be fixed before merging this.

tengomucho (approved):
LGTM, thanks a lot

PR #803, opened 2025-03-07
Training test refactoring

tengomucho:
# What does this PR do?

This refactoring is the first step of changes that aim at simplifying the training tests. 
For now functionally nothing has really changed, but the idea is to try to have multiple utility functions instead of a big class that handles a huge matrix of combinations, many of them skipped anyway.

So here there are many small changes in the test_model_parallelization.py and in particular in one test, that is split in two and where accepted parameters are then reduced, leading to exactly the same test coverage. 
While functionally it is the same, the intention is to cast a base to accelerate further refactoring and simplifications.


dacorvo (approved):
LGTM, thanks !

michaelbenayoun (approved):
LGTM thanks

PR #806, opened 2025-03-10
Enable back continuous batching for phi3 and cleanup hlo modeling

dacorvo:
# What does this PR do?

This disables the flash attention NKI kernel for phi3 models and refactors the HLO backend to cleanup the modeling.

### DIFF
--- a/optimum/exporters/neuron/model_configs/decoder_configs.py
+++ b/optimum/exporters/neuron/model_configs/decoder_configs.py
@@ -181,4 +181,4 @@ class GraniteNeuronConfig(NeuronDecoderExportConfig):
 @register_in_tasks_manager("phi3", "text-generation")
 class Phi3NeuronConfig(NeuronDecoderExportConfig):
     NEURONX_CLASS = Phi3HloModel
-    CONTINUOUS_BATCHING = False
+    CONTINUOUS_BATCHING = True
--- a/optimum/exporters/neuron/model_configs/decoder_configs.py
+++ b/optimum/exporters/neuron/model_configs/decoder_configs.py
@@ -55,6 +55,7 @@ class NeuronDecoderExportConfig(NeuronExportConfig):
 
     INPUT_ARGS = ("batch_size", "sequence_length")
     NEURONX_CLASS = None
+    ALLOW_FLASH_ATTENTION = False
     CONTINUOUS_BATCHING = False
     ATTENTION_lAYOUT = "HSB"
     FUSE_QKV = True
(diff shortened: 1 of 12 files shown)


tests/decoder/test_decoder_generation.py
```diff
@@ -171,3 +171,67 @@ def test_decoder_generation_stop_strings(model_and_tokenizer):
...
+    assert model.config.neuron["batch_size"] >= 3
+
+    inputs = tokenizer("Once upon a time", return_tensors="pt")
+
+    # A few helper functions we need while prefilling and decoding
+    def get_next_tokens(input_ids, cache_ids, start_ids):
+        outputs = model.forward(input_ids, cache_ids, start_ids)
+        return outputs.logits[:, -1, :].argmax(dim=-1, keepdim=True)
+
+    def increase_attention_mask(attention_mask):
```
tengomucho, line 200: if you increase the attention mask this way, inputs and attention mask change at each call. Could you use an input and a mask padded with zeros instead, and at each call place the right token for the next input and do the same for the ones in the mask? Otherwise you will end up having the function compiled at each step, don't you?

optimum/exporters/neuron/model_configs/decoder_configs.py
```diff
@@ -55,6 +55,7 @@ class NeuronDecoderExportConfig(NeuronExportConfig):
 
     INPUT_ARGS = ("batch_size", "sequence_length")
     NEURONX_CLASS = None
+    ALLOW_FLASH_ATTENTION = False
```
tengomucho, line 58: Since it works most of the time, would it make sense to do a flag with the opposite flag instead? something like `DISABLE_FLASH_ATTENTION = False` and then you set it only for Phi

tengomucho (commented):
Few questions, otherwise LGTM

optimum/exporters/neuron/model_configs/decoder_configs.py
dacorvo: That's what I would have done if the flag meant something for all models, but it only applies to those that are using the internal HLO modeling, so the default has to be false (for gpt2, etc...).

tests/decoder/test_decoder_generation.py
dacorvo: No, this is only the runtime tensors that are eventually padded and copied over to the device. With HLO modeling, there is no implicit compilation: the model is compiled explicitly once and invoked by the runtime code (look for program execute command in the HLO backend).

tengomucho: I see, thanks for the explanation. Do you know if there is going to be an overhead to allocate new buffers (instead of reusing the ones with the given shapes)?

tengomucho (approved):

PR #807, opened 2025-03-11
Training remove gpt neo models support

tengomucho:
# What does this PR do?

This is part of the training tests cleanup. The idea is to remove support for models that are not very well tested and have a relatively small audience, such as gpt-neo and gpt-neox.

### DIFF
--- a/optimum/neuron/distributed/decoder_models.py
+++ b/optimum/neuron/distributed/decoder_models.py
@@ -22,7 +22,6 @@ import torch.nn as nn
 import torch.nn.functional as F
 from transformers.cache_utils import Cache
 from transformers.models.gpt_neo.modeling_gpt_neo import GPTNeoBlock, GPTNeoSelfAttention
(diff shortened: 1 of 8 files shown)


michaelbenayoun (commented):
I wonder if we should at least keep GPTNeoX?
I think it was requested at some point.

tengomucho:
GPT NeoX 20b model has been downloaded only 20k times over last month, that is very low compared to other models we support or should support (e.g: Llama 3.2 3B has more than 3M alone,  Llama 3.1 8B 6M, Mistral 0.3 7B was downloaded 120k times). I think we should remove it, it increases the code base without adding much value to it (Without mentioning that in the future we should add custom modeling code to support it). If somebody asks for it, we can always add it back.

dacorvo (approved):
LGTM, thanks. We should also remove support for this model architecture in LLM inference (they are not cached and absolutely no one complained).

michaelbenayoun (approved):
LGTM

PR #811, opened 2025-03-19
Fix broken cache for traced models & fix runtime error of diffusion models when batch_size > 1

JingyaHuang:
# What does this PR do?

as per title.

### DIFF
--- a/optimum/exporters/neuron/__main__.py
+++ b/optimum/exporters/neuron/__main__.py
@@ -213,7 +213,7 @@ def normalize_stable_diffusion_input_shapes(
             f"Shape of {mandatory_axes} are mandatory for neuron compilation, while {mandatory_axes.difference(args.keys())} are not given."
         )
     mandatory_shapes = {name: args[name] for name in mandatory_axes}
-    mandatory_shapes["num_images_per_prompt"] = args.get("num_images_per_prompt", 1)
+    mandatory_shapes["num_images_per_prompt"] = args.get("num_images_per_prompt", 1) or 1
     mandatory_shapes["sequence_length"] = args.get("sequence_length", None)
     input_shapes = build_stable_diffusion_components_mandatory_shapes(**mandatory_shapes)
     return input_shapes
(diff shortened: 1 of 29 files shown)


optimum/neuron/utils/hub_cache_utils.py
```diff
@@ -78,8 +78,9 @@ def create_compile_cache():
     "sample_size",
     "projection_dim",
     "_use_default_values",
+    "_attn_implementation_autoset",
 ]
-NEURON_CONFIG_WHITE_LIST = ["input_names", "output_names", "model_type"]
+NEURON_CONFIG_WHITE_LIST = ["input_names", "output_names", "model_type", "static__commit_hash"]
```
tengomucho, line 83: Not sure, but shouldn't this be
```suggestion
NEURON_CONFIG_WHITE_LIST = ["input_names", "output_names", "model_type", "static_commit_hash"]
```

setup.py
```diff
@@ -32,7 +32,7 @@
     "diffusers>=0.29.0, <=0.30.3",
     "safetensors",
     "sentence-transformers >= 2.2.0",
-    "peft>=0.14.0",
+    "peft>=0.14.0, <0.15.0",
```
tengomucho, line 35: we could pin this on 0.14.0, as in training extra

optimum/neuron/utils/hub_cache_utils.py
JingyaHuang: Good catch!!! it comes from latest transformers, I should have add the arg instead of just adding it to the white list. Thanks Alvaro!

tengomucho (approved):
LGTM

optimum/exporters/neuron/__main__.py
```diff
@@ -213,7 +213,7 @@ def normalize_stable_diffusion_input_shapes(
             f"Shape of {mandatory_axes} are mandatory for neuron compilation, while {mandatory_axes.difference(args.keys())} are not given."
         )
     mandatory_shapes = {name: args[name] for name in mandatory_axes}
-    mandatory_shapes["num_images_per_prompt"] = args.get("num_images_per_prompt", 1)
+    mandatory_shapes["num_images_per_prompt"] = args.get("num_images_per_prompt", 1) or 1
```
dacorvo, line 216: I don't understand the change here.

JingyaHuang: `num_images_per_prompt` could be `None`, so set to 1 if so.

PR #814, opened 2025-03-24
chore(test): add test comparing Linear and RowParallelLinear outputs

tengomucho:
A test is added to compare the outputs of a Linear to the ones of the layer RowParallelLinear, that is supposed to be equivalent. The tests reveal that outputs are close when the input type is float32, but they are not when inputs are of type bfloat16.

### DIFF
--- a/tests/distributed/test_linears.py
+++ b/tests/distributed/test_linears.py
@@ -89,18 +89,26 @@ def _test_parallel_model_check(weights_dtype, inputs_dtype, input_size, output_s
 
 
 @is_trainium_test
-@pytest.mark.parametrize("weights_dtype, inputs_dtype", [
-    (torch.float32, torch.float32),
-    (torch.bfloat16, torch.bfloat16),
-    (torch.bfloat16, torch.float32),
-    (torch.float16, torch.float16),
-], ids=["weights=float32-inputs=float32",
+@pytest.mark.parametrize(
+    "weights_dtype, inputs_dtype",
+    [
+        (torch.float32, torch.float32),
+        (torch.bfloat16, torch.bfloat16),
+        (torch.bfloat16, torch.float32),
+        (torch.float16, torch.float16),
+    ],
+    ids=[
+        "weights=float32-inputs=float32",
         "weights=bfloat16-inputs=bfloat16",
         "weights=bfloat16-inputs=float32",
-        "weights=float16-inputs=float16",])
-@pytest.mark.parametrize("input_size, output_size", [
-    (8192, 2048), (400, 300)
-], ids=["input_size=8192-output_size=2048", "input_size=400-output_size=300"])
+        "weights=float16-inputs=float16",
+    ],
+)
+@pytest.mark.parametrize(
+    "input_size, output_size",
+    [(8192, 2048), (400, 300)],
+    ids=["input_size=8192-output_size=2048", "input_size=400-output_size=300"],
+)
 def test_parallel_linears(weights_dtype, inputs_dtype, input_size, output_size):
     run_fn = partial(
         _test_parallel_model_check,


tests/distributed/test_linears.py
```diff
@@ -0,0 +1,118 @@
...
+        tp_size=2,
+        pp_size=1,
+    )
+
+
+@is_trainium_test
+@pytest.mark.xfail(reason="The output of the parallel model is not matching the original one")
+def test_parallel_layers_bfloat16_inputs_bfloat16():
+    _test_parallel_linears(torch.bfloat16, inputs_dtype=torch.bfloat16)
+
```
dacorvo, line 110: Pretty sure you need to skip one more line here.

tests/distributed/test_linears.py
```diff
@@ -0,0 +1,118 @@
...
+    launch_procs(
+        run_fn,
+        num_procs=2,
+        tp_size=2,
+        pp_size=1,
+    )
+
+
+@is_trainium_test
+@pytest.mark.xfail(reason="The output of the parallel model is not matching the original one")
```
dacorvo, line 107: What about float16 ?

tests/distributed/test_linears.py
```diff
@@ -0,0 +1,118 @@
...
+        # Filtering items to slice only the weights of this layer
+        filtered_items = filter(lambda x: x[0].startswith(prefix), state_dict.items())
+        world_size = xrt.world_size()
+        for k, v in filtered_items:
+            if k.endswith("linear1.weight"):
+                axis_len = v.shape[1]
+                split_len = (axis_len + world_size - 1) // world_size
+                partition_stride = 1  # assuming that is always 1
+                state_dict[k] = create_local_weight(v, 1, split_len, partition_stride)
+                # state_dict[k] = slice_tensor(v, 1)
```
dacorvo, line 67: Maybe remove this comment ?

dacorvo (commented):
Be careful with bfloat16, as this type has a very high resolution:
```
>>> torch.finfo(torch.bfloat16)
finfo(resolution=0.01, min=-3.38953e+38, max=3.38953e+38, eps=0.0078125, smallest_normal=1.17549e-38, tiny=1.17549e-38, dtype=bfloat16)
```
Which means that atol should be at least 1e-2.

Some discussion about the precision to be expected for half here:
https://scicomp.stackexchange.com/questions/43111/float-equality-tolerance-for-single-and-half-precision

tests/distributed/test_linears.py
```diff
@@ -0,0 +1,118 @@
...
+
+class ParallelSampleModel(nn.Module):
+    def __init__(self, weights_dtype):
+        super().__init__()
+        self.linear1 = RowParallelLinear(
+            INPUT_SIZE,
+            OUTPUT_SIZE,
+            bias=False,
+            dtype=weights_dtype,
+            sequence_parallel_enabled=False,
```
michaelbenayoun, line 49: Can you try that as well?

```suggestion
            dtype=weights_dtype,
            reduce_dtype=weights_dtype,
            sequence_parallel_enabled=False,
```

The default is `torch.float32`. Maybe in the non-parallel implemetation the reduce dtype is `bfloat16`.

tengomucho: Here's the params in the example they provide [here](https://github.com/aws-neuron/neuronx-distributed/blob/7b69388089e29a4210c3f88a5aa355cd41da36d4/examples/training/llama/modeling_llama_nxd.py#L184). 
You are right I can play with the reduce_dtype, I would set it to the input_dtype though, it makes more sense to me: that is the type supposed to be at output.

dacorvo (approved):
lgtm, thanks !

Issue #676, opened 2024-08-06
Add support for new Black Forest's model (Flux)


mrrfr:
### Feature request

Hello Guys,

The new model of Black Forest Labs is very nice (https://huggingface.co/black-forest-labs/FLUX.1-schnell) It have already been adopted to diffusers (https://github.com/huggingface/diffusers/tree/926daa30f977a75c01b835405c1878730388cb94/src/diffusers/pipelines/flux)

I've tried something on my own, but nothing successful.

### Motivation

Flux is actually the best model for txt2img, I've tried whole the weekend to adapt it for inf2 with neuronx but nothing successful.  

### Your contribution

So far, none of the work done can be shared as I have only encountered failures.

Thank u & Have a nice day :))

JingyaHuang:
Thanks @mrrfr for raising the issue. There will be soon some updates on diffusion transformer support with Neuron SDK. Happy to support flux in Optimum Neuron by this time, and all contribution is very welcomed!

yahavb:
We would like to see that too.

Suprhimp:
Does this feature will be added soon?

JingyaHuang:
Flux will be the next (after pixart) diffusion model that we will support. @Suprhimp

zwx00:
Hey guys, any update? We'd love to test this out!

JingyaHuang:
Hi @zwx00, we are actively working on this, you can follow it here #763

PR #815, opened 2025-03-24
Add BlackForest Flux Support 

### LINKED ISSUE #763: Support for Flux model in diffusers
### Feature request

Hey @yahavb, to continue our discussion on supporting the Flux model here. 

### Add export support for each component

* Text encoder 1

No need to add the wrapper, it should be already handled by `CLIPTextNeuronConfig`.

* Text encoder 2

Can reuse the `T5EncoderForDiffusersNeuronConfig` class, in which the `T5EncoderWrapper` is equivalent to `TracingT5TextEncoderWrapper`. 
For tp support, will need to enable `tensor_parallel_size` arg for many functions in `optimum/exporters/neuron/__main__.py` (eg. `_get_submodels_and_neuron_configs_for_stable_diffusion`). And we need to write a `get_parallel_callable` for the neuron config class to shard the model, we can take the one in `T5EncoderForTransformersNeuronConfig` as a reference.

* Transformer

In the [script](https://github.com/yahavb/aws-neuron-eks-samples/blob/master/flux_serve/app/src/transformer/compile.py), transformer_embedders / transformer_blocks / ingle_transformer_blocks / transformer_out_layers are traced separately, would it be possible to fit into neuron devices while tracing the whole transformer together? If it's not the case, in Optimum Neuron logic, each traced component means a NeuronConfig class, then you will need to create a class for each component and build a function to fetch each module (similar to `_get_submodels_and_neuron_configs_for_stable_diffusion`), otherwise, you would be able to create a single `FluxTransformerNeuronConfig`, and add the tp config there.

* VAE decoder

Nothing to do, already supported

### Test

Here is an example of testing the compilation:

```bash
optimum-cli export neuron --model black-forest-labs/FLUX.1-dev --tensor_parallel_size 8 --batch_size 1 --height 1024 --width 1024 --num_images_per_prompt 1 --sequence_length 512 --torch_dtype bfloat16 flux_neuron/
```

## 2nd step: Inference support

Depends on the 1st step to decide if we need to override some functions in `FluxPipeline` (if we traced transformer in multiple module), but whatever we need to put under `optimum/neuron/pipelines/diffusers`. We will need to add the following class under `optimum/neuron/modeling_diffusion.py`:

```python
class NeuronFluxPipeline(NeuronDiffusionPipelineBase, FluxPipeline):
    main_input_name = "prompt"
    auto_model_class = FluxPipeline
```

Reference PRs:

* Add TP support for T5 model: #697
* Add Pixart support: #762 

### Motivation

m

### Your contribution

,

JingyaHuang:
# What does this PR do?

Fixes #763,  #676 

## Compilation

- [x] Export of Flux pipeline (`TP=8` for Flux transformer 2D) with `neuronx_distributed.trace.model_builder.ModelBuilder`

### Export via CLI

* Regular 

```bash
optimum-cli export neuron --model black-forest-labs/FLUX.1-dev --tensor_parallel_size 8 --batch_size 1 --height 768 --width 1360 --num_images_per_prompt 1 --torch_dtype bfloat16 flux_neuron/
```

* Tiny test

```bash
optimum-cli export neuron --model hf-internal-testing/tiny-flux-pipe-gated-silu --tensor_parallel_size 2 --batch_size 1 --height 8 --width 8 --num_images_per_prompt 1 --sequence_length 256 --torch_dtype bfloat16 tiny_flux_neuron/
```

### Export with `NeuronFluxPipeline` API

```python
from optimum.neuron import NeuronFluxPipeline

if __name__ == "__main__":
    compiler_args = {"auto_cast": "none"}
    input_shapes = {"batch_size": 1, "height": 1024, "width": 1024}

    pipe = NeuronFluxPipeline.from_pretrained(
        "black-forest-labs/FLUX.1-dev",
        torch_dtype=torch.bfloat16,
        export=True,
        tensor_parallel_size=8,
        # disable_neuron_cache=True,
        **compiler_args,
        **input_shapes
    )

    # Save locally
    pipe.save_pretrained("flux_dev_neuron_1024_tp8/")

    # Upload to the HuggingFace Hub
    pipe.push_to_hub(
        "flux_dev_neuron_1024_tp8/", repository_id="Jingya/FLUX.1-dev-neuronx-1024x1024-tp8"  # Replace with your HF Hub repo id
    )

```

You can find an example of compiled artifacts [here (`Jingya/flux.1-dev_neuronx_tp8` )](https://huggingface.co/Jingya/flux.1-dev_neuronx_tp8/tree/main)

## Inference

- [x] Flux Inference

For generating an image with `NeuronFluxPipeline`:

```python
from optimum.neuron import NeuronFluxPipeline

pipe = NeuronFluxPipeline.from_pretrained("flux_neuron")
prompt = "A cat holding a sign that says hello world"
out = pipe(
    prompt,
    guidance_scale=3.5,
    num_inference_steps=50,
    generator=torch.Generator("cpu").manual_seed(0)
).images[0]
out.save("flux_optimum.png")
```

## Other
- [x] Tests
- [x] Doc

cc. @yahavb


optimum/exporters/neuron/convert.py
```diff
@@ -570,34 +626,72 @@ def export_neuronx(
...
+            model_builder.add(
+                key=subfolder,
+                model_instance=model,
+                example_inputs=[dummy_inputs],
+                priority_model_idx=0,
+                compiler_args=compiler_args,
+            )
+            neuron_model = model_builder.trace(initialize_model_weights=False)
+
+            model_builder.shard_checkpoint(serialize_path=output.parent / "weights/")
```
dacorvo, line 675: Eventually this could be omitted: the weights can be sharded at loading time only, which makes export a lot faster. You only need to remember where the weights are (local dir or hub).

optimum/neuron/models/inference/flux/modeling_flux.py
```diff
@@ -0,0 +1,904 @@
...
+        )
+        self.proj_out = ColumnParallelLinear(
+            self.inner_dim,
+            patch_size * patch_size * self.out_channels,
+            bias=True,
+            gather_output=True,
+            reduce_dtype=reduce_dtype,
+        )
+
+    def forward(
```
dacorvo, line 206: It does not look like this method has anything neuron specific: wouldn't have been possible to inherit from the `diffusers` class and just override `__init__` ?

tests/inference/test_nxd.py
```diff
@@ -0,0 +1,66 @@
+# coding=utf-8
```
dacorvo: You should rename the test file to make it flux specific, like `test_flux_pipeline.py`.

tests/inference/test_nxd.py
```diff
@@ -0,0 +1,66 @@
...
+    NeuronModelVaeEncoder,
+)
+from optimum.neuron.utils.testing_utils import is_inferentia_test, requires_neuronx
+from optimum.utils.testing_utils import require_diffusers
+
+
+@is_inferentia_test
+@requires_neuronx
+@require_diffusers
+class NeuronFluxPipelineIntegrationTest(unittest.TestCase):
```
dacorvo: Is it required to wrap this in a `unittest.TestCase` ? It looks like it could be a simple test function, which is easier to test individually and parametrize with `pytest`.

docs/source/model_doc/diffusers/flux.mdx
```diff
@@ -0,0 +1,120 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Flux
+
+Flux is a series of text-to-image generation models based on diffusion transformers.
+
+> [!TIP]
+> We recommend using a `inf2.24xlarge` instance with tensor parallel size 8 for the model compilation and inference.
```
tengomucho, line 22: why 24x if we only do TP8?

tests/inference/test_nxd.py
```diff
@@ -0,0 +1,66 @@
...
+            **input_shapes,
+            **compiler_args,
+        )
+        self.assertIsInstance(neuron_pipeline.text_encoder, NeuronModelTextEncoder)
+        self.assertIsInstance(neuron_pipeline.text_encoder_2, NeuronModelTextEncoder)
+        self.assertIsInstance(neuron_pipeline.transformer, NeuronModelTransformer)
+        self.assertIsInstance(neuron_pipeline.vae_encoder, NeuronModelVaeEncoder)
+        self.assertIsInstance(neuron_pipeline.vae_decoder, NeuronModelVaeDecoder)
+
+        prompt = "Mario eating hamburgers."
```
dacorvo: Another food-based prompt 🤔 ... is it your signature 😉 ?

dacorvo (changes requested):
Massive pull-request: impressive !!!! I only have a few minor comments.
Eventually, we should think about how some of the `ModelBuilder` related code can be reused for non-decoder models: it is unclear to me yet if there are some specifics that would prevent that though.

optimum/exporters/neuron/base.py
```diff
@@ -380,17 +382,24 @@ def unflatten_inputs(cls, inputs: dict[str, Any]) -> dict[str, Any]:
...
-        dummy_inputs: dict[str, torch.Tensor] | None = None,
+        input_names: list[str] = None,
         forward_with_tuple: bool = False,
         eligible_outputs: list[str | int] | None = None,
         device: str | None = None,
     ):
         """
-        Checks if inputs order of the model's forward pass correspond to the generated dummy inputs to ensure the dummy inputs tuple used for
-        tracing are under the correct order.
+        Patch the model and generate aliased for tracing.
```
tengomucho:
```suggestion
        Patch the model and generate aliases for tracing.
```

optimum/exporters/neuron/convert.py
```diff
@@ -529,20 +532,68 @@ def export_neuronx(
...
+        dummy_inputs_tuple = tuple(dummy_inputs.values())
+        return dummy_inputs_tuple
+
+
+def prepare_compiler_flags(
+    config: "NeuronDefaultConfig",
+    auto_cast: str | None = None,
+    auto_cast_type: str = "bf16",
+    optlevel: str = "2",
+    instance_type: str = "trn1",
```
tengomucho, line 595: I thought you suggested to use inf2

optimum/exporters/neuron/utils.py
```diff
@@ -436,42 +471,47 @@ def get_submodels_for_export_diffusion(
...
-            if attn_mask is not None:
-                return sdpa_original(query, key, value, attn_mask=attn_mask, dropout_p=dropout_p, is_causal=is_causal)
-            else:
-                return neuron_scaled_dot_product_attention(
-                    query, key, value, attn_mask=attn_mask, dropout_p=dropout_p, is_causal=is_causal
-                )
-
-        torch.nn.functional.scaled_dot_product_attention = attention_wrapper
-        models_for_export.append((DIFFUSION_MODEL_TRANSFORMER_NAME, copy.deepcopy(transformer)))
+        if not is_flux:  # The following will be handled by `ModelBuilder` if `is_flux`.
```
tengomucho, line 480: since you do not have an "else" here, maybe you can combine the two if and reduce indentation.

optimum/exporters/neuron/convert.py
JingyaHuang: Do you mean we shard the weights during the loading time instead of after the tracing? Indeed, it increases the whole export time, but somehow, I would rather spend more time exporting one shot and have faster loading + warmup during the deployment.

tengomucho (approved):
Few nits, otherwise LGTM!

optimum/neuron/models/inference/flux/modeling_flux.py
JingyaHuang: here I change the inputs from `txt_ids` and `img_ids` in diffusers to `image_rotary_emb` that we will compute during the inference runtime on CPU.

optimum/exporters/neuron/convert.py
dacorvo: Maybe in your case it is not that bad, because sharding does not mean loading the weights on the device, does it ?
It is just that you don't cache the sharded weights, so it is only useful when you want to push the exported model to the hub.
This is what I do for decoders:
- when using the optimum-cli: export (cache or fetch NEFFs)
- when using from_pretrained: export (cache or fetch NEFFs)  + load_weights
For large models like llama 70B this makes a huge difference as loading weights takes several minutes.

tests/inference/test_nxd.py
JingyaHuang: no, pytest is much better. Let me change it.

tests/inference/test_nxd.py
JingyaHuang: doing so much physio these days, need energy from these food prompts...

docs/source/model_doc/diffusers/flux.mdx
JingyaHuang: Because inf2 jump from either you want to have just 1 neuron device, or you want to have 6
image

optimum/exporters/neuron/convert.py
JingyaHuang: interestingly the compiler flag we put for inf2 export is "trn1"

JingyaHuang: don't ask why

optimum/exporters/neuron/utils.py
JingyaHuang: yeah but for pixart (case diffusion transformer but not flux) I would still need to apply the following:

https://github.com/huggingface/optimum-neuron/blob/7746b1c27fefe87b95e667705a3d1c16cf3e1670/optimum/exporters/neuron/utils.py#L497-498

docs/source/model_doc/diffusers/flux.mdx
tengomucho: oh, right

optimum/exporters/neuron/convert.py
JingyaHuang: na na, `shard_checkpoint` just shard the weights, then we either serialize things to disk like I do here or load to Neuron.

JingyaHuang: Yeah, I should definitly look into the neff cache part of ModelBuilder (not yet the case) and the work you have done that I could reuse. will do it next!

dacorvo (approved):
LGTM, thanks !

tengomucho (approved):

PR #824, opened 2025-04-02
test(training): add flash attention test

tengomucho:
Added test for nki flash attention as it is used in training modeling code. The test reveals that flash attention does not seem reliable when used with float32 dtype.

This should help implementation of modeling using this kernel implementation.

### DIFF
--- a/tests/distributed/test_flash_attn.py
+++ b/tests/distributed/test_flash_attn.py
@@ -1,0 +1,105 @@
+import pytest
+import torch
+from torch import nn
+from transformers import AutoConfig, set_seed
+
+from optimum.neuron.utils.import_utils import (
+    is_neuronx_distributed_available,
+    is_torch_xla_available,
+)
+from optimum.neuron.utils.testing_utils import is_trainium_test
+
+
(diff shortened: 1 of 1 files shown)


dacorvo (approved):
LGTM, thanks !

tests/distributed/test_flash_attn.py
```diff
@@ -0,0 +1,105 @@
...
+    [
+        pytest.param(
+            "meta-llama/Llama-3.2-1B",
+            torch.float32,
+            marks=pytest.mark.xfail(strict=True, reason="Flash attention does not seem to work right in float32"),
+        ),
+        ("ibm-granite/granite-3.2-2b-instruct", torch.bfloat16),
+    ],
+    ids=["llama", "granite"],
+)
```
michaelbenayoun, line 44: I do not know if it is needed to parametrize on the model configs since we use random values anyways. 
I am not against it at all, but I'm not sure it adds value.

tengomucho: The reason is that this way we can test with dimensions that come from models we want to support.

PR #791, opened 2025-02-20
Initial PR for the documentation refactoring

JingyaHuang:
# What does this PR do?

Related: #790 

This is the 1st PR for the refactoring of Optimum Neuron's documentation, here below are the goals for this PR:

- [x] Re-structure the doc catalog
- [x] A representative model doc for each category
    - [x] transformers: text/vision/audio/multimodal; 
    - [x] diffusers; 
    - [x] sentence-transformer
* All docstring now moved to `optimum/neuron/utils/doc.py` -> according changes in `modeling.py`, `modeling_diffusion.py`, `modeling_seq2seq.py`. 
* Modeling code for each representative under `optimum/neuron/models`, these are more or less the same as previous `NeuronModelForXXX` but crafted more for the specific model type according to their inputs. Nothing new invented.
* Several bug fixes in the `pipeline` API while testing the code snippet
* CLIP
    * Rename CLIP wrapper and neuron config (with projection layer it more proper to name this way to distinguish from ClipVisionModel)
    * Well separate using CLIP via transformers v.s. sentence transformers
- [ ] Inference modeling code for each model type, so far I just think about replicating code as usually there is only `forward` to be overridden for each class (if it becomes more complicated, we could apply a similar mechanism to [Modular Transformers](https://huggingface.co/docs/transformers/en/modular_transformers))

PR #825, opened 2025-04-03
Update modeling_seq2seq.py - comment nit

jimburtoft:
Missing AutoProcessor import in example

# What does this PR do?

You have some example code you display that uses AutoProcessor, but doesn't import it

This isn't really code I'm fixing, just an example / comment that has code.

Metacode?

## Before submitting
- [Kind of ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).


JingyaHuang:
Hi @jimburtoftm, That's a good catch. Thanks for opening the PR. I'm working on a refactoring of Optimum Neuron's documentation (#791 ) that would make this obsolete.

PR #829, opened 2025-04-08
Refactor hub neuronx cache

dacorvo:
# What does this PR do?

This refactors the hub neuronx cache to prepare the support of cached NxD LLM models.

### Refactor the code to make it more modular

The current implementation is used in several configurations.

1 - Using caching at the NEFF level
- for the inference of LLM
- for the training of decoder models (either legacy ones like Bert of LLMs like LLama).

2 - Using caching at the model level (torch script modules)
- for the inference of transformers single models (like Bert for sequence classification),
- for the inference of  transformers multiple models (like encoder-decoder models),
- for the inference of diffusion pipelines.

This makes the code contained in the single `hub_cache_utils.py` difficult to understand and modify.

This pull-request splits the code in multiple files under the `optimum/neuron/cache` directory:

- the `hub_cache.py` file contains the high-level cache API,
- the `entries` directory defines the `ModelCacheEntry` abstract class that is specialized in two entries for single model and multiple models entries,
- the `traced.py` and `training.py` contains specific methods that are used only by these configurations.

### Add task in cached model registry

The pull-request also introduces the `task` as a mandatory parameter when creating an entry in the registry, because some models will have a different static graph for different tasks.

Adding this parameter required small modifications:
- in the traced model export code: task is now passed to `export_models`,
- in the neuron trainer: task is deduced from model.

### Better isolate standard model config from neuron specific configuration parameters

Finally, the pull-request tries to preserve the model config/configs as stored in the cache registry to simplify the compatibility checks. For that reason, the `task` parameter is now stored in the `neuron_config` for traced models only (and not in the `config` itself).

### DIFF
--- a/optimum/exporters/neuron/convert.py
+++ b/optimum/exporters/neuron/convert.py
@@ -25,6 +25,9 @@ import torch
 from transformers import PreTrainedModel
 
 from ...exporters.error_utils import OutputMatchError, ShapeError
+from ...neuron.cache.entries.multi_model import MultiModelCacheEntry
+from ...neuron.cache.entries.single_model import SingleModelCacheEntry
+from ...neuron.cache.traced import cache_traced_neuron_artifacts
 from ...neuron.utils import (
     DiffusersPretrainedConfig,
     convert_neuronx_compiler_args_to_neuron,
@@ -34,11 +37,6 @@ from ...neuron.utils import (
     store_compilation_config,
 )
 from ...neuron.utils.cache_utils import get_model_name_or_path
-from ...neuron.utils.hub_cache_utils import (
-    ModelCacheEntry,
-    build_cache_config,
-    cache_traced_neuron_artifacts,
-)
 from ...neuron.utils.version_utils import get_neuroncc_version, get_neuronxcc_version
 from ...utils import (
     is_diffusers_available,
@@ -295,6 +293,7 @@ def export_models(
     models_and_neuron_configs: Dict[
         str, Tuple[Union["PreTrainedModel", "ModelMixin", torch.nn.Module], "NeuronDefaultConfig"]
     ],
+    task: str,
     output_dir: Path,
     disable_neuron_cache: Optional[bool] = False,
     compiler_workdir: Optional[Path] = None,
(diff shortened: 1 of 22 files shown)


optimum/neuron/trainers.py
```diff
@@ -218,10 +223,15 @@ def __init__(self, *args, **kwargs):
             "tensor_parallel_size": self.args.tensor_parallel_size,
             "pipeline_parallel_size": self.args.pipeline_parallel_size,
         }
-        self.model_cache_entry: Optional[ModelCacheEntry] = None
+        self.model_cache_entry: Optional[SingleModelCacheEntry] = None
         if model_name_or_path_for_cache_entry is not None:
+            task = TasksManager.infer_task_from_model(self.model)
```
tengomucho, line 228: hopefully this guesses the right task!

tengomucho (approved):
LGTM, thanks!

optimum/neuron/trainers.py
dacorvo: Yes, it is usually straightforward if you have the model, since each model class is a FooModelForTask.

optimum/neuron/utils/argument_utils.py
```diff
@@ -318,6 +319,4 @@ def store_compilation_config(
 
         update_func("_diffusers_version", diffusers.__version__)
 
-    update_func("task", task)
```
JingyaHuang, line 321: why moving task under config.neuron?

JingyaHuang (commented):
Thanks for the refactoring! Left some questions especially on the modification of "task".

optimum/neuron/utils/argument_utils.py
dacorvo: Because I wanted to keep the config unmodified for easier comparison between models.
Ultimately, I would like the config to be untouched and neuron_config serialized independently.
Also, for diffusion models, there is a different task for each sub-model, so it is better to have it in neuron_config (since there is a neuron_config for each sub-model from what I see).

optimum/exporters/neuron/convert.py
```diff
@@ -310,6 +309,8 @@ def export_models(
     Args:
         models_and_neuron_configs (`Dict[str, Tuple[Union["PreTrainedModel", "ModelMixin", torch.nn.Module], `NeuronDefaultConfig`]]):
             A dictionnary containing the models to export and their corresponding neuron configs.
+        task (`str`):
```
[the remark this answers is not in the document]
dacorvo: We still do for sub-models, but this one is the "meta" task, like "text-to-image" for diffusion models.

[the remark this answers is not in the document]
JingyaHuang: for diffusion models we compile all components in the pipeline, `text-to-image` or `image-to-image` they all lead to the same artifacts, the difference is which components to load and use during the inference. With this PR, recompilation will be triggered for launching the same compilation process.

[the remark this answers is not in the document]
dacorvo: OK, I hadn't understood that. This can be avoided by passing null to the task when creating the cache entry for diffusion models then.

PR #830, opened 2025-04-09
Granite modeling for training

tengomucho:
# What does this PR do?

This PR introduces the modeling code for Granite, based on NxD core modeling example for Llama. 
So far it is possible to instantiate `ibm-granite/granite-3.2-2b-instruct` and perform an inference on the sharded model.
The modeling code is based on transformers, and adapted to include sharding (only tested with TP parallelism).

- MLP Linear layers are parallelized and fused.
- Linear layers in Attention are parallelized too.

Flash attention is not used so far because it requires not-padded prompts to generate correct outputs. It was not possible to use fused layers, because in Granite `self.num_heads != self.num_key_value_heads`. The attention kernel implementation is then just the eager kernel.

This is NOT a complete implementation for training: parallel loss is not implemented, as well as a complete training test and/or example. These will follow up soon, but I thought it was better to submit this to get feedback sooner.

### DIFF
--- a/optimum/neuron/configuration_utils.py
+++ b/optimum/neuron/configuration_utils.py
@@ -56,7 +56,7 @@ class NeuronConfig(PushToHubMixin):
     """
     Class that holds a configuration for a neuron model.
 
-    Must be subclassed by backend type.
+    Must be subclassed.
 
     """
 
@@ -222,7 +222,7 @@ class NeuronConfig(PushToHubMixin):
         Returns:
             [`NeuronConfig`]: The configuration object instantiated from those parameters.
         """
-        for key in ["optimum_neuron_version", "backend", "neuronxcc_version"]:
+        for key in ["optimum_neuron_version", "_serialized_key", "neuronxcc_version"]:
             if key not in config_dict:
                 raise ValueError(f"Invalid neuron configuration: missing mandatory {key} attribute.")
         # Check versions
@@ -238,17 +238,17 @@ class NeuronConfig(PushToHubMixin):
                 "This neuron configuration corresponds to a model exported using version"
                 f" {cfg_neuronxcc_version} of the neuronxcc compiler, but {neuronxcc_version} is installed."
             )
-        # Retrieve neuron config class from backend
-        backend = config_dict.pop("backend")
+        # Retrieve neuron config class from serialized key
+        _serialized_key = config_dict.pop("_serialized_key")
         if cls is NeuronConfig:
-            # We need to identify the actual neuron configuration class for this backend
-            if backend is None:
-                raise ValueError("Neuron configuration is invalid: unable to identify the backend")
-            cls = _NEURON_CONFIG_FOR_KEY.get(backend, None)
+            # We need to identify the actual neuron configuration class for this serialized key
+            if _serialized_key is None:
+                raise ValueError("Neuron configuration is invalid: unable to identify the serialized key")
+            cls = _NEURON_CONFIG_FOR_KEY.get(_serialized_key, None)
             if cls is None:
-                raise ValueError(f"No neuron configuration registered for the {backend} backend")
+                raise ValueError(f"No neuron configuration registered for the {_serialized_key} serialized key")
         else:
-            assert _KEY_FOR_NEURON_CONFIG[cls] == backend
+            assert _KEY_FOR_NEURON_CONFIG[cls] == _serialized_key
         config = cls(**{**config_dict})
         logger.info(f"Neuron config {config}")
         return config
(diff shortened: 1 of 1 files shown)


optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -0,0 +1,958 @@
...
+                padding_mask = causal_mask[:, :, :, :mask_length] + attention_mask[:, None, None, :]
+                padding_mask = padding_mask == 0
+                causal_mask[:, :, :, :mask_length] = causal_mask[:, :, :, :mask_length].masked_fill(
+                    padding_mask, min_dtype
+                )
+
+        return causal_mask
+
+
+class GraniteForCausalLM(GranitePreTrainedModel, GenerationMixin):
```
dacorvo, line 848: I don't think it is useful to inherit from GenerationMixin, since I doubt the generate method will actually work.

optimum/neuron/accelerate/accelerator.py
```diff
@@ -218,8 +219,8 @@ def prepare_data_loader(
                     "cause failure if the last batch size is not divisible by the number of microbatches for the pipeline."
                 )
         else:
-            num_replicas = xm.xrt_world_size()
-            rank = xm.get_ordinal()
+            num_replicas = xr.world_size()
+            rank = xr.global_ordinal()
```
michaelbenayoun, line 223: Thanks!

optimum/neuron/accelerate/state.py
```diff
@@ -87,8 +88,8 @@ def __init__(self, cpu: bool = False, **kwargs):
                     set_neuron_cc_flags_for_torch_amp()
                 if not torch.distributed.is_initialized():
                     init_process_group()
-                self.num_processes = xm.xrt_world_size()
-                self.process_index = xm.get_ordinal()
+                self.num_processes = xr.world_size()
+                self.process_index = xr.global_ordinal()
```
michaelbenayoun, line 92: Thanks!

Could you check that `xm.xrt_world_size` and `xm.get_ordinal` are not present anywhere else in the code please?

optimum/neuron/models/training/granite/configuration_granite.py
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Granite model configuration extended for Neuron"""
+
+from transformers.models.granite import GraniteConfig
+
+
+class NeuronGraniteConfig(GraniteConfig):
```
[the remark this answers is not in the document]
michaelbenayoun: In the PR I am working on to enable custom modeling, I just attach the `mp_config` which already existed before.
You can check it [here](https://github.com/huggingface/optimum-neuron/pull/801/files#diff-7fda1f4deb9a79e4201376095bd5eabc0650da1fb05672493d7949af0820be18R600) and [here](https://github.com/huggingface/optimum-neuron/pull/801/files#diff-b2d1c406484e1ef14ae392ed9db6136bdb27c2354951c4ac6ec7155b7b38a2b0R58).

optimum/neuron/models/training/granite/modeling_granite.py
michaelbenayoun, line 1: For weight fusing / QGAQKV etc I have this API: https://github.com/huggingface/optimum-neuron/pull/801/files#diff-7fda1f4deb9a79e4201376095bd5eabc0650da1fb05672493d7949af0820be18R101-R107

WDYT?

Basically the idea is to have a common API so that we can share the code for splitting the loaded weights in `from_pretrained` and consolidating the checkpoints afterwards.

optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -0,0 +1,958 @@
...
+            sin = emb.sin()
+
+        # Advanced RoPE types (e.g. yarn) apply a post-processing scaling factor, equivalent to scaling attention
+        cos = cos * self.attention_scaling
+        sin = sin * self.attention_scaling
+
+        return cos.to(dtype=x.dtype), sin.to(dtype=x.dtype)
+
+
+GRANITE_START_DOCSTRING = r"""
```
michaelbenayoun, line 540: In my opinion we should not put all the documentation. It adds a lot of boilerplate and people already know what to expect or we can point them to Transformers.`

optimum/neuron/accelerate/state.py
tengomucho: done, I checked

optimum/neuron/models/training/granite/modeling_granite.py
tengomucho: I agree it would be good to have a common API for these kind of things, and in general even for layers. I prefer having simpler functions though. I propose to isolate this, and then we can make the API uniform once your PR is ready.

optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -0,0 +1,958 @@
...
+        return self.lm_head
+
+    def set_output_embeddings(self, new_embeddings):
+        self.lm_head = new_embeddings
+
+    def set_decoder(self, decoder):
+        self.model = decoder
+
+    def get_decoder(self):
+        return self.model
```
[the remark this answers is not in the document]
tengomucho: I will only leave get_output_embeddings, the only one used.

michaelbenayoun (approved):
LGTM!

Let's address the remaining points once #801 is merged as agreed offline.

Thanks @tengomucho !!

dacorvo (approved):
LGTM, thanks !

PR #834, opened 2025-04-11
ci: add cv2 to workaround transformers spurious import

dacorvo:
# What does this PR do?

This fixes an error due to a missing `cv2` library when importing transformers that prevents the cache workflow from running.


JingyaHuang (approved):
LGTM, btw I think perhaps we don't need this anymore by bumping transformers to 4.50.0.

Context: https://github.com/huggingface/optimum-neuron/pull/789#discussion_r2001204108

tengomucho (approved):

PR #836, opened 2025-04-15
Cache Hub API Changes

tengomucho:
# What does this PR do?

Drop the registry usage from training, given that we did not identify cache registry usage for training. This allows to simplify the API for hub handling, that requires less parameters.
Also, the "inference" subfolder is removed from the registry tree.

### DIFF
--- a/optimum/commands/neuron/cache.py
+++ b/optimum/commands/neuron/cache.py
@@ -191,21 +191,12 @@ class LookupRepoCommand(BaseOptimumCLICommand):
             default=None,
             help="The optional task to lookup cached versions for models supporting multiple tasks.",
         )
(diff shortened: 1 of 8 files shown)


michaelbenayoun (commented):
Overall looks good to me, thanks for cleaning this!

QQ: We do not push to the registry under "inference" when doing training right?

tengomucho:
> QQ: We do not push to the registry under "inference" when doing training right?

That is right, training is not pushing to the registry anymore, it will only use the graph cache.

michaelbenayoun (approved):

dacorvo (approved):
LGTM, thanks !

PR #839, opened 2025-04-25
Initial PR for peft

michaelbenayoun:
# What does this PR do?

- [x] LoRA on linears / parallel linears works
- [x] LoraEmbedding works
- [x] LoRA on GQAQKVColumnParallelLinear works
- [x] `NeuronPeftModel` class
- [x] `from_pretrained`
- [x] `save_pretrained`
- [x] Load adapters via`NeuronModel.from_pretrained` ? **We decide not to support it for now.**
- [x] Checkpoints consolidation
- [x] Test gradient checkpointing
- [x] Feature to update the target modules with the transformation specs


optimum/neuron/peft/peft_model.py
michaelbenayoun, line 1: To remove, not relevant.

optimum/neuron/models/training/modeling_utils.py
```diff
@@ -856,6 +856,12 @@ def from_pretrained(
         else:
             _adapter_model_path = None
 
+        if _adapter_model_path is not None:
+            raise NotSupportedError(
+                f"Loading adapters directly from {cls.__name__}.from_pretrained is not supported. "
+                "Please use the NeuronPeftModelForXXX classes to load adapters."
+            )
```
michaelbenayoun: There is subsequent reference of `_adapter_model_path` in the code. Should I remove it or keep it in case it is added later? I know we tend to remove code like that so I'm asking.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -287,7 +443,81 @@ def adapt_state_dict(
...
+        for param_name in param_names:
+            lora_A_weight_names = [
+                f"{module_fully_qualified_name}.{name}.lora_A.{param_name}" for name in self.linear_names
+            ]
+
+            # Question: how do we "mix" the multiple LoRA A weights?
+            lora_A_weight = torch.mean(
+                torch.stack([state_dict.pop(name) for name in lora_A_weight_names], dim=0),
+                dim=0,
+            )
```
michaelbenayoun: In the case where we load lora weights, there is a choice to be made because we will go from having one `lora_A` linear per linear layers, to only one `lora_A` linear for the fused linear layer. So how do we "mix" these `lora_A` weights for the fused layer? Here I decide to do the mean.

optimum/neuron/utils/misc.py
```diff
@@ -640,3 +640,12 @@ def map_torch_dtype(dtype: Union[str, torch.dtype]):
         dtype = dtype_mapping.get(dtype)
 
     return dtype
+
+
+def is_custom_modeling_model(model) -> bool:
```
tengomucho: I would move this to a dedicated training_utils module instead of putting it in a generic "misc" one.

optimum/neuron/distributed/checkpointing.py
```diff
@@ -309,49 +317,66 @@ def consolidate_model_parallel_checkpoints_to_unified_checkpoint(
...
+        state_dict_split = split_torch_state_dict_into_shards(
+            state_dict, filename_pattern=safe_weights_name if save_format == "safetensors" else weights_name
+        )
+        # Save index if sharded
+        if state_dict_split.is_sharded:
+            index = {
+                "metadata": state_dict_split.metadata,
+                "weight_map": state_dict_split.tensor_to_filename,
+            }
+            save_index_file = SAFE_WEIGHTS_INDEX_NAME if save_format == "safetensors" else WEIGHTS_INDEX_NAME
```
tengomucho, line 368: could we just support only the safetensors format to reduce complexity?

optimum/neuron/models/training/modeling_utils.py
tengomucho: yes, I think we can remove it, if it is possible to load it with `NeuronPeftModelForXXX` I do not see the need to keep the code here too.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -93,11 +166,47 @@ def adapt_state_dict(
...
+    @abstractmethod
+    def _lora_to_original_weights(
+        self,
+        module_fully_qualified_name: str,
+        sharded_state_dicts: Dict[str, list[torch.Tensor]],
+        parameters_metadata: Dict[str, Dict[str, Any]],
+    ) -> tuple[Dict[str, torch.Tensor], list[str]]:
         pass
 
     @abstractmethod
```
tengomucho: this is not abstract anymore I think

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -240,7 +358,45 @@ def __post_init__(self):
...
+        fused_linear_qualified_name = f"{module_fully_qualified_name}.{self.fused_linear_name}"
+        fused_linear = model.get_submodule(fused_linear_qualified_name)
+        if isinstance(fused_linear, ParallelLinear):
+            return "lora"
+        return None
+
+    def adapt_peft_config(self, peft_config: PeftConfig, inplace: bool = False) -> PeftConfig:
+        if not inplace:
+            peft_config = copy.deepcopy(peft_config)
+        if peft_config.peft_type == "LORA":
```
tengomucho, line 381: Wouldn't it be better to add an "else" for other types?

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -240,7 +358,45 @@ def __post_init__(self):
...
+                )
+            if all_linears_in_target_modules:
+                for name in self.linear_names:
+                    target_modules.remove(name)
+                target_modules.add(self.fused_linear_name)
+        return peft_config
+
+    def to_original_peft_config(self, peft_config: PeftConfig, inplace: bool = False) -> PeftConfig:
+        peft_config = copy.deepcopy(peft_config) if not inplace else peft_config
+        if peft_config.peft_type == "LORA":
```
tengomucho, line 405: same here

optimum/neuron/models/training/transformations_utils.py
tengomucho: Couldn't you stack them instead? This will produce an intermediary result that is 3x bigger then the original one, and then you can do the same with the LoRA B weights. I am not sure if this is better, but results should be closer to the original, shouldn't it?
Otherwise, I think at least a warning should be displayed.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -287,7 +443,81 @@ def adapt_state_dict(
                 )
         return state_dict
 
-    def to_original_weights(
+    def _lora_adapt_state_dict(
```
tengomucho, line 464: LoRA specific methods are huge. I wonder if we would benefit from some refactoring to isolate them in a separated class.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -240,7 +358,45 @@ def __post_init__(self):
...
+        return None
+
+    def adapt_peft_config(self, peft_config: PeftConfig, inplace: bool = False) -> PeftConfig:
+        if not inplace:
+            peft_config = copy.deepcopy(peft_config)
+        if peft_config.peft_type == "LORA":
+            target_modules = peft_config.target_modules
+            at_least_one_linear_in_target_modules = any(name in target_modules for name in self.linear_names)
+            all_linears_in_target_modules = all(name in target_modules for name in self.linear_names)
+            if at_least_one_linear_in_target_modules and not all_linears_in_target_modules:
```
tengomucho, line 387: Perhaps you can check if target_modules is "all-linear" to avoid this check

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -476,7 +827,48 @@ def create_gqa_query_or_output_projection_weight_from_full_weight(
 
         return full_weight
 
-    def adapt_state_dict(
+    def guess_peft_type(self, model: torch.nn.Module, module_fully_qualified_name: str) -> Optional[str]:
+        from ...peft.tuners.lora.layer import GQAQKVColumnParallelLinear
```
tengomucho, line 850: can you move this to the top of the file?

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -476,7 +827,48 @@ def create_gqa_query_or_output_projection_weight_from_full_weight(
...
+        gqa_qkv_projection_qualified_name = f"{module_fully_qualified_name}.{self.gqa_qkv_projection_name}"
+        qkv_linear = model.get_submodule(gqa_qkv_projection_qualified_name)
+        if isinstance(qkv_linear, GQAQKVColumnParallelLinear):
+            return "lora"
+        return None
+
+    def adapt_peft_config(self, peft_config: PeftConfig, inplace: bool = False) -> PeftConfig:
+        if not inplace:
+            peft_config = copy.deepcopy(peft_config)
+        if peft_config.peft_type == "LORA":
```
tengomucho, line 861: same here: 
- consider adding an "else" for the `peft_type`
- consider checking `all-linear` in the `target_modules`

optimum/neuron/peft/tuners/lora/layer.py
```diff
@@ -0,0 +1,504 @@
...
+        self.lora_alpha[adapter_name] = lora_alpha
+        if lora_dropout > 0.0:
+            lora_dropout_layer = nn.Dropout(p=lora_dropout)
+        else:
+            lora_dropout_layer = nn.Identity()
+
+        self.lora_dropout.update(nn.ModuleDict({adapter_name: lora_dropout_layer}))
+        # Actual trainable parameters
+        # There are two cases:
+        #   1. The base linear layer is a RowParallelLinear, then:
```
tengomucho, line 148: If I get it right, you choose the layers substitutions based on the base_layer. This might make sense for the `RowParallelLinear` when the `input_is_parallel` might be set to `True`, but I wonder if
- it would make sense just to gather the input if it is parallel, and always use a Linear, because these linears should have rather small weights
- it would make sense instead to keep the parallelization, but even use `ParallelLinears` on both cases.

optimum/neuron/peft/tuners/lora/layer.py
```diff
@@ -0,0 +1,504 @@
...
+            with gather_params_ctx(self.get_base_layer().weight):
+                self.pissa_init(adapter_name, init_lora_weights)
+        elif isinstance(init_lora_weights, str) and init_lora_weights.lower() == "olora":
+            with gather_params_ctx(self.get_base_layer().weight):
+                self.olora_init(adapter_name)
+        elif init_lora_weights == "loftq":
+            with gather_params_ctx(self.get_base_layer().weight):
+                self.loftq_init(adapter_name)
+        elif init_lora_weights == "eva":
+            nn.init.zeros_(self.lora_B[adapter_name].weight)
```
tengomucho, line 193: I trust you on these 🤯

tests/distributed/test_train_overfit.py
```diff
@@ -196,3 +202,42 @@ def test_overfit_causal_lm(
...
+    ],
+    ids=[
+        "dp=1,tp=8,pp=1",
+        # This is to test the case where we have more than 8 TP workers, which will use GQAGQAColumnParallelLinear.
+        "dp=1,tp=32,pp=1",
+    ],
+)
+def test_overfit_lora_causal_lm(world_size, tp_size, pp_size, tmpdir):
+    peft_config = LoraConfig(
+        r=16,
```
tengomucho, line 221: is there a restriction on the r size? (perhaps yes to be correctly sharded)

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Done.

optimum/neuron/utils/misc.py
michaelbenayoun: Done

optimum/neuron/distributed/checkpointing.py
michaelbenayoun: I think it is good to keep both, and it does not make up the complexity in this code IMO, the rest is.

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: Good catch!

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: What do you mean?
No other type is supported for now.

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: I think it is okay as it is, with seperate methods and a main method that dispatches, but we could also imagine a subclass for every time. I dont know how much code save it would provide though.

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: I do not cover this case, fixing it!

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: From memory, I have done that to avoid circular imports.

optimum/neuron/models/training/transformations_utils.py
michaelbenayoun: Addressed for `all-linear` but still not getting what you mean for "else"

optimum/neuron/peft/tuners/lora/layer.py
michaelbenayoun: - `input_is_parallel` is always true otherwise it behaves as a regular linear.
- We could try that, but would change a lot in the PR. Also it will be slower for sure, while not bringing a big benefit since would still need to parallelize for `ColumnParallelLinear`s
- What do you mean?

optimum/neuron/peft/tuners/lora/layer.py
michaelbenayoun: Actually I have not tested these, I can investigate or do it in another PR.

tests/distributed/test_train_overfit.py
michaelbenayoun: Yes there is, it needs to be divisible by the TP size.

optimum/neuron/models/training/transformations_utils.py
tengomucho: I mean you can just raise an exception is the pet_type is something else

optimum/neuron/models/training/transformations_utils.py
tengomucho: Ok let's leave it like this for now then!

optimum/neuron/models/training/transformations_utils.py
tengomucho: Add a comment maybe then

optimum/neuron/models/training/transformations_utils.py
tengomucho: OK you can do it in a different way, something like adding
```python
        if peft_config.peft_type != "LORA":
            raise ValueError("Only peft_type 'LORA' is supported")
```
and then remove the if.

tests/distributed/test_train_overfit.py
tengomucho: Is it possible to raise a clear error when it is not the case, so to guide the user?

michaelbenayoun: Actually I was wrong, there is no constraint!

optimum/neuron/utils/__init__.py
```diff
@@ -112,6 +115,7 @@
         ENCODER_NAME,
         NEURON_FILE_NAME,
     )
+    from .errors import NotSupportedError
```
tengomucho: to be honest, I'd rather remove this one and use `NotImplementedError` instead.

tengomucho (approved):
Just a nit, otherwise LGTM

optimum/neuron/utils/__init__.py
michaelbenayoun: Done

PR #844, opened 2025-05-13
training dlc doc

pagezyhf:
I was checking if we were up to date on the doc and we missed this one:
https://github.com/aws/deep-learning-containers/releases/tag/v2.2-hf-4.48.1-pt-2.1.2-tr-neuronx-sdk2.20.0-py310


tengomucho (approved):
Well spotted, LGTM!

PR #845, opened 2025-05-15
🪨 Granite Training

tengomucho:
# What does this PR do?

This PR modifies Granite modeling code to use the recently added `NeuronModelMixin`, and to define the modeling code in a similar way as it is done in transformers when doing modular modeling code, so its definition is based on Llama.
This allows to quickly reuse the code that allows training to work. The overfit test is used to validate that training works. Currently, the model allows training to be done with parallel sequence enabled or disabled, but only with eager attention implementation, because the flash attention loss is returning NaN. This issue will be addressed in a follow-up PR.


optimum/neuron/accelerate/utils/dataclasses.py
```diff
@@ -21,6 +21,7 @@
 
 import torch
 
+from ...configuration_utils import NeuronConfig, register_neuron_config
```
dacorvo, line 24: I am not so sure about this change: There is nothing in the name of the config that allows to identify to what it actually relates. I think the previous name was slightly better (although I would have preferred that it refers to NxD somehow).

tests/distributed/test_custom_modeling.py
```diff
@@ -44,10 +48,14 @@ def _test_parallel_granite():
...
-    mp_config = TrainingNeuronConfig(
+    tp_size = get_tensor_model_parallel_size()
+    mp_config = ModelParallelismConfig(
+        tensor_parallel_size=tp_size,
         sequence_parallel_enabled=False,
+        use_flash_attention=False,
     )
-    model = GraniteForCausalLM.from_pretrained(model_id, mp_config, torch_dtype=torch_dtype).to(device="xla")
+    model = GraniteForCausalLM.from_pretrained(model_id, mp_config, torch_dtype=torch_dtype)
+    move_model_to_device(model, xm.xla_device())
```
dacorvo, line 58: Why is this required now ?

tests/distributed/test_model_parallelization.py
```diff
@@ -650,7 +650,7 @@ def on_log(self, args, state, control, logs=None, **kwargs):
         if last_loss is None:
             raise ValueError("No loss found in the logs.")
         print("Last loss", last_loss)
-        assert last_loss != 0.0, "The model did not overfit the dataset."
+        assert last_loss == 0.0, "The model did not overfit the dataset."
```
dacorvo: So does it mean that the CI was failing when the previous pull-request was merged ?

optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -394,162 +174,85 @@ def forward(
...
-        causal_mask = self._update_causal_mask(attention_mask, inputs_embeds, cache_position)
+        if self.mp_config.recompute_causal_mask:
+            causal_mask = None
+        else:
+            causal_mask = self._update_causal_mask(attention_mask, inputs_embeds, cache_position)
 
         hidden_states = inputs_embeds
-
+        # print(f"🟡 hidden_states {hidden_states.shape}")
+        # raise ValueError("hidden_states")
```
dacorvo, line 255: Remove these two lines

optimum/neuron/models/training/__init__.py
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from .granite.modeling_granite import GraniteForCausalLM
```
dacorvo, line 16: I don't see how it relates to transformers modeling code: AFAIK in transformers modeling classes are not shared between models but rather rewritten by tools using shared templates.

dacorvo (changes requested):

tests/distributed/test_custom_modeling.py
tengomucho: I have seen the `move_model_to_device` is the API used in NxD, it does the same with a context that reties shared weights after exiting the context, and preserves tensor and sequence parallelism when moving it. 
The previous API seems to work fine for this test, but I preferred to use the other one to keep it coherent with the way we do it elsewhere for training.

tests/distributed/test_model_parallelization.py
tengomucho: This function was not even run by the CI, I figured this was wrong when running it myself 😭

optimum/neuron/models/training/__init__.py
tengomucho: You are right, I thought it was the approach they were using, but it is not, I am just using an inheritance system. We could support Transformers' approach, but not in this PR.

docs/source/guides/distributed_training.mdx
```diff
@@ -183,11 +183,11 @@ Just as for ZeRO-1, it is possible to wrap the optimizer class to make it lazy.
 ```python
 from torch.optim import AdamW
 from optimum.neuron import NeuronAccelerator
-from optimum.neuron.accelerate.utils import ModelParallelismConfig
+from optimum.neuron.models.training.config import TrainingNeuronConfig
 from optimum.neuron.distributed import lazy_load_for_parallelism
 
 tensor_parallel_size = 8
-mp_config = ModelParallelismConfig(
+mp_config = TrainingNeuronConfig(
```
michaelbenayoun, line 190: Then I would call it `DistributedTrainingNeuronConfig`, but maybe it's too long.

optimum/neuron/accelerate/utils/dataclasses.py
```diff
@@ -15,23 +15,17 @@
...
 
 
 if is_neuronx_distributed_available():
-    from neuronx_distributed.parallel_layers import parallel_state
+    pass
 
 
 if TYPE_CHECKING:
-    from transformers import PreTrainedModel
+    pass
```
michaelbenayoun, line 28: Can be removed.

optimum/neuron/models/training/config.py
```diff
@@ -1,51 +1,90 @@
...
+    gradient_checkpointing: bool = False
+    checkpoint_dir: Optional[Union[str, Path]] = None
+    num_local_ranks_per_step: int = 8
+    use_xser: bool = True
+    async_save: bool = False
+    fuse_qkv: bool = False
+    use_flash_attention: bool = True
+    recompute_causal_mask: bool = True
+
+    def __post_init__(self):
```
michaelbenayoun, line 40: Does it still have a `__post_init__` method in this context of inheritance?

optimum/neuron/models/training/config.py
```diff
@@ -1,51 +1,90 @@
...
+            parallelize_embeddings=self.parallelize_embeddings,
+            sequence_parallel_enabled=self.sequence_parallel_enabled,
+            kv_size_multiplier=self.kv_size_multiplier,
+            pipeline_parallel_num_microbatches=self.pipeline_parallel_num_microbatches,
+            pipeline_parallel_use_zero1_optimizer=self.pipeline_parallel_use_zero1_optimizer,
+            pipeline_parallel_gradient_checkpointing_enabled=self.gradient_checkpointing,
+            checkpoint_dir=self.checkpoint_dir,
+            num_local_ranks_per_step=self.num_local_ranks_per_step,
+        )
+        return parallelized_model
```
michaelbenayoun, line 90: Once we removed the former approach we should be able to remove this.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -468,8 +468,10 @@ def forward(
         cos, sin = position_embeddings
         query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
 
-        if self.config._attn_implementation == "flash_attention_2":
-            attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]
+        # We allow the user to specify the attention implementation in the config, but also use the
+        # mp_config.use_flash_attention flag.
+        if self.config._attn_implementation == "flash_attention_2" or self.mp_config.use_flash_attention:
+            attention_interface = ALL_ATTENTION_FUNCTIONS["flash_attention_2"]
```
michaelbenayoun, line 474: nit: I'm more in favor to keep the `self.config._attn_implementation` only and then actually set this attribute in the `from_pretrained` method using `self.mp_config`.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -88,6 +88,7 @@ def adapt_state_dict(
         named_parameters: Dict[str, torch.nn.Parameter],
         orig_state_dict: Dict[str, torch.Tensor],
         inplace: bool = False,
+        upstanding_sharded_params: Dict[str, torch.Tensor] = {},
```
michaelbenayoun, line 91:
```suggestion
        upstanding_sharded_params: Optional[Dict[str, torch.Tensor]] = None,
```

Then do:
```
if upstanding_sharded_params is None:
    upstanding_sharded_params = {}
```

Because mutable defaults can be problematic, especially when there are multiple calls of the function.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -463,6 +482,7 @@ def adapt_state_dict(
...
```
michaelbenayoun, line 485: Same comment.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -634,7 +654,10 @@ def set_module_names_in_transformation_specs(model: torch.nn.Module):
 
 
 def adapt_state_dict(
-    model: torch.nn.Module, state_dict: Dict[str, torch.Tensor], inplace: bool = False
+    model: torch.nn.Module,
+    state_dict: Dict[str, torch.Tensor],
+    inplace: bool = False,
+    upstanding_sharded_params: Dict[str, torch.Tensor] = {},
```
michaelbenayoun, line 660: Same comment.

optimum/neuron/models/training/transformations_utils.py
tengomucho: I will make this mandatory then.

optimum/neuron/models/training/transformations_utils.py
tengomucho: see above

optimum/neuron/models/training/llama/modeling_llama.py
tengomucho: I am not sure keeping the info on both the `config` and `mp_config` is a good idea, and I think a more radical refactoring would be a better solution. I could remove it from `mp_config`, what do you think?

tengomucho: I am not sure keeping the info on both the `config` and `mp_config` is a good idea, and I think a more radical refactoring would be a better solution. I could remove it from `mp_config`, what do you think?

tengomucho: After discussions, it seems convenient to keep it in the `training_args` that generate the `mp_config` that is then used in the `from_pretrained`. So I will do this change in this direction, making sure the parameters are coherent along the way.

tests/distributed/test_model_parallelization.py
michaelbenayoun: Good catch!

optimum/neuron/models/training/llama/modeling_llama.py
tengomucho: After further discussions, `use_flash_attention` is going to be removed from `mp_config`. If flash attention is to be enabled, it is possible to do it either using the `use_flash_attention_2` flag when creating the model and calling `from_pretrained`, or by setting the implementation in the config.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -647,7 +670,12 @@ def adapt_state_dict(
             continue
         # If a submodule is a CustomModule, it has transformation specs and we use them to transform the associated
         # weights from the original state dict.
-        state_dict = module.specs.adapt_state_dict(named_parameters, state_dict, inplace=inplace)
+        state_dict = module.specs.adapt_state_dict(
+            named_parameters,
+            state_dict,
+            upstanding_sharded_params=upstanding_sharded_params,
```
michaelbenayoun, line 676:
```suggestion
            upstanding_sharded_params,
```

optimum/neuron/training_args.py
```diff
@@ -133,12 +134,6 @@ class NeuronTrainingArgumentsMixin:
             ),
         },
     )
-    use_flash_attention: bool = field(
-        default=True,
-        metadata={
-            "help": "Whether to use the flash attention kernel for self-attention layers.",
-        },
-    )
```
michaelbenayoun, line 141: No we should keep this one no?

optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -415,162 +173,81 @@ def forward(
...
             [GraniteDecoderLayer(config, mp_config, layer_idx) for layer_idx in range(config.num_hidden_layers)]
         )
-        self.norm = GraniteRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
-        self.rotary_emb = GraniteRotaryEmbedding(config=config)
-        self.gradient_checkpointing = False
-        self.embedding_multiplier = config.embedding_multiplier
+        self.norm = LlamaRMSNorm(
+            config.hidden_size, eps=config.rms_norm_eps, sequence_parallel_enabled=mp_config.sequence_parallel_enabled
+        )
+        self.rotary_emb = LlamaRotaryEmbedding(config=config)
```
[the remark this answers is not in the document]
tengomucho: They are the same, so I did not implement Granite version for these.

optimum/neuron/models/training/transformations_utils.py
```diff
@@ -170,7 +172,11 @@ def adapt_state_dict(
             if not isinstance(spec, ModelWeightTransformationSpec):
                 raise TypeError(f"spec must be of type ModelWeightTransformationSpec, but got {type(spec)}")
             orig_state_dict = spec.adapt_state_dict(
-                self.module_fully_qualified_name, named_parameters, orig_state_dict, inplace=inplace
+                self.module_fully_qualified_name,
+                named_parameters,
+                orig_state_dict,
+                upstanding_sharded_params=upstanding_sharded_params,
```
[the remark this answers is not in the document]
tengomucho: This is the same

optimum/neuron/models/training/transformations_utils.py
tengomucho: same

optimum/neuron/training_args.py
tengomucho: I do not see why we should keep this. The flash attention usage does not need to be in the args. To enable it there are two ways:
1. by setting the `config._attn_implementation = "flash_attention2"`
2. by passing `use_flash_attention_2=True` when calling `from_pretrained`. This will end up setting the config above.
I do not see why we would need to set these in the training arguments.

optimum/neuron/models/training/granite/modeling_granite.py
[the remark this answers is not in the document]
michaelbenayoun: Alright!

optimum/neuron/models/training/transformations_utils.py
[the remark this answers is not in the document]
michaelbenayoun: Yes, it's just nitpicking to match what is done for the other arguments.

optimum/neuron/training_args.py
michaelbenayoun: To be able to enable or disable the use of flash attention easily in our training scripts instead of going to the python code.

PR #848, opened 2025-05-22
Activate continuous batching for Llama on NxD

dacorvo:
# What does this PR do?

This pull-request fixes a few issues related to continuous batching and activates it by default for the Llama models as soon as the batch size is > 1.

The reason why continuous batching is always set is because with that flag, the prefill (context encoding) always happens with a batch size of 1, which drastically reduces the device memory and allows to use a higher batch size.

This pull-request also aligns some components inspired from NxDI with NxDI 0.2.0. 

Finally, this pull-request adds a test to verify that the assisted decoding with an assistant model works as expected.

Note: while integrating the speculationtest, some issue with the current implementation imported from NxDI were found, leading to suboptimal token matching. This will be adressed in a subsequent pull-request.

### DIFF
--- a/optimum/neuron/models/inference/nxd/backend/modules/generation/generation_utils.py
+++ b/optimum/neuron/models/inference/nxd/backend/modules/generation/generation_utils.py
@@ -281,6 +281,9 @@ class NxDGenerationMixin(GenerationMixin):
         eos_token_id = generation_config.eos_token_id
         assistant_model = candidate_generator.assistant_model
 
+        if assistant_model.neuron_config.on_device_sampling:
+            raise ValueError("Assistant model must not use on-device sampling")
+
         # Initialize the num_assistant_tokens used for speculation.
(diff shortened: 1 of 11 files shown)


optimum/neuron/models/inference/nxd/backend/modules/generation/generation_utils.py
```diff
@@ -186,7 +186,6 @@ def prepare_inputs_for_generation(
             position_ids.masked_fill_(attention_mask == 0, 1)
             if is_decode:
                 position_ids = torch.amax(position_ids, 1, keepdim=True)
-                position_ids = position_ids + 1
```
tengomucho, line 188: oh well spotted!

tengomucho (approved):
LGTM!

michaelbenayoun (approved):
LGTM!

PR #849, opened 2025-05-22
Training granite warning flash attention

michaelbenayoun:
# What does this PR do?

PR #850, opened 2025-05-22
Add Qwen3 modeling for training

tengomucho:
# What does this PR do?

This PR adds the possibility to train Qwen3 modeling code, and it adds a test to show training is possible when using Qwen3 models, such as `Qwen/Qwen3-0.6B`.

Note transformers dependency has been updated to reuse the Qwen3 support that has been included there.


optimum/neuron/models/training/modeling_utils.py
```diff
@@ -1258,7 +1281,7 @@ def from_pretrained(
         config.name_or_path = pretrained_model_name_or_path
 
         # Instantiate model.
-        init_contexts = [no_init_weights(_enable=_fast_init)]
+        init_contexts = [no_init_weights()]
```
michaelbenayoun, line 1284: Why?

optimum/neuron/models/training/qwen3/modeling_qwen3.py
```diff
@@ -0,0 +1,223 @@
...
+    LlamaRotaryEmbedding,
+    apply_rotary_pos_emb,
+    eager_attention_forward,
+    repeat_kv,
+)
+from ..modeling_utils import ALL_ATTENTION_FUNCTIONS
+
+
+if is_torch_xla_available():
+    pass
```
michaelbenayoun, line 45: Can be removed.

optimum/neuron/models/training/qwen3/modeling_qwen3.py
```diff
@@ -0,0 +1,223 @@
...
+
+logger = logging.get_logger(__name__)
+
+
+def _init_normal(std, w):
+    return nn.init.normal_(w, mean=0.0, std=std)
+
+
+class Qwen3Attention(LlamaAttention):
+    def __init__(self, config: Qwen3Config, mp_config: TrainingNeuronConfig, layer_idx: int):
```
michaelbenayoun, line 61: `mp_config` was called liked that for "Model Parallelism config", do you think we should `mp_config -> trn_config` ?

optimum/neuron/models/training/qwen3/modeling_qwen3.py
```diff
@@ -0,0 +1,223 @@
...
+        cos, sin = position_embeddings
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        if self.config._attn_implementation == "flash_attention_2":
+            attention_interface = ALL_ATTENTION_FUNCTIONS["flash_attention_2"]
+            attn_output = attention_interface(
+                query_states,
+                repeat_kv(key_states, self.num_key_value_groups),
+                repeat_kv(value_states, self.num_key_value_groups),
+                dropout_p=0.0 if not self.training else self.attention_dropout,
```
michaelbenayoun, line 108: Be careful, if `self.attention_dropout > 0.0` it will fail (#849)

optimum/neuron/models/training/granite/modeling_granite.py
```diff
@@ -317,6 +316,7 @@ def __init__(self, config, mp_config: TrainingNeuronConfig):
         # Initialize weights and apply final processing
         self.post_init()
 
+    @can_return_tuple
```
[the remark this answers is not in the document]
tengomucho: yes!

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -495,10 +495,10 @@ def forward(
 
         if self.mp_config.sequence_parallel_enabled:
             attn_output = attn_output.permute(2, 0, 1, 3)
-            attn_output = attn_output.reshape(q_len, bsz, self.hidden_size // get_tensor_model_parallel_size())
+            attn_output = attn_output.reshape(q_len, bsz, self.num_heads * self.head_dim)
         else:
             attn_output = attn_output.transpose(1, 2).contiguous()
-            attn_output = attn_output.reshape(bsz, q_len, self.hidden_size // get_tensor_model_parallel_size())
+            attn_output = attn_output.reshape(bsz, q_len, self.num_heads * self.head_dim)
```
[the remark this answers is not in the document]
tengomucho: That is correct. In case of llama that is not a problem, because `hidden_size // head_dim == num_heads`, but I guess it is more correct to reshape explicitly to the dimension that is expected before `o_proj`.

optimum/neuron/models/training/modeling_utils.py
michaelbenayoun: Ok, we might need to check if this is assumed or enabled in the `_load_pretrained_weights` method.

michaelbenayoun (approved):
LGTM!

Awesome to be able to add support for new models this fast, thanks a lot for the contribution @tengomucho !

PR #856, opened 2025-06-02
chore: bump aws neuron sdk version to 2.24.0

JingyaHuang:
# What does this PR do?

as per title.

## Inference

With the latest neuron SDK, we must initialize neuron runtime with NxD if TP (eg. for T5 only the decoder is compiled with TP, we will get a runtime error if we compile the encoder first.). So I added `_reorder_models_and_neuron_configs` to avoid the issue.

### DistributedTest refactor

The update "broke" the `DistributedTest` class and related functions. Since it was something we wanted to improve for a while with @tengomucho, this PR also refactors this. Instead of having a complexe class and a function to perform distributed test, we now have:

- A decorator, that accepts `world_size`, `tp_size` and `pp_size` arguments, but also can "fetch" these values from fixtures / parametrize from Pytest. In then runs the function it decorates in a distributed setting using a custom spawning method compared to standard `subprocess` module from Python. **This method can handle tests hanging gracefully compared to what was happening before, every test will be run until completion but will stop as soon as one worker raises an exception, regardless of the state of the others.**

- A function that replaces `launch_procs` called `run_distributed_test` that uses the same mechanism underneath. The purpose of such function is to be able to choose `world_size`, `tp_size` and `pp_size` values depending on other parameters during the runtime.

- A way to exit distributed "early": with the `EarlyExit` exception, if it raised, the test stops with the `returncode` specified in the exception. By default `returncode=0` which means success. It allows to stop a test as soon as we know it suceeded. A good example of this are the tests in `test_overfit.py`. 

### Update modeling

This release seems to introduce a `transpose_nki_inputs` argument to the flash attention kernel. We update the custom modelings to account for that.

### Enable `model.tie_weights()`

For the custom modelings, we disabled `model.tie_weights()` because of a compiler bug. It seems to be solved so we enable it and add a test for it.

### Granite 

- [x] Flash attention with dropout > 0.0 would produce NaN so it was not supported for common Granite architectures, we implemeted a fix by applying the dropout outside of the flash attention kernel
- [x] Pipeline parallelism with granite produced NaNs before, same, not fixed yet.

### Lighter trainium workflow for PRs 

There are now two ways to trigger the workflow:

- Push on the main branch or on a pull request: it will run all the trainium tests but it will run the overfit tests only on our flagship model (for now it is Llama)
- Cron job: it will run all the trainium tests and will run the overfit tests on all of the models with support


setup.py
```diff
@@ -73,12 +73,12 @@
...
-        "torch-neuronx==2.5.1.2.6.0",
-        "torch==2.5.1.*",
-        "torchvision==0.20.*",
-        "neuronx_distributed==0.11.0",
-        "libneuronxla==2.2.1630.0",
+        "neuronx-cc==2.18.121.0",
+        "torch-neuronx==2.6.0.2.7.5413",
+        "torch==2.6.0.*",
+        "torchvision==0.21.*",
+        "neuronx_distributed==0.12.12111",
```
michaelbenayoun: Can you update the neuronx_distributed version for the training extra please?

tengomucho: this dependency should be also updated in TRAINING_REQUIRES (or removed from there)

tengomucho:
Note there is an error when trying to use `/opt/aws_neuronx_venv_pytorch_2_5/bin/activate`, I think you need to change that too

JingyaHuang:
Ran locally 
* Inference tests
* generation tests
* exporter tests (sd, sdxl ✔️ )
All green ✔️

setup.py
```diff
@@ -24,7 +24,9 @@
 TESTS_REQUIRE = [
     "pytest <= 8.0.0",
     "pytest-rerunfailures",
+    "cloudpickle",
```
dacorvo, line 27: We only need cloudpickle once

dacorvo (changes requested):
One small change is required.

setup.py
michaelbenayoun: Good catch

dacorvo (approved):

PR #857, opened 2025-06-04
Support for PP with custom modeling

michaelbenayoun:
# What does this PR do?

This PR adds support for pipeline-parallelism with custom modeling.

## Features

- [x] `NeuronModelMixin.from_pretrained` can load a model while taking PP into consideration by not loading unnecessary weights in memory
- [x] `NeuronModelMixin.save_pretrained` saves the relevant metadata for consolidation
- [x] `NeuronAccelerator.prepare_model` prepares the model properly for custom modeling + PP
- [x] `NeuronTrainer` can log loss and save the model properly when PP is enabled

## Tests
- [x] Test that checks that only the parameters in the current stage are loaded in memory for each PP rank
- [x] Test that checks that custom model with PP produces the same loss as a regular model (done only for LLama)
- [x] Test that checks overfitting (done for every custom model implementation)
- [x] Test that checks consolidation of sharded checkpoints

### DIFF
--- a/optimum/neuron/models/training/llama/modeling_llama.py
+++ b/optimum/neuron/models/training/llama/modeling_llama.py
@@ -796,6 +796,7 @@ class KwargsForCausalLM(FlashAttentionKwargs, LossKwargs): ...
 class LlamaForCausalLM(NeuronModelMixin, LlamaPreTrainedModel):
     _tied_weights_keys = ["lm_head.weight"]
 
+    SUPPORTS_PIPELINE_PARALLELISM = True
     PIPELINE_TRANSFORMER_LAYER_CLS = LlamaDecoderLayer
     PIPELINE_INPUT_NAMES = ["input_ids", "attention_mask", "labels"]
     PIPELINE_LEAF_MODULE_CLASSE_NAMES = ["LlamaRMSNorm", "LlamaRotaryEmbedding"]
--- a/tests/distributed/test_custom_modeling.py
+++ b/tests/distributed/test_custom_modeling.py
@@ -180,13 +180,19 @@ def _custom_model_matches_original_model(
     # the parallel linears accumulates over the layers.
     torch_dtype=torch.float32,
 ):
-    monkeypatch.setattr(
-        optimum.neuron.models.training.loss_utils, "_PARALLEL_CROSS_ENTROPY_SHOULD_PRESERVE_INPUT", True
-    )
-
     world_size, tp_size, pp_size = parallel_sizes
     dp_size = world_size // (tp_size * pp_size)
     pp_rank = get_pipeline_model_parallel_rank()
+
+    training_mod = importlib.import_module("optimum.neuron.models.training")
+    custom_model_class = getattr(training_mod, model_class_name)
+
+    if pp_size > 1 and not custom_model_class.supports_pipeline_parallelism():
+        pytest.skip(f"The model {model_class_name} does not support pipeline parallelism, skipping the test.")
+
+    monkeypatch.setattr(
+        optimum.neuron.models.training.loss_utils, "_PARALLEL_CROSS_ENTROPY_SHOULD_PRESERVE_INPUT", True
+    )
 
     static_seed_patcher = StaticSeedPatcher(SEED)
 
@@ -242,8 +248,6 @@ def _custom_model_matches_original_model(
         recompute_causal_mask=False,  # Recomputing the causal mask does not impact the loss but it impacts the logits.
     )
 
-    training_mod = importlib.import_module("optimum.neuron.models.training")
-    custom_model_class = getattr(training_mod, model_class_name)
     with static_seed_patcher:
         model = custom_model_class.from_pretrained(
             model_name_or_path, trn_config, attn_implementation=attn_implementation, torch_dtype=torch_dtype
--- a/tests/distributed/test_train_overfit.py
+++ b/tests/distributed/test_train_overfit.py
@@ -15,6 +15,7 @@ from optimum.neuron.utils.import_utils import (
     is_neuronx_distributed_available,
 )
 from optimum.neuron.utils.misc import is_precompilation
+from optimum.neuron.utils.testing_utils import is_trainium_test
 from optimum.neuron.utils.training_utils import is_main_worker_for_metrics
 
 from .. import launch_procs
@@ -39,6 +40,12 @@ def _overfit_causal_lm(
     output_dir,
     peft_config=None,
 ):
+    training_mod = importlib.import_module("optimum.neuron.models.training")
+    model_class = getattr(training_mod, model_class_name)
+
+    if pp_size > 1 and not model_class.supports_pipeline_parallelism():
+        pytest.skip(f"The model {model_class_name} does not support pipeline parallelism, skipping the test.")
+
     # Dataset creation.
     sample_to_overfit = "Paris is the most beautiful city in the world."
     tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
@@ -78,7 +85,7 @@ def _overfit_causal_lm(
         per_device_train_batch_size=1,
         gradient_accumulation_steps=16,
         lr_scheduler_type="cosine",
-        gradient_checkpointing=True,
+        gradient_checkpointing=False,
         max_grad_norm=1,
         bf16=True,
         logging_steps=1,
@@ -90,8 +97,6 @@ def _overfit_causal_lm(
     )
 
     # Model creation.
-    training_mod = importlib.import_module("optimum.neuron.models.training")
-    model_class = getattr(training_mod, model_class_name)
     model = model_class.from_pretrained(
         model_name_or_path,
         training_args.trn_config,
@@ -155,7 +160,7 @@ def _overfit_causal_lm(
         [
             "GraniteForCausalLM",
             "ibm-granite/granite-3.2-2b-instruct",
-            2e-3,
+            1e-4,
             0,
             {},
             # For now we disable flash attention because the default configuration has a dropout for the attention
@@ -187,24 +192,25 @@ def _overfit_causal_lm(
         [32, 2, 4],
     ],
     ids=[
-        "dp=1,tp=8",
+        "dp=4,tp=8",
         "dp=4,tp=2,pp=4",
     ],
 )
 @pytest.mark.neuron_parallel_compile
+@is_trainium_test
 def test_overfit_causal_lm(
     model_class_name,
     model_name_or_path,
     learning_rate,
     warmup_ratio,
     training_kwargs,
+    use_flash_attention_2,
     max_expected_loss,
     max_length,
     num_steps,
     world_size,
     tp_size,
     pp_size,
-    use_flash_attention_2,
     tmpdir,
     set_cache_for_ci,  # This fixture will handle setting the remote cache to make this test faster.
 ):
@@ -239,6 +245,7 @@ def test_overfit_causal_lm(
     ],
 )
 @pytest.mark.neuron_parallel_compile
+@is_trainium_test
 def test_overfit_lora_causal_lm(world_size, tp_size, pp_size, tmpdir, set_cache_for_ci):
     peft_config = LoraConfig(
         r=16,
--- a/optimum/neuron/models/training/granite/modeling_granite.py
+++ b/optimum/neuron/models/training/granite/modeling_granite.py
@@ -53,10 +53,6 @@ if is_neuronx_distributed_available():
 logger = logging.get_logger(__name__)
 
 
-def _init_normal(std, w):
-    return nn.init.normal_(w, mean=0.0, std=std)
-
-
 def rotate_half(x):
     """Rotates half the hidden dims of the input."""
     x1 = x[..., : x.shape[-1] // 2]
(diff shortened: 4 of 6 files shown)


optimum/neuron/models/training/loss_utils.py
```diff
@@ -28,6 +29,7 @@
 _PARALLEL_CROSS_ENTROPY_SHOULD_PRESERVE_INPUT: bool = False
 
 
+@torch.fx.wrap
```
michaelbenayoun, line 32: This is needed since PP uses `torch.fx.symbolic_trace` to create the pipeline stages. It simply tells the symbolic function to consider this function as a leaf (do not trace through it).

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -624,7 +624,7 @@ def get_input_embeddings(self):
     def set_input_embeddings(self, value):
         self.embed_tokens = value
 
-    @can_return_tuple
+    # @can_return_tuple
```
dacorvo: You should probably either uncomment or remove.

.github/workflows/test_neuron_parallel_compile.yml
```diff
@@ -0,0 +1,42 @@
+name: Neuron Parallel Compile Tests
+
+on:
+  workflow_dispatch:
```
tengomucho: perhaps we could run this on release and weekly too, so they the test not need to be manually triggered

.github/workflows/test_neuron_parallel_compile.yml
```diff
@@ -0,0 +1,42 @@
...
+          source aws_neuron_venv_pytorch/bin/activate
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest -m "neuron_parallel_compile" ${{ inputs.test_path }} --collect-only
+      - name: Run neuron_parallel_compile tests
+        run: |
+          source aws_neuron_venv_pytorch/bin/activate
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} neuron_parallel_compile pytest -m "neuron_parallel_compile" ${{ inputs.test_path }} -v --durations=0 -x
+
+      - name: Synchronize with HF cache repo
+        run: |
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} python tests/synchronize_with_cache_repo_for_ci.py
```
dacorvo: Alternatively, you could have called sync in a fixture, like I do for LLM tests (except that you have to wait for the test to complete, so it should happen in teardown code). This way:
- the NEFFs are also cached when you run the tests locally,
- even if the test step fails, all NEFFs for tests that did not have compilation errors are cached.

optimum/neuron/models/training/pipeline_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def get_pipeline_model_parallel_size():
+        return 0
+
+    class NxDPPModel:
+        def __init__(self, *args, **kwargs):
+            pass
+
+
+class MetaParametersOnly:
```
dacorvo, line 55: Are these context manager inspired by other libraries, or are they brand new ? If they come from other libraries we should add pointers.

optimum/neuron/models/training/llama/modeling_llama.py
tengomucho: why is this commented?

dacorvo (commented):
LGTM, only one nit, a comment and a question.

optimum/neuron/models/training/llama/modeling_llama.py
michaelbenayoun: Good catch, it is meant to be uncommented.

.github/workflows/test_neuron_parallel_compile.yml
michaelbenayoun: No I cannot do that. `neuron_parallel_compile` compiles the accumulated graphs after the program it runs has completed so after teardown. We need an extra step.

optimum/neuron/models/training/pipeline_utils.py
michaelbenayoun: Brand new.

optimum/neuron/models/training/pipeline_utils.py
```diff
@@ -0,0 +1,205 @@
...
+from ...distributed.utils import OptimumNeuronFXTracer
+from ...utils.import_utils import is_neuronx_distributed_available, is_torch_xla_available
+from .transformations_utils import get_tensor_model_parallel_attributes
+
+
+if is_torch_xla_available():
+    import torch_xla.core.xla_model as xm
+else:
+    # This is a placeholder for doc building.
+    def xm():
```
tengomucho: Nit: it's a bit weird to define it as a function since it is a module. I would have just done `xm = None`

tests/distributed/test_common.py
```diff
@@ -701,3 +704,34 @@ def test_consolidate_custom_lora_model_parallel_checkpoints(
...
                     torch.testing.assert_close(orig_tensor, consolidated_tensor)
+
+    @pytest.mark.parametrize(
+        "world_size,tp_size,pp_size",
+        [
+            [32, 2, 4],
+        ],
+        ids=["dp=4,tp=2,pp=4"],
+    )
+    def test_each_pp_rank_only_loads_relevant_parameters(self, world_size, tp_size, pp_size):
```
tengomucho, line 715: what about moving PP related tests to a dedicated file, to avoid having to keep track to all of those ? (If you prefer we could do this in another PR though)

tests/distributed/test_custom_modeling.py
```diff
@@ -79,3 +124,216 @@ def test_parallel_granite():
...
+        # For now ignoring past_key_values because they do not match and it is not needed for training.
+        if "past" in output_name:
+            continue
+        if all(output is None for output in outputs):
+            continue
+        if pp_size == 1 or pp_rank == pp_size - 1:
+            _check_output(output_name, outputs[0], outputs[1])
+
+
+CUSTOM_MODELINGS_TO_TEST = [("LlamaForCausalLM", "michaelbenayoun/llama-2-tiny-4kv-heads-4layers-random")]
```
tengomucho: consider moving this to the top

tests/distributed/test_custom_modeling.py
```diff
@@ -1,21 +1,66 @@
...
 import torch
-from transformers import AutoModelForCausalLM, AutoTokenizer
+import torch.utils._pytree as pytree
+import transformers
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
 
+import optimum
+import optimum.neuron.models.training
 from optimum.neuron.models.training.config import TrainingNeuronConfig
-from optimum.neuron.models.training.granite.modeling_granite import GraniteForCausalLM
```
tengomucho, line 5: Did you not break the previously existing test (`test_parallel_granite`) by doing this? I do not mind removing the test if we have something else to test it (e.g.: the overfit one), but I do not want to leave a broken test there.

tests/distributed/test_custom_modeling.py
```diff
@@ -79,3 +124,216 @@ def test_parallel_granite():
...
+    dp_size = world_size // (tp_size * pp_size)
+    pp_rank = get_pipeline_model_parallel_rank()
+
+    static_seed_patcher = StaticSeedPatcher(SEED)
+
+    accelerator = create_accelerator(
+        tp_size, pp_size, parallelize_embeddings=False, sequence_parallel_enabled=sequence_parallel_enabled
+    )
+
+    orig_model_class = getattr(transformers, model_class_name)
```
tengomucho, line 203: this is a custom modeling tests file, but here it seems to me you are instantiating the model from transformers and patching it, the legacy way. Isn't that something we want to move on from?

tests/distributed/test_custom_modeling.py
```diff
@@ -79,3 +124,216 @@ def test_parallel_granite():
...
+            gathered_output = torch.cat(gathered, dim=gather_dim)
+            xm.mark_step()
+            output = gathered_output.to("cpu")
+
+        # In this case, we assume GQAQKVColumnParallelLinear was used, we retrieve only the non-repeated KV heads.
+        if "past" in name and original_output.size(1) != output.size(1):
+            kv_size_multiplier = len(get_kv_shared_group(as_list=True)[0])
+            output = torch.chunk(output, kv_size_multiplier, dim=1)[0]
+
+        torch.testing.assert_close(original_output, output)
```
tengomucho, line 166: you have added an assert_close function, consider using that one.

tests/distributed/test_custom_modeling.py
```diff
@@ -79,3 +124,216 @@ def test_parallel_granite():
...
+    tmpdir = Path(tmpdir)
+    new_model_name_or_path = tmpdir / "my_custom_model"
+    model_class_name, model_name_or_path = model_specs
+
+    config = AutoConfig.from_pretrained(model_name_or_path)
+    tokenizer = AutoTokenizer.from_pretrained(model_name_or_path)
+
+    if qkv_implementation == "fuse_qkv":
+        config.num_key_value_heads = config.num_attention_heads
+        model_class = getattr(transformers, model_class_name)
```
tengomucho, line 321: same here it is instantiated from transformers.

tests/distributed/test_train_overfit.py
```diff
@@ -55,7 +55,17 @@ def gen():
         yield inputs
 
     dataset = datasets.Dataset.from_generator(gen)
-    dataset = dataset.select([0] * 1000)
+    dataset = dataset.select([0] * 10000)
+
+    # Wandb setup.
+    os.environ["WANDB_MODE"] = "online" if not is_precompilation() else "disabled"
+    os.environ["WANDB_PROJECT"] = "test-train-overfit"
```
tengomucho, line 69: do you think we can move this to the `_dist_run` function in `distributed_utils.py`? This will allow to reuse this for other tests too.

tests/distributed/test_train_overfit.py
```diff
@@ -139,39 +150,48 @@ def on_log(self, args, state, control, logs=None, **kwargs):
             True,
             0.0,
             2048,
+            30,
```
tengomucho, line 158: I thought we could actually increase the `max_expected_loss` and reduce this to, say, half, so that tests are shorter. We just want to check the trainer works and loss goes down.

tengomucho (changes requested):
I just want to be clear about the custom modeling / transformers classes patching way of doing things, otherwise LGTM.

tests/distributed/test_common.py
michaelbenayoun: Let's do it in another PR. After this is merged I will work on cleaning up the training part of the library. 
It will contain complete tests restructuring.

.github/workflows/test_neuron_parallel_compile.yml
michaelbenayoun: Done.

optimum/neuron/models/training/pipeline_utils.py
michaelbenayoun: Done.

tests/distributed/test_custom_modeling.py
michaelbenayoun: Done.

tests/distributed/test_custom_modeling.py
michaelbenayoun: It is still there just imported below. I re-moved it. It was not my doing but ruff's I think.

tests/distributed/test_custom_modeling.py
michaelbenayoun: No. I get the original model and run it on Neuron to see if it produces the same thing as the custom model.
The `patch_model_for_neuron` method is just patching general stuff. I will investigate if we still need this at cleanup.

tests/distributed/test_custom_modeling.py
michaelbenayoun: No, the one I have added is because for some reason with the proper precision it was failing. This one is the "official" torch one, and should be preferred whenever possible.

tests/distributed/test_custom_modeling.py
michaelbenayoun: Yes, we create a model that matches the configuration needed for the test. And then proceed on using this model checkpoints for the test.

tests/distributed/test_train_overfit.py
michaelbenayoun: I can but how do I get the proper `WANDB_PROJECT` name in the general case?

tests/distributed/test_train_overfit.py
michaelbenayoun: Yes we can do that as well. But it's quite fast already tbh. The slow part is compilation when needed.

tests/distributed/test_train_overfit.py
tengomucho: you could use somethign like this:

```python
import inspect

...

def _dist_run(params...):
    ....
    # Get test name
    project_name = inspect.stack()[2].function

```

michaelbenayoun: I will keep that in mind for the next PR and will add it there.

PR #860, opened 2025-06-05
Avoid loading weights when exporting an NxD model using the CLI

dacorvo:
# What does this PR do?

When exporting through the CLI, the exported model is immediately saved to disk, so there is no need to load the weights.
Also, it seems that the `model.pt` is much bigger when the weights have been loaded.


tengomucho (approved):
LGTM

PR #866, opened 2025-06-16
Cleanup legacy parallelism support

michaelbenayoun:
# What does this PR do?

- [x] Update GitHub actions
- [ ] Update the docs: another PR ?
- [x] Update the examples
- [x] Clean Trainer
- [x] Clean Accelerator
- [x] Clean tests
- [x] Test overfit for regular model (without custom modeling)

## What's next?

- [ ] Add a custom modeling for T5 and use this implementation for inferentia
- [ ] Remove the `optimum/neuron/disitributed` subpackage once it is done
- [ ] Test that logits match in `test_custom_modeling` even when PP > 1


.github/workflows/test_trainium_common.yml
michaelbenayoun, line 1: It was testing very basic features, and was very flaky.

.github/workflows/test_trainium_examples.yml
michaelbenayoun, line 1: Same, it was testing examples but it was very flaky and buggy. I think it's too much effort for what it represents.

.github/workflows/test_trainium_training.yml
michaelbenayoun, line 1: All the tests related to training are now under `tests/training`.

tests/cli/test_neuron_cache_cli.py
michaelbenayoun, line 1: It was very buggy because it is based on the runners, which themselves are based on fetching examples dynamically from GH. It's just not worth it. We never had issues reported even though the CI was not passing, it is not used.

tests/test_examples.py
michaelbenayoun, line 1: Very buggy, never actually triggered.

.github/workflows/test_trainium_examples.yml
tengomucho: I think later on we should re-add another pipeline to test the scripts we have in the examples, at least to verify they run and do not break.

.github/workflows/test_trainium_training.yml
```diff
@@ -5,15 +5,15 @@ on:
     branches: [ main ]
     paths:
       - "setup.py"
-      - "optimum/neuron/distributed/**.py"
-      - ".github/workflows/test_trainium_distributed.yml"
+      - "optimum/neuron/training/**.py"
```
tengomucho: maybe add also `      - "optimum/neuron/models/training/**.py"`

optimum/neuron/accelerate/utils/misc.py
```diff
@@ -34,6 +33,11 @@
 if TYPE_CHECKING:
     import os
 
+    # Dummy class to avoid import errors in type checking.
+    class NeuronPeftModel:
+        def __init__(self, *args, **kwargs):
```
tengomucho, line 38: why do you need to specify the init?

optimum/neuron/models/training/checkpointing.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+else:
+    PEFT_SAFETENSORS_WEIGHTS_NAME = PEFT_WEIGHTS_NAME = ""
+
+
+@requires_torch_xla
+def xser_load_on_cpu(path: str):
+    """
+    Modified version from neuronx_distributed `_xser_load` function load located at:
+    https://github.com/aws-neuron/neuronx-distributed/blob/main/src/neuronx_distributed/parallel_layers/checkpointing.py#L265-L283.
```
tengomucho: I'd rather point to a given committed version, e.g.:
https://github.com/aws-neuron/neuronx-distributed/blob/e83494557cb4c5b7e185ccf6c9240bfed9a1993d/src/neuronx_distributed/parallel_layers/checkpointing.py#L252

optimum/neuron/models/training/checkpointing.py
```diff
@@ -0,0 +1,240 @@
...
+            rewritten_tensors.append(torch.load(os.path.join(path + ".tensors", "tensor_{}.pt".format(t.tid))))
+        return rewritten_tensors
+
+    def select_fn(v):
+        return type(v) is xser.TensorReference
+
+    return xm.ToXlaTensorArena(convert_fn, select_fn).transform(ref_data)
+
+
+def consolidate_tensor_parallel_checkpoints(
```
tengomucho, line 72: wait what? where di you move this code from?

tests/training/test_custom_modeling.py
```diff
@@ -60,69 +61,12 @@
...
-    launch_procs(
-        _test_parallel_granite,
-        num_procs=2,
-        tp_size=2,
-        pp_size=1,
-    )
+CUSTOM_MODELINGS_TO_TEST = [
+    ("LlamaForCausalLM", "michaelbenayoun/llama-2-tiny-4kv-heads-4layers-random"),
+    ("GraniteForCausalLM", "michaelbenayoun/granite-tiny-4kv-heads-4layers-random"),
+    ("Qwen3ForCausalLM", "michaelbenayoun/qwen3-tiny-4kv-heads-4layers-random"),
```
tengomucho, line 67: why can't we test with the real 0.6B model? I this it's better to test with real models somewhere

tests/training/test_overfit.py
```diff
@@ -144,79 +184,99 @@ def on_log(self, args, state, control, logs=None, **kwargs):
...
             0.04,
             {},
             True,
-            0.0,
+            0.5,  # Use a smaller value when tie_word_embeddings is fixed.
+            2048,
+            50,
+        ],
+        [
+            "LlamaForCausalLM",
```
tengomucho, line 228: maybe we can remove the `Llama-3.2-1B-Instruct` if you add this one

.github/workflows/test_trainium_training.yml
michaelbenayoun: Good catch

optimum/neuron/accelerate/utils/misc.py
michaelbenayoun: It's only for the linter to shut up.

optimum/neuron/models/training/checkpointing.py
michaelbenayoun: It was in `optimum/neuron/distributed/checkpointing.py`

optimum/neuron/models/training/checkpointing.py
michaelbenayoun: Done.

tests/training/test_custom_modeling.py
michaelbenayoun: In the custom modeling, we test tiny version and many of the combinations that we support. It just tests that the forward pass works. Then we test that "real" models (e.g the 0.6B Qwen3 model for instance) can overfit in `test_overvit.py`. 

This way we cover the combinations and actual training (which also validates that things work).

tests/training/test_overfit.py
michaelbenayoun: No it's not the same. 

The first `Llama-3.2-1B-Instruct` test is using the custom modeling implementation and TP.
This one is using `HuggingFaceTB/SmolLM2-135M-Instruct` (much much smaller) that is using the regular transformers implementation and no tp / pp. It is to validate that "regular" models work with the trainer and everything as well.

tengomucho: Oh ok I missed that, that is very cool we test it. Maybe we can replace this and use something with a completely different architecture that is not implemented in  optimum neuron, e.g.: `distilbert/distilgpt2` or `bert` (or just add a test with that).

michaelbenayoun: Yes I used llama because it's a causal lm and the test tests that. I can try with distilgpt2.

tests/training/test_custom_modeling.py
tengomucho: OK, but while this is might be ok for some models, it uses configs that do not put into evidence some challenges faced by the real world models. E.g.: we do not shard the lm_head of granite because of its quirky vocab size. Still, as you said, perhaps it's enough to have tiny models here and real ones only in the overfit test.

tests/training/test_overfit.py
michaelbenayoun: It failed for shape issues, most likely sequence length, using SmolLM as the goal is to simply test the "path" taken for non custom models.

tengomucho: ok

optimum/neuron/models/training/qwen3/modeling_qwen3.py
```diff
@@ -72,7 +72,10 @@ def forward(
         else:
             bsz, q_len, _ = hidden_states.size()
 
-        if self.qkv_linear:
+        if self.trn_config.fuse_qkv and self.num_heads == self.num_key_value_heads and self.kv_size_multiplier == 1:
+            qkv_states = self.qkv_proj(hidden_states)
+            query_states, key_states, value_states = qkv_states.split(self.split_size, dim=2)
+        elif self.qkv_linear:
```
tengomucho, line 78: good catch

tests/training/test_custom_modeling.py
```diff
@@ -233,8 +233,7 @@ def _custom_model_matches_original_model(
 
 @pytest.mark.parametrize("qkv_implementation", ["regular_qkv", "fuse_qkv", "qkv_linear"])
 @pytest.mark.parametrize("model_specs", CUSTOM_MODELINGS_TO_TEST, ids=[specs[0] for specs in CUSTOM_MODELINGS_TO_TEST])
-@pytest.mark.timeout(300)  # If a test takes more that 5 minutes, it is likely stuck.
-@pytest.mark.flaky(reruns=3, reruns_delay=5)
+@pytest.mark.flaky(reruns=5, reruns_delay=5)
```
tengomucho, line 236: nice one

PR #867, opened 2025-06-16
Add support for phi3 models using the nxd backend

dacorvo:
# What does this PR do?

This modifies the custom modeling for the inference of `phi3` models to use the NxD backend instead of the legacy HLO backend.


tengomucho (approved):
LGTM

PR #872, opened 2025-06-23
Improve diffusers cache CIs

JingyaHuang:
# What does this PR do?

Divide the compilations of Diffusers ckpts to multiple jobs to avoid being canceled. Configs could be found here:https://huggingface.co/aws-neuron/optimum-neuron-cache/tree/main/inference-cache-config.


dacorvo (approved):
LGTM, thanks !

PR #876, opened 2025-06-24
Refactoring T5 implementation for NxD support

JingyaHuang:
# What does this PR do?

As per title, with the PR, we are not relied on the training Parallelizer anymore.

- [x] Test with TP=8 manually.

### DIFF
--- a/optimum/exporters/neuron/model_configs.py
+++ b/optimum/exporters/neuron/model_configs.py
@@ -968,7 +968,7 @@ class T5EncoderForTransformersNeuronConfig(T5EncoderBaseNeuronConfig):
         model.config.use_cache = True
         with saved_model_in_temporary_directory(model) as ckpt_path:
             # Plug in parallel layers
-            from optimum.neuron.models.inference.nxd.t5.modeling_t5 import parallelize
+            from optimum.neuron.models.inference.t5.modeling_t5 import parallelize
 
             parallel_model = parallelize(model)
             # Load the weights into the parallel layers
@@ -1115,7 +1115,7 @@ class T5DecoderNeuronConfig(TextSeq2SeqNeuronConfig):
         model.config.use_cache = True
         with saved_model_in_temporary_directory(model) as ckpt_path:
             # Plug in parallel layers
-            from optimum.neuron.models.inference.nxd.t5.modeling_t5 import parallelize
+            from optimum.neuron.models.inference.t5.modeling_t5 import parallelize
 
             parallel_model = parallelize(model)
             # Load the weights into the parallel layers
--- a/optimum/neuron/models/inference/t5/modeling_t5.py
+++ b/optimum/neuron/models/inference/t5/modeling_t5.py
@@ -82,7 +82,7 @@ def prune_linear_layer(layer: BaseParallelLinear, index: torch.LongTensor, dim: 
     return new_layer
 
 
-class ParallelAttention(T5Attention):
+class NeuronT5Attention(T5Attention):
     def __init__(
         self,
         config: T5Config,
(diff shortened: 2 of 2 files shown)


JingyaHuang:
> How do you handle the sharding of the weights you load?

It's done within the wrappers as it used to be, eg. https://github.com/huggingface/optimum-neuron/blob/4b51935e189ed7413cae9c228cc67ed7712b39c1/optimum/exporters/neuron/model_wrappers.py#L191

https://github.com/huggingface/optimum-neuron/blob/4b51935e189ed7413cae9c228cc67ed7712b39c1/optimum/exporters/neuron/model_wrappers.py#L353

I agree that the contents in the previous wrappers could now go into the custom modeling, but I would leave that refactoring for later.

optimum/neuron/models/inference/nxd/t5/modeling_t5.py
```diff
@@ -0,0 +1,330 @@
+# coding=utf-8
```
dacorvo: The nxd directory has been remove, so you need to move t5 one directory up.

optimum/neuron/models/inference/nxd/t5/modeling_t5.py
```diff
@@ -0,0 +1,330 @@
...
+        if config.is_gated_act:
+            self.DenseReluDense = ParallelDenseGatedActDense(config)
+        else:
+            self.DenseReluDense = ParallelDenseActDense(config)
+
+        self.layer_norm = T5LayerNorm(config.d_model, eps=config.layer_norm_epsilon)
+        self.dropout = nn.Dropout(config.dropout_rate)
+
+
+def parallelize(model):
```
dacorvo: Instead of loading the model using transformers, then convert the layers, you could just instantiate the correct layers from the model config, like it is done in the training/inference custom modeling.

dacorvo (changes requested):
Thanks ! This deviates a bit from the pattern we used so far of recreating the model from the config, but otherwise it looks good to me.
You need to move the T5 directory up, since the nxd submodule has been removed.

optimum/neuron/models/inference/nxd/t5/modeling_t5.py
JingyaHuang: The way how custom modeling works for llm inference and training is different to what we have implemented for encoders and diffusion models. We could aim for a custom implementation to gather what's currently done with Neuron model config + NeuronModelForXXX, it's gonna be a pretty large refactoring. This PR is just to enable Michael to deprecate Parallelizer.

optimum/exporters/neuron/model_configs.py
```diff
@@ -966,10 +966,11 @@ def get_parallel_callable(
             library_name="transformers",
         )  # TODO: add extra args, eg. revision, trust_remote_code, etc.
         model.config.use_cache = True
-        parallelizer = ParallelizersManager.parallelizer_for_model(model)
-        with parallelizer.saved_model_in_temporary_directory(model) as ckpt_path:
-            # Replace parallel layers
-            parallel_model = parallelizer._parallelize(model, parallelize_embeddings=False)
+        with saved_model_in_temporary_directory(model) as ckpt_path:
+            # Plug in parallel layers
+            from optimum.neuron.models.inference.t5.modeling_t5 import parallelize
```
tengomucho, line 971: why isn't this import at the top of the file?

dacorvo (approved):
LGTM, thanks !

optimum/exporters/neuron/model_configs.py
JingyaHuang: It could be! Moving it to the top!

JingyaHuang: Nope, I would prefer to import inside the config, as I might have this func for other different models.

PR #877, opened 2025-06-25
Remove `optimum/neuron/distributed`

michaelbenayoun:
# What does this PR do?

Since we moved all the training features in `optimum/neuron/models/training`, everything inside `optimum/neuron/distributed` is deprecated and should be removed. That's what this PR does.

Remaining: cleaning the doc. Since it is in the roadmap anyways, I think we can focus on that later.


dacorvo (approved):
LGTM, thanks !

PR #879, opened 2025-07-01
[Inference] Fix head_dim usage in modeling

dacorvo:
# What does this PR do?

This fixes a serious issue in the decoder modeling code for Inference.

The legacy modeling code wrongly always assumed that `head_dim` was equal to `hidden_size // num_attention_heads`: this is true for a lot of models, but not for all of them for which a specific `head_dim` parameter is added to the config.

The issue was revealed by compilation errors due to shape mismatches with `Qwen3-0.6B` and `Qwen3-32B`.

### DIFF
--- a/optimum/neuron/models/inference/backend/modules/attention/attention_base.py
+++ b/optimum/neuron/models/inference/backend/modules/attention/attention_base.py
@@ -99,7 +99,7 @@ class NeuronAttentionBase(nn.Module):
         self.hidden_size = config.hidden_size
         self.num_attention_heads = config.num_attention_heads
         self.num_key_value_heads = config.num_key_value_heads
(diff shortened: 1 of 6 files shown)


dacorvo:
cc @jlonge4 @anishaga

anishaga:
> cc @jlonge4 @anishaga

Thank you so much for the blazingly fast fix 🚀😃. 
I've tested and can confirm that Qwen3-32B now compiles fine. However for inference, the [neuronx-tgi](https://github.com/huggingface/optimum-neuron/pkgs/container/neuronx-tgi) image isn't still updated with the latest transformers version. Can you please also release a new version of the neuronx-tgi docker container with the upgraded transformers version and neuron libraries so that we can start a TGI server with the Qwen3 model.

Issue #882, opened 2025-07-04
Auto classes for custom modeling


michaelbenayoun:
Create auto classes for custom modeling to be able to do:

```python
from optimum.neuron.models.training import AutoModelForCausalLM

model = AutoModelForCausalLM.from_pretrained("my_llama_model")
```

PR #883, opened 2025-07-04
Auto model classes for custom modeling

michaelbenayoun:
# What does this PR do?

As per title.

Fixes: #882

### DIFF
--- a/optimum/neuron/models/training/modeling_auto.py
+++ b/optimum/neuron/models/training/modeling_auto.py
@@ -1,0 +1,55 @@
+# coding=utf-8
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Union
+
+from transformers import AutoConfig
+
+from ..auto_model import get_neuron_model_class
+from .config import TrainingNeuronConfig
+
+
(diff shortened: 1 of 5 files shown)


optimum/neuron/models/training/modeling_auto.py
```diff
@@ -0,0 +1,55 @@
...
+        return neuron_model_class.from_pretrained(pretrained_model_name_or_path, trn_config, *model_args, **kwargs)
+
+    @classmethod
+    def from_config(cls, config, trn_config: TrainingNeuronConfig):
+        model_type = config.model_type
+        neuron_model_class = get_neuron_model_class(model_type, cls._task_name, "training")
+        return neuron_model_class(config, trn_config)
+
+
+class AutoModel(_BaseAutoModelClass):
```
JingyaHuang: Why do we use an API name that already exists in transformers? For inference, we opted for NeuronModelForXXX.

michaelbenayoun: What about `AutoNeuronModelForXXX`?

I want to keep the same spirit as the Transformers API.

JingyaHuang: For me, it's better to differentiate from transformers by marking down that it's neuron-specific, `AutoNeuronModelForXXX` could be the answer. 
In this case, we are going to use:
* `AutoNeuronModelForXXX` as auto model for training
* `NeuronModelForXXX` as auto model for inference
maybe some inputs @dacorvo @tengomucho

dacorvo: I agree with @JingyaHuang we should keep the pattern that has been used for two years+ for inference.

dacorvo: Ideally we should use the exact same names, regardless of whether it is inference or training

dacorvo: We can requst the user to import the training/inference version of each `NeuronModelFor`, or we find a way to dynamically figure out what version is requested.

michaelbenayoun: Done.

JingyaHuang (approved):
LGTM, thx!

dacorvo (approved):
LGTM, thanks !

PR #893, opened 2025-07-10
Remove example scripts

michaelbenayoun:
# What does this PR do?

There are a lot of example scripts taken from Transformers. First, they do not work anymore, and I do not think that they serve a real purpose. Removing them with the goal to replace them with less example scripts, but well tested and relevant to what the library offers:

- Inference examples
- Training examples: regular models, distributed training, mutli-node training. Basically these examples would be mapped to the tutorials in the docs.

### DIFF
--- a/examples/text-generation/generation.py
+++ b/examples/text-generation/generation.py
@@ -1,0 +1,96 @@
+import argparse
+import time
+
+import torch
+from transformers import AutoTokenizer, set_seed
+
+from optimum.neuron import NeuronModelForCausalLM
+
+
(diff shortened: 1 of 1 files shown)


examples/text-generation/generation.py
```diff
@@ -1,96 +0,0 @@
-import argparse
```
dacorvo: Please keep this one: it is not taken from transformers.

michaelbenayoun: Done

dacorvo (approved):
LGTM, thanks !

PR #894, opened 2025-07-10
Finetune llm example

michaelbenayoun:
# What does this PR do?

- Adds an example script to finetune LLMs with LoRA + SFTTrainer, it uses Qwen3.
- Creates the corresponding tutorial in the docs

This example script is 99% inspired from @tengomucho 's previous work.


docs/source/training_tutorials/finetune_qwen3.mdx
```diff
@@ -13,22 +13,26 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-In this example, we will use the `trn1.32xlarge` instance on AWS with 16 Accelerators, including 32 Neuron Cores and the Hugging Face Neuron Deep Learning AMI. 
-The Hugging Face AMI comes with all important libraries, like `datasets`, `transformers`, `optimum-neuron` and Neuron SDK packages pre-installed this makes it super easy to get started, since there is no need for environment management.
+We'll use a `trn1.32xlarge` instance with 16 Trainium Accelerators (32 Neuron Cores) and the Hugging Face Neuron Deep Learning AMI.
 
-To create such instance you can follow the guide [here](https://huggingface.co/docs/optimum-neuron/guides/setup_aws_instance).
+The Hugging Face AMI includes all required libraries pre-installed:
+- `datasets`, `transformers`, `optimum-neuron`
+- Neuron SDK packages
+- No additional environment setup needed
```
tengomucho, line 29: are you 100% sure about this? I think it might be necessary to install the training extra, that will install trl, peft, evaluate.

michaelbenayoun: Actually you are right. But it should be true as long as we install with the `training` extra when building the AMI.

tengomucho: OK for me, but we should make sure we add the `training` extra when building the AMI, I am not sure we do that right now.

PR #901, opened 2025-07-12
Bump accelerate to 1.3.0 + peft to 0.15.2

JingyaHuang:
# What does this PR do?

As per title, we need to update the accelerate and peft versions to support Flux model integration, thus the PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,8 +14,8 @@ except Exception as error:
 
 INSTALL_REQUIRES = [
     "transformers ~= 4.51.0",
-    "accelerate == 0.29.2",
-    "optimum ~= 1.23.3",
+    "accelerate == 1.3.0",
+    "optimum ~= 1.24.0",
     "huggingface_hub >= 0.29.0",
     "numpy>=1.22.2, <=1.25.2",
     "protobuf>=3.20.3, <4",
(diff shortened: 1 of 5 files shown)


optimum/exporters/neuron/model_configs.py
```diff
@@ -22,11 +22,20 @@
...
+from optimum.neuron.utils import (
+    ASTDummyAudioInputGenerator,
+    DummyBeamValuesGenerator,
+    DummyControNetInputGenerator,
+    DummyIPAdapterInputGenerator,
+    DummyMaskedPosGenerator,
+    DummyTimestepInputGenerator,
+    WhisperDummyTextInputGenerator,
+    saved_model_in_temporary_directory,
+)
```
michaelbenayoun, line 34: Why not keep the relative import?

setup.py
```diff
@@ -1,130 +0,0 @@
-import re
```
dacorvo, line 1: You also need to remove the references to setup.py in:
- Makefile,
- all ci workflows (also make sure they all reference use pyproject.toml in the `path` sections instead).

michaelbenayoun: Done!

PR #913, opened 2025-07-23
[docs] Move inference API section

michaelbenayoun:
# What does this PR do?

- Move the Inference API below, since it's package reference
- Move the Auto model classes API for inference under the corresponding section
- Remove the export API reference page since it was not containing anything meanigful


JingyaHuang (approved):
lgtm!!

PR #918, opened 2025-07-25
Trainers refactor

michaelbenayoun:
# What does this PR do?

Move all the related training files to `optimum/neuron/trainers`.

It also divides the several trainers in multiple files to make it easier to read.


optimum/neuron/trainers/transformers.py
```diff
@@ -0,0 +1,30 @@
...
+from .base import _TrainerForNeuron
+
+
+class NeuronTrainer(_TrainerForNeuron, Trainer):
+    """
+    Trainer that is suited for performing training on AWS Trainium instances.
+    """
+
+
+class Seq2SeqNeuronTrainer(_TrainerForNeuron, Seq2SeqTrainer):
```
tengomucho, line 27: have we got any test that uses this trainer?

michaelbenayoun: No, and it overwrite only evaluation related methods.

PR #925, opened 2025-07-30
fix: fixes barrier issue at the end of training with hub sync

michaelbenayoun:
# What does this PR do?

There was nested barriers during cache hub sync causing failure. This PR fixes that.

cc @jimburtoft

### DIFF
--- a/optimum/neuron/trainers/base.py
+++ b/optimum/neuron/trainers/base.py
@@ -276,13 +276,14 @@ class _TrainerForNeuron:
     def synchronize_hub_cache(self):
         cache_path = get_neuron_cache_path()
         repo_id = get_hf_hub_cache_repos()[0]
-        try:
-            synchronize_hub_cache(cache_path=cache_path, cache_repo_id=repo_id)
-        except Exception as e:
-            logger.warning(
-                f"Failed to synchronize the hub cache for {repo_id}. This is not a critical error, but it prevents"
-                f"compilation caching with the Hugging Face Hub. Error: {e}"
-            )
+        if not self.args.skip_cache_push:
+            try:
+                synchronize_hub_cache(cache_path=cache_path, cache_repo_id=repo_id)
+            except Exception as e:
+                logger.warning(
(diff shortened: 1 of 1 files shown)


optimum/neuron/trainers/base.py
```diff
@@ -275,13 +274,15 @@ def create_accelerator_and_postprocess(self):
...
+        cache_path = get_neuron_cache_path()
         repo_id = get_hf_hub_cache_repos()[0]
-        if not self.args.skip_cache_push and xr.global_ordinal() == 0:
-            has_write_access = has_write_access_to_repo(repo_id)
-            if has_write_access:
-                cache_path = get_neuron_cache_path()
-                synchronize_hub_cache(cache_path=cache_path, cache_repo_id=repo_id)
-        xm.rendezvous("Hub cache synchronization done")
+        try:
+            synchronize_hub_cache(cache_path=cache_path, cache_repo_id=repo_id)
```
tengomucho: hold on, you do not check if the ordinal is 0, meaning that all the workers will try to sync at the same time, is this something we want?

michaelbenayoun: Yes, it's specifically what was causing the error. This check is already done in `synchronize_hub_cache`, which was causing nested barriers and failing since we were already filtering on the ordinal.

tengomucho (approved):

optimum/neuron/trainers/base.py
jimburtoft: @michaelbenayoun I think we still want the option to to skip the cache push.  I use it in our workshop because ONE TIME it caused a job failure because of bad network connectivity, so we were just minimizing the places to fail.  However, with the try, if the call fails for ANY reason, it will still recover, right?

I don't know if there is any other use case for the flag or not, but if you get rid of it, you should remove the documentation for it and examples.  https://github.com/search?q=repo%3Ahuggingface%2Foptimum-neuron%20skip_cache_push&type=code

michaelbenayoun: Alright, yes I will keep it.

PR #935, opened 2025-08-13
chore: Bump diffusers to 0.35.*

JingyaHuang:
* Patch incompatibility introcued by diffusers 0.35.0
* Bump diffusers along with its requirements on some dependencies

# What does this PR do?

Solve incompatibilities with diffusers 0.35.0

* Cache Context: https://github.com/huggingface/diffusers/pull/11180

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #937, opened 2025-08-20
fix: flux neuron cache detection

JingyaHuang:
# What does this PR do?

TP size was not considered for diffusers before Flux, thus Optimum Neuron was unable to identify cached Flux artifacts, this PR would fix it.

## Before submitting

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ tests = [
     "mediapipe",
     "timm >= 1.0.0",
     "hf_transfer",
-    "torchcodec",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -75,7 +75,7 @@ tests = [
     "mediapipe",
     "timm >= 1.0.0",
     "hf_transfer",
-    "torchcodec",
+    "torchcodec < 0.6.0",
```
dacorvo, line 78: I think this conflicts with the changes in pr #935

PR #938, opened 2025-08-25
Trainer simplification

michaelbenayoun:
# What does this PR do?

This PR takes the `Trainer` from tranformers and tries to maintains most of its most important features and adapt them to work on AWS Trainium via the `NeuronTrainer`. Prior to this PR we would inherint from the `Trainer`, now we are completely disconnected from it, having full control on what we change and when we change it, as well as removing a lot of code that is simply not relevant to us for Neuron. 

We also define our own independent `NeuronTrainingArguments` in the same fashion.

For TRL Trainers, we simply adapt the `NeuronSFTTrainer` to work with it, we might apply the same philosophy in a later PR where we will not rely on the TRL library, let's see if it makes sense. We remove the `NeuronORPOTrainer` that was implemented only for the non-distributed setting and was not maintained.

Overall this PR simplifies a lot of the code and should make maintenance easier in the future.

Left to be done / checked in later PR:

- [ ] Test and check that ZeRO-1 works for DP, DP + TP, PP and with PEFT
- [ ] Test and check that gradient checkpointing works properly
- [ ] Solve the cast to bf16 approach instead of providing the `torch_dtype`?
- [ ] Add metrics to the trainer: # token seen in training, #tokens per seconds, mfu, etc
- [ ] Add resume training feature
- [ ] Update to latest TRL version and update the `NeuronSFTTrainer` accordingly.
- [ ] Add master weight saved in optimizer / optimizer state in FP32


optimum/neuron/accelerate/utils/dataclasses.py
michaelbenayoun, line 1: Not needed anymore, we simplify this by always using `torch.autocast` as suggested now by the Neuron SDK doc.

optimum/neuron/trainers/sft_trainer.py
michaelbenayoun, line 1: The changes consist only in transitioning from the former NeuronTrainer inheritance to the new implem.

michaelbenayoun:
The custom modeling tests fail in other [PRs](https://github.com/huggingface/optimum-neuron/actions/runs/17459874839/job/49581834947?pr=919) so it is likely not linked to this one. 
Also they pass locally which is weird, it needs investigation. I suggest we fix them in a small dedicated PR.

optimum/neuron/models/training/config.py
```diff
@@ -51,6 +51,11 @@ def __post_init__(self):
                 pipeline_model_parallel_size=self.pipeline_parallel_size,
             )
 
+        # Here we override the values with the ones actually set by NxD so we can use them interchangeably.
+        self.data_parallel_size = parallel_state.get_data_parallel_size()
+        self.tensor_parallel_size = parallel_state.get_tensor_model_parallel_size()
+        self.pipeline_parallel_size = parallel_state.get_pipeline_model_parallel_size()
```
tengomucho, line 57: why don't you make properties for these that call the corresponding parallel_state function? I think that will make it easier to read later on

optimum/neuron/trainers/sft_trainer.py
```diff
@@ -66,12 +61,17 @@ class PeftConfig:
...
 
 
-class _SFTTrainerTrainerInit(SFTTrainer):
-    def __init__(self, *args, **kwargs):
-        return Trainer.__init__(self, *args, **kwargs)
+_SFTTrainer = type(
+    "_SFTTrainer",
+    (NeuronTrainer,),
+    SFTTrainer.__dict__.copy(),
+)
```
tengomucho, line 71: I think this deserves an explanation. Here's a generated one:
```
Define a new class called _SFTTrainer using Python's type() function. The new class inherits from 
NeuronTrainer and copies the attributes and methods from SFTTrainer. This approach enables the code to 
combine or override behavior from both trainers.
```

optimum/neuron/models/training/config.py
michaelbenayoun: Why would it be easier?
Basically we do that to have one consistent value and use this across the code.

optimum/neuron/trainers/sft_trainer.py
michaelbenayoun: Done

optimum/neuron/trainers/sft_trainer.py
```diff
@@ -61,6 +61,9 @@ class PeftConfig:
         pass
 
 
+# Create a new class that inherints from NeuronTrainer to use this class instead of the transformers Trainer,
```
tengomucho: inherits

optimum/neuron/trainers/sft_trainer.py
```diff
@@ -61,6 +61,9 @@ class PeftConfig:
...
+# but has the same methods and attributes as SFTTrainer.
+# We can then inherint from this class to create our NeuronSFTTrainer.
```
tengomucho: inherit

tengomucho (commented):
There is a failure in the CI, do you think you can fix it before merging this?
Otherwise overall the change seems OK to me, thought I only did a quick review, given the massive amount of code.

optimum/neuron/trainers/sft_trainer.py
michaelbenayoun: Good catch!

optimum/neuron/trainers/sft_trainer.py
michaelbenayoun: Thanks

michaelbenayoun:
> There is a failure in the CI, do you think you can fix it before merging this? Otherwise overall the change seems OK to me, thought I only did a quick review, given the massive amount of code.

The CI failing is completely unrelated to this PR, and the fix is already ready here: #954

tengomucho (approved):

PR #939, opened 2025-08-25
allow safetensors to be downloaded for flux

Abdennacer-Badaoui:
When trying to download Flux models with `from_pretrained` from the Hub, the `.safetensors` weights inside the `transforms/weights` folder are currently ignored and therefore not downloaded.

Example:

```python
from optimum.neuron import NeuronFluxPipeline
pipe = NeuronFluxPipeline.from_pretrained("Jingya/FLUX.1-dev-neuronx-1024x1024-tp8")
```

This leads to missing weight files, since for Flux the weights and graph are stored separately, and the weights are provided as `.safetensors`.

Fix:

This PR updates the download logic to include `.safetensors` files, ensuring that the Flux pipeline loads correctly and all necessary weights are available.

### DIFF
--- a/optimum/neuron/modeling_diffusion.py
+++ b/optimum/neuron/modeling_diffusion.py
@@ -643,7 +643,7 @@ class NeuronDiffusionPipelineBase(NeuronTracedModel):
             # Downloads all repo's files matching the allowed patterns
             is_flux_model = "flux" in model_id.lower()
             if is_flux_model:
-                ignore_patterns = ["*.msgpack", "*.bin"]  # Weights an
+                ignore_patterns = ["*.msgpack", "*.bin"] 
             else:
                 ignore_patterns = ["*.msgpack", "*.safetensors", "*.bin"]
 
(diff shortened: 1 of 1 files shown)


optimum/neuron/modeling_diffusion.py
```diff
@@ -641,6 +641,12 @@ def _from_pretrained(
                 }
             )
             # Downloads all repo's files matching the allowed patterns
+            is_flux_model = "flux" in model_id.lower()
+            if is_flux_model:
+                ignore_patterns = ["*.msgpack", "*.bin"]
+            else:
+                ignore_patterns = ["*.msgpack", "*.safetensors", "*.bin"]
+
```
JingyaHuang: Maybe it's not necessary, let's just set the `ignore_patterns = ["*.msgpack", "*.bin"]`

JingyaHuang (changes requested):

optimum/neuron/modeling_diffusion.py
Abdennacer-Badaoui: Okay, whenever we have them, we should download them.

JingyaHuang: I think it should be applicable for all compiled neuron models for diffusers. sd and pixart won't have safetensors anyway

PR #943, opened 2025-08-28
[CIs]Refactoring Inference tests & Improve CIs speed

JingyaHuang:
# What does this PR do?

* Refactoring diffusers inference tests
* Increase CI speed

## Before submitting


tengomucho (approved):
LGTM!

dacorvo (approved):
LGTM, thanks ! I think we could go one step further and reduce the number of tests in the amount of legacy models tests, so that the corresponding pipelines are completed in less than 30 minutes

PR #946, opened 2025-08-29
Remove neff cache tool

dacorvo:
# What does this PR do?

Since `neuronx_distributed` now creates explicitly a `libneuronxla` `Cache` object to cache NEFF files, we don't need to patch the `generate_neff` function to do the same thing ourselves.


JingyaHuang (approved):
LGTM!

tengomucho (approved):
This is great, thanks! LGTM!

PR #948, opened 2025-09-02
Cleanup inference backend modules

dacorvo:
# What does this PR do?

This simplifies the inference backend modules by removing code paths conditioned by config parameters whose value is fixed for the models we currently support.


tengomucho (approved):
LGTM

PR #950, opened 2025-09-03
Remove unsupported modeling flags

dacorvo:
# What does this PR do?

This removes some optimized code paths that are not supported with the current models (mainly Llama and its variants) and/or deployment configurations (Trainium instances).


dacorvo:
@tengomucho thank you for the review, but I realized that while removing flash decoding I had also inadvertently removed flash attention for prefill. Will force-push.

optimum/neuron/models/inference/backend/modules/attention/attention_base.py
```diff
@@ -297,37 +281,8 @@ def get_flash_attention_strategy(self, q_len) -> FlashAttentionStrategy:
         if q_len >= 4096:
             return FlashAttentionStrategy.UNSHARDED_KERNEL
 
-        # At lower seq lens, enable only if explicitly enabled.
-        if self.attn_kernel_enabled and q_len >= 512:
-            return FlashAttentionStrategy.UNSHARDED_KERNEL
-
         return FlashAttentionStrategy.NONE
 
-    def compute_for_flash_decoding(self, Q, K, V, past_key_value, attention_mask, active_mask) -> Tensor:
```
tengomucho, line 306: I see, you need to restore these bits

PR #955, opened 2025-09-08
Remove optimized model dependency in LLM models

dacorvo:
# What does this PR do?

Before this pull-request `NeuronModelForCausalLM` used to inherit from `OptimizedModel`, but all methods were actually overridden, and the inheritance was more a hindrance than a benefit, with the awkward `from_pretrained` API for export. This removes the inheritance and simplifies the export of LLM models.

### DIFF
--- a/optimum/neuron/modeling_decoder.py
+++ b/optimum/neuron/modeling_decoder.py
@@ -107,7 +107,7 @@ class NeuronModelForCausalLM(NeuronModel, ABC):
     def get_neuron_config(
         cls,
         model_name_or_path: str | Path,
-        config: "PretrainedConfig",
+        config: PretrainedConfig | None = None,
         token: bool | str | None = None,
         revision: str | None = None,
         batch_size: int | None = None,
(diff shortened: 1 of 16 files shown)


tests/decoder/test_decoder_export.py
```diff
@@ -103,7 +109,39 @@ def test_decoder_export_save_reload(
...
-        check_neuron_model(model, **export_kwargs)
+        check_neuron_config(model.neuron_config, **export_kwargs)
+        check_neuron_model(model)
         del model
         model = NeuronModelForCausalLM.from_pretrained(model_path)
-        check_neuron_model(model, **export_kwargs)
+        check_neuron_model(model)
+
+
+@pytest.mark.parametrize("is_local", [True, False], ids=["local", "from_hub"])
```
tengomucho: is it realy necessary to do both local and from_hub

optimum/neuron/modeling_decoder.py
```diff
@@ -177,7 +175,7 @@ def get_neuron_config(
             elif config.torch_dtype == "bfloat16":
                 auto_cast_type = "bf16"
 
-        if type(cls) is NeuronModelForCausalLM:
+        if cls is NeuronModelForCausalLM:
```
tengomucho, line 184: this will not work with derived classes I think, though I do not think you planned to use it that way.

dacorvo: This is precisely to detect if we are not in a Derived class

PR #956, opened 2025-09-08
ZeRO-1 and mixed-precision

michaelbenayoun:
# What does this PR do?

It adds a better support for mixed-precision training. The following cases are supported:

- No ZeRO-1 optimizer, full bf16 model weights. By default in the `NeuronTrainer` there is no support for fp32 grads or optimizer states in this case. It can be done manually.
- No ZeRO-1 optimizer, `torch.autocast`. In this case the optimizer states will be in fp32, and only the relevant ops will be casted automatically for the forward / backward by `torch.autocast`.
- ZeRO-1 optimizer. In this case we support: bf16 / autocast weights with fp32 grads and optimizer states

On top of this, we refactor the tests at the same time Instead of having a `test_common.py` file, we now have:

- `test_checkpointing.py` for checkpointing tests
- `test_zero_1.py` for ZeRO-1 tests
- `test_optimizer.py` for optimizer tests that do not use ZeRO-1 optimization
- `test_mixed_precision.py` for mixed precision tests


optimum/neuron/trainers/training_args.py
```diff
@@ -324,7 +324,42 @@ class NeuronTrainingArguments:
             "help": "Whether to use torch autocast to perform mixed precision training.",
         },
     )
-    zero_1: bool = field(default=False, metadata={"help": "Whether to use  ZeRO Stage 1 Optimization."})
+    zero_1: bool = field(default=True, metadata={"help": "Whether to use  ZeRO Stage 1 Optimization."})
```
tengomucho, line 327: You are adding many options to the training args, but I guess it is very expensive to test and maintain them. We could just set them with some default values that make sense, e.g.: if you enable zero_1 the optimizer_use_master_weights and optimizer_use_fp32_grad_acc could be enabled, and we could just not expose these. What do you think?

tests/training/test_zero1.py
```diff
@@ -0,0 +1,299 @@
...
+        xm.mark_step()
+
+        # Gradient clipping seems to not work properly with ZeRO-1.
+        # max_grad_norm = 0.01
+        # accelerator.clip_grad_norm_(
+        #     model.parameters(), # This is not used when using ZeRO-1 but we need to pass something.
+        #     max_norm=max_grad_norm,
+        #     norm_type=2,
+        #     postpone_clipping_to_optimizer_step=True,
+        # )
```
tengomucho: I'd remove these commented code parts. consider adding a comment at the beginning of the test instead, explaining you expected gradient clipping to work but it does not.

optimum/neuron/trainers/training_args.py
michaelbenayoun: I agree and we set good defaults for that but we should still give the possibility not to do it. These options require a lot more memory so disabling it can be very useful sometimes. 

Keeping the optimizer states in fp32 accounts for 4 model sizes if we store the model in bf16.

tests/training/test_zero1.py
michaelbenayoun: Done

optimum/neuron/trainers/training_args.py
tengomucho: I do not see who is going to make use of all these combinations. I still think the price of maintaining all of them is way too high compared to the benefit of having all the options.

michaelbenayoun: We do not maintain much since those are features provided by `torch_xla`, we are just passing these options. 

Also it is a bit unconvenient to not expose a feature that could allow you to divide the memory usage by more than 33% if needed, but alright I will remove it.

michaelbenayoun: Done.

tengomucho (approved):

PR #957, opened 2025-09-09
Add tests for modules used in inference

tengomucho:
# What does this PR do?

Added few tests to verify accuracy of NxD modules used in inference used in some models (mainly Llama). 
This will also allow to verify that model implementation is still correct if we do some refactoring of those modules.


dacorvo (approved):
Awesome, thanks !

PR #960, opened 2025-09-11
Wiser ci workflow triggers

dacorvo:
# What does this PR do?

This pull-request attempts to limit the number of workflows that are triggered when submitting a pull-request or when merging it by adjusting the `paths` section of each workflow.

### DIFF
--- a/.github/workflows/test_inf2_seq2seq.yml
+++ b/.github/workflows/test_inf2_seq2seq.yml
@@ -12,6 +12,7 @@ on:
       - "optimum/neuron/modeling_seq2seq.py"
       - "optimum/neuron/models/inference/t5/**.py"
       - "optimum/neuron/utils/**.py"
+      - "tests/inference/seq2seq/**.py"
       - ".github/workflows/test_inf2_seq2seq.yml"
   pull_request:
     branches: [ main ]
@@ -24,6 +25,7 @@ on:
       - "optimum/neuron/modeling_seq2seq.py"
       - "optimum/neuron/models/inference/t5/**.py"
       - "optimum/neuron/utils/**.py"
+      - "tests/inference/seq2seq/**.py"
       - ".github/workflows/test_inf2_seq2seq.yml"
 
 concurrency:
--- a/.github/workflows/test_inf2_transformers.yml
+++ b/.github/workflows/test_inf2_transformers.yml
@@ -17,6 +17,7 @@ on:
       - "optimum/neuron/models/inference/yolos/**.py"
       - "optimum/neuron/pipelines/transformers/**.py"
       - "optimum/neuron/utils/**.py"
+      - "tests/inference/transformers/**.py"
       - ".github/workflows/test_inf2_transformers.yml"
   pull_request:
     branches: [ main ]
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -79,7 +79,7 @@ tests = [
 ]
 quality = [
     "pre-commit",
-    "ruff",
+    "ruff==0.13.0",
```
JingyaHuang, line 82: why?

.github/workflows/test_inf2_transformers.yml
```diff
@@ -5,14 +5,36 @@ on:
...
+      - "optimum/commands/**.py"
+      - "optimum/exporters/**.py"
+      - "optimum/neuron/cache/**.py"
+      - "optimum/neuron/modeling_base.py"
+      - "optimum/neuron/modeling_traced.py"
+      - "optimum/neuron/models/inference/bert/**.py"
+      - "optimum/neuron/models/inference/clip/**.py"
+      - "optimum/neuron/models/inference/t5/**.py"
+      - "optimum/neuron/models/inference/whisper/**.py"
+      - "optimum/neuron/models/inference/yolos/**.py"
```
JingyaHuang, line 35: shall we create some subfolders under `optimum/neuron/models/inference`? like for transformers / diffusers, otherwise we will need to add path each time we add support for a new model.

.github/workflows/test_inf2_seq2seq.yml
```diff
@@ -0,0 +1,64 @@
+name: Optimum neuron / Test INF2 seq2seq
```
JingyaHuang, line 1: why do we want a ci just for seq2seq? I feel that it shall be a part of `.github/workflows/test_inf2_transformers.yml`

.github/workflows/test_inf2_transformers.yml
tengomucho: would it be possible to filter the files in some other way, e.g. if the files contain `NeuronTracedModel` or something like that?

pyproject.toml
dacorvo: We need to pin the version otherwise the CI takes the latest by default, and it may be different from the local version. The goal here is to avoid repushing and relaunching the CI many times just for style errors.

.github/workflows/test_inf2_transformers.yml
dacorvo: @JingyaHuang I personally don't mind adding one line in this file when a new model is added. I would rather not have subfolders just to avoid that tiny little extra effort.
@tengomucho I don't think it is possible with github actions.

This is the best solution I found to avoid the current situation where irrelevant 30 minutes+ workflows (inf2 export, transformers and diffusers) are triggered if I just change one line in the modeling of an existing LLM model.

.github/workflows/test_inf2_seq2seq.yml
dacorvo: It could, but it wasn't: it was actually included in test_inf2.yml, after the cache and cli tests that have been dispatched. I just renamed the workflow to seq2seq.

PR #961, opened 2025-09-11
Enable CPU compilation

Abdennacer-Badaoui:
When compiling a model to Neuron format, the compilation itself does not require running on a Neuron-based instance; at least for encoder models (other architectures still need testing).
It wasn’t working with optimum-neuron because the keyword argument `cpu_backend` was missing in `torch_neuronx.trace()`.
The solution is to simply use:

```python 
model_neuron = torch_neuronx.trace(model, example_inputs, compiler_args="--target inf2", cpu_backend=True)
``` 
This keyword should be used with the `compiler_args` option to set the `--target` flag. This flag specifies the Neuron instance family on which the compiled model will run.
When running this on a CPU instance, you might encounter the following `RuntimeError`:

`If you want to compile on CPU, please supply compiler argument target, with one of options trn1, inf2, trn1n, or trn2. Ex: "--target trn1"` 

This happens because the NeuronX libraries attempt to auto-detect the target platform, expecting one of the options: trn1, inf2, trn1n, or trn2. On a CPU instance, it won’t find a compatible target.

If you have a Neuron-enabled instance, CPU compilation works without issue. But on a CPU-only instance, you can still compile, provided you set the platform target before importing NeuronX libraries. You can do:

```python
os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = "inf2"
```

With these changes, you can compile encoder models on CPU like this:

```python
import os
import sys
from pathlib import Path

# Configure environment for CPU compilation
sys.path.insert(0, str(Path(__file__).parent))
os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = "inf2"

from optimum.neuron import NeuronModelForFeatureExtraction, NeuronModelForMaskedLM, NeuronModelForQuestionAnswering, NeuronModelForSequenceClassification

def main():
    """
    Demo: Compile a model using cpu backend.
    """
    # Configuration
    model_name = "distilbert-base-uncased"
    output_dir = "./neuron_compiled_model"
    try: 
        # Load the model config first
        config = AutoConfig.from_pretrained(model_name) 
        
        # Compile the model on CPU
         neuron_model = NeuronModelForFeatureExtraction._export(
            model_id=model_name,
            config=config,
            batch_size=1,
            sequence_length=128,
            cpu_backend=True,
        )
        
        print("Compilation successful!")
        return 0
        
    except Exception as e:
        print(f"Compilation failed: {e}")
        return 1

if __name__ == "__main__":
    sys.exit(main())
```

Updates: 
- Seq2Seq models are now also supported for CPU compilation. 
- For decoder models, the export is already CPU-only. If you want to export your model on a CPU instance, follow the example below:

```python
import os
import sys
from pathlib import Path

# Configure environment for CPU compilation
sys.path.insert(0, str(Path(__file__).parent))
os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = "inf2"

from optimum.neuron import NeuronModelForCausalLM

def main():
    """
    Demo: Compile a model using cpu backend.
    """
    # Configuration
    output_dir = "./neuron_compiled_model"
    try: 
        export_decoder_id = "Qwen/Qwen2.5-0.5B"
        export_kwargs = {"batch_size": 1, "sequence_length": 1024, "tensor_parallel_size": 2, "auto_cast_type": "bf16"}
        neuron_config = NeuronModelForCausalLM.get_neuron_config(model_name_or_path=export_decoder_id, **export_kwargs)
        model = NeuronModelForCausalLM.export(
                model_id=export_decoder_id, neuron_config=neuron_config
            )
        model.save_pretrained(output_dir)

        print("Compilation successful!")
        return 0
        
     except Exception as e:
        print(f"Compilation failed: {e}")
        return 1

if __name__ == "__main__":
    sys.exit(main())
```


optimum/neuron/modeling_traced.py
```diff
@@ -176,6 +192,7 @@ def _from_pretrained(
         local_files_only: bool = False,
         model_save_dir: str | Path | TemporaryDirectory | None = None,
         neuron_config: "NeuronDefaultConfig | None" = None,
+        cpu_backend: bool = False,
```
JingyaHuang: `_from_pretrained` is for loading the checkpoint, it would be meaningful only on neuron devices in our case. So instead if overriding the func which should never be called if there is no neuron equipped, you can just skip loading the exported artifacts in `_export`.

JingyaHuang (commented):
Thanks for working on it, seems that you are still working on unblocking the cpu-only compilation for other architectures. Just a small comment on the encoder case, and it would be nice that you add an cpu-only CI for testing the export as well. Great job!

.github/workflows/test_cpu_compilation.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Prepare virtual environment
+        uses: ./.github/actions/prepare_venv
+      - name: Install optimum-neuron
+        uses: ./.github/actions/install_optimum_neuron
+      - name: Setup PATH
+        run: echo "/home/ubuntu/.local/bin" >> $GITHUB_PATH
+      - name: Set environment variables for CPU backend
+        run: |
+          echo "NEURON_PLATFORM_TARGET_OVERRIDE=inf2" >> $GITHUB_ENV
+      - name: Run CPU compilation encoder tests
```
JingyaHuang: Why do you run encoder test v.s. seq2seq test separately? And not running the stable diffusion test?

Abdennacer-Badaoui: Yeah, I was testing something and forgot to change it back. I’ll correct it right away, thanks!

optimum/exporters/neuron/__main__.py
```diff
@@ -626,6 +626,8 @@ def main_export(
     compiler_workdir: str | Path | None = None,
     inline_weights_to_neff: bool = True,
     optlevel: str = "2",
+    instance_type: str = "trn1",
```
JingyaHuang, line 629: Have you checked the difference of compiled artifacts between `trn1` and `inf2`? Guess the neuron doc set it to `trn1`, but I find it confusing that we set `NEURON_PLATFORM_TARGET_OVERRIDE` to "inf2" while having `trn1` as default value.

optimum/exporters/neuron/convert.py
```diff
@@ -326,6 +328,10 @@ def export_models(
                 1: enables the core performance optimizations in the compiler, while also minimizing compile time.
                 2: provides the best balance between model performance and compile time.
                 3: may provide additional model execution performance but may incur longer compile times and higher host memory usage during model compilation.
+        instance_type (`str`, defaults to `"trn1"`):
+            The instance type to use for the Neuron device.
+        cpu_backend (`bool`, defaults to `False`):
+            Whether to compile the model with CPU backend.
```
JingyaHuang:
```suggestion
            Whether to trace the model completely on CPU.
```

optimum/neuron/modeling_seq2seq.py
```diff
@@ -395,11 +397,17 @@ def _export(
...
         )
 
-        return cls._from_pretrained(
-            model_id=save_dir_path,
-            config=config,
-            model_save_dir=save_dir,
-        )
+        if cpu_backend:
+            logger.warning(
+                "Model was compiled with cpu_backend=True. Model loading is skipped as it requires Neuron hardware."
```
JingyaHuang:
```suggestion
                "Since `cpu_backend` is set to `True` during compilation, model loading is skipped."
```

optimum/neuron/modeling_traced.py
```diff
@@ -387,7 +400,13 @@ def _export(
             )
             config = AutoConfig.from_pretrained(save_dir_path)
 
-        return cls._from_pretrained(save_dir_path, config, model_save_dir=save_dir)
+        if cpu_backend:
+            logger.warning(
+                "Model was compiled with cpu_backend=True. Model loading is skipped as it requires Neuron hardware."
+                "The model compilation was successful and the artifacts were saved."
+            )
```
JingyaHuang, line 402: return None

tests/exporters/test_cpu_compilation.py
```diff
@@ -0,0 +1,406 @@
...
+# Models to test for CPU backend compilation
+CPU_BACKEND_ENCODER_MODELS = {
+    "bert": "hf-internal-testing/tiny-random-BertModel",
+}
+
+CPU_BACKEND_SEQ2SEQ_MODELS = {
+    "t5": "hf-internal-testing/tiny-random-t5",
+}
+
+CPU_BACKEND_DECODER_MODELS = {
```
JingyaHuang: To remove?

optimum/neuron/modeling_traced.py
```diff
@@ -94,18 +94,23 @@ class NeuronTracedModel(OptimizedModel, NeuronModel):
...
         model_save_dir: str | Path | TemporaryDirectory | None = None,
         model_file_name: str | None = None,
         preprocessors: list | None = None,
         neuron_config: "NeuronDefaultConfig | None" = None,
         **kwargs,
     ):
-        super().__init__(model, config)
-        if hasattr(model, "device"):
-            self.device = model.device
+        # Handle the case where model is None (cpu_backend compilation without loading)
```
JingyaHuang: btw, do we really need to change `__init__()`? If using the cpu backend for compilation, we won't go to the instantiation in `_from_pretrained`, and I don't feel the instantiation of Neuron model with None make too much sense.

optimum/exporters/neuron/__main__.py
JingyaHuang: Where did you find `NEURON_PLATFORM_TARGET_OVERRIDE`?

Abdennacer-Badaoui:
Thanks @JingyaHuang, I’ll look at the review and make the necessary changes.

optimum/exporters/neuron/__main__.py
Abdennacer-Badaoui: I didn’t check the difference, to be honest. The important thing is to override the target to one of the Neuron devices before importing the NeuronX libraries. Otherwise, on a CPU, we’ll get an error saying that no Neuron device is available.

I can just set it to trn1; I’m only using it to bypass the error below:

```
RuntimeError(f"Unsupported Platform - {fc}. If you want to compile on CPU, \
RuntimeError: Unsupported Platform - c5.4xlarge
. If you want to compile on CPU, please supply compiler argument target, with one of: trn1, inf2, trn1n, or trn2. 
Ex: "--target trn1"
```

This error comes from:
`/home/ec2-user/neuronx-env/lib64/python3.11/site-packages/torch_neuronx/utils/utils.py:46
`
In that file, we find the NEURON_PLATFORM_TARGET_OVERRIDE environment variable, which is used by the Neuron SDK to override automatic hardware platform detection.

optimum/neuron/modeling_traced.py
Abdennacer-Badaoui: That's make sense!

optimum/exporters/neuron/__main__.py
JingyaHuang: Thanks, I will do a following-up PR to enable auto-detection of instance type and passing the target via `compiler_args`, let's merge this first.

PR #964, opened 2025-09-12
PEFT and PP

michaelbenayoun:
# What does this PR do?

This PR adds support for performing pipeline parallelism training with PEFT models.


optimum/neuron/models/training/modeling_utils.py
michaelbenayoun, line 1: Soft tie was originally added because there was a compiler bug. But doing this prevented training with PEFT. It now works.

Issue #171, opened 2023-08-17
Invalid target for neuronx-cc on inf2 using trn1 instead of inf2 


fafriat:
Hi,
I have remarked the target of neuronx-cc when using NeuronXXX models on inf2 is trn1. I checked that it is compatible to inf2 anyway, but do we lose any optimization using this wrong target ?

JingyaHuang:
Hi @fafriat, what do you mean by "target"? 

`NeuronModelForXXX` classes support inference on both neuron device 1st and 2nd generation. So you can use it on INF2, INF1 and TRN1 instances.  But it's true that some features are only supported on INF2 not INF1.

You can try it out with our showcase examples: https://huggingface.co/docs/optimum-neuron/guides/models. Feel free to reach out if you have any questions.

fafriat:
Hi @JingyaHuang,

When using NeuronModelForXXX, it is automatically compiling for Neuron when export=True,
 In that case, when running on inf2 instance, I have remarked the command triggered from neuronx-cc is using --target trn1 instead of --target inf2...

I know that it is also working on inf2 when target is trn1 but do we lose any  optimization using trn1 instead of inf2 as target ?

aws-donkrets:
Hi @fafriat,
Just noticed this older ticket. The Neuron compiler accepts both the inf2 and trn1 target option-arguments. The ISA between these two instances is similar enough such that the compiler (currently) performs the same optimizations so you are not losing any instance-specific optimizations.  These are defined as separate option-arguments in case, in the future, we identify optimizations that are beneficial specifically for interference or training models. So it is good practice to specify the option-argument based upon the model's purpose.

JingyaHuang:
Close the issue thanks to the answer of @aws-donkrets. Please feel free to re-open if there is any further questions.

PR #974, opened 2025-09-24
Enable `instance_type` tag to export

### LINKED ISSUE #971: Export: add support for trn2
### Feature request

When using the export functionality it is hardcoded to pass the `--target` param to be set to `trn1`. There should be a way to either 1/ pass in the instance type as a parameter to the export call or 2/ automatically detect the instance type. There is a parameter called `instance_type` but it cannot be passed into the export function it seems.

### Motivation

Need to support the latest instance types

### Your contribution

Can give feedback on the feature

JingyaHuang:
# What does this PR do?

Fixes #971 

- [x] Enable passing a custom target instance type during the compilation.
- [x] Compatibility with CPU-only compilation
- [x] Improve CPU-only compilation on CLI & diffusion models
- [x] Reduce CPU-only tests workload
- [x] CPU-only Compilation doc
- [x] Update export doc

## Before submitting


mmcclean-aws:
Runs fine on trn2

optimum/exporters/neuron/__main__.py
```diff
@@ -688,20 +691,24 @@ def main_export(
...
-        do_validation = False
+    if do_validation:
+        if tensor_parallel_size > 1:
+            # TODO: support the validation of tp models.
+            logger.warning(
+                "The validation is not yet supported for tensor parallel model, the validation will be turned off."
+            )
+            do_validation = False
+        elif cpu_backend:
+            logger.info("The validation is diabled since you are using CPU backend for the compilation.")
```
tengomucho: diabled -> disabled

optimum/commands/export/neuronx.py
```diff
@@ -42,6 +48,18 @@ def parse_args_neuronx(parser: "ArgumentParser"):
...
+    optional_group.add_argument(
+        "--cpu_backend",
+        action="store_true",
+        help="Whether to trace the model completely on CPU.",
+    )
+    optional_group.add_argument(
+        "--instance_type",
+        type=str,
+        default=None,
+        choices=["inf2", "trn1", "trn1n", "trn2"],
```
tengomucho: the list of instance types is repeated in several places in the code, could you make a constant that defines them once instead?

optimum/neuron/utils/system.py
```diff
@@ -17,10 +17,13 @@
 import os
 import re
 
+from torch_neuronx.utils import get_platform_target
+
 
 NEURON_DEV_PATTERN = re.compile(r"^neuron\d+$", re.IGNORECASE)
 MAJORS_FILE = "/proc/devices"
 NEURON_MAJOR_LINE = re.compile(r"^\s*(\d+)\s+neuron\s*$")
+SUPPORTED_INSTANCE_TYPES = ["trn1", "inf2", "trn1n", "trn2"]
```
tengomucho: you have it already! use this

optimum/neuron/utils/system.py
```diff
@@ -76,3 +79,17 @@ def get_available_cores() -> int:
             visible_cores = 1
     visible_cores = min(visible_cores, num_cores)
     return visible_cores
+
+
+def get_neuron_instance_type(instance_type: str | None) -> str:
```
tengomucho: why not moving this to a dedicated file, something like `instance_types.py`, so that we avoid having generic system function?

optimum/exporters/neuron/__main__.py
JingyaHuang: good catch!

optimum/neuron/utils/instance.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    # Normalize instance type
+    instance_type = INSTANCE_VALUE_MAP[instance_type]
+    os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = instance_type
+
+    return instance_type
+
+
+def is_cpu_only_instance():
```
dacorvo: I think it is better to simply use `get_neuron_major` like @tengomucho did in his pull-request:
https://github.com/huggingface/optimum-neuron/pull/985/commits

docs/source/guides/export_model.mdx
```diff
@@ -114,6 +114,19 @@ optimum-cli export neuron --model distilbert-base-uncased-distilled-squad \
 Here the model was exported with a static input shape of `(1, 16)`, and with compiler arguments specifying
 that matmul operation must be performed with `float16` precision for faster inference.
 
+<Tip>
+
+You can also compile the model on a CPU-only instance. In this case, specify the target instance type by passing `--instance_type` from `{inf2, trn1, trn1n, trn2}`.
```
dacorvo, line 119: I did not modify that in my pull-request because I wanted to make it easier for you to rebase on top of it but I think we should restrict the choices to trn1/trn2, at least in the CLI.

optimum/commands/export/neuronx.py
```diff
@@ -333,5 +335,21 @@ def parse_args(parser: ArgumentParser):
         return parse_args_neuronx(parser)
 
     def run(self):
+        self.setup_target_instance(self.args_string)
```
dacorvo: Alternatively, you could have detected that we were on the CPU instance when adding arguments and set `required=True` when adding `instance_type`.

docs/source/guides/export_model.mdx
```diff
@@ -114,6 +114,19 @@ optimum-cli export neuron --model distilbert-base-uncased-distilled-squad \
 Here the model was exported with a static input shape of `(1, 16)`, and with compiler arguments specifying
 that matmul operation must be performed with `float16` precision for faster inference.
 
+<Tip>
+
+You can also compile the model on a CPU-only instance. In this case, specify the target instance type by passing `--instance_type` from `{inf2, trn1, trn1n, trn2}`. 
+
+If you are using a `NeuronModelForXXX` class to export the model on a CPU-only instance, set the target instance type through the environment variable, for example:
```
tengomucho: Since this is an explanation on how to do that programmatically, it might be worth mentioning you need to export the env var before the imports, right?

optimum/exporters/neuron/convert.py
```diff
@@ -713,6 +706,7 @@ def trace_neuronx(
             neuronx_distributed.trace.parallel_model_save(neuron_model, output)
     else:
         # Case 3: Using `torch_neuronx.trace`
+        cpu_backend = get_available_cores() == 0
```
tengomucho: I think you can just do
```suggestion
        cpu_backend = (get_neuron_major() == -1)
```
It is going to be faster.

optimum/neuron/modeling_diffusion.py
```diff
@@ -1034,13 +1040,19 @@ def _export(
...
                 **input_shapes,
             )
-
-        return cls._from_pretrained(
-            model_id=save_dir_path,
-            config=config,
-            model_save_dir=save_dir,
-            data_parallel_mode=data_parallel_mode,
-        )
+        if get_available_cores() == 0:
```
tengomucho: here too you can just use `get_neuron_major`

optimum/neuron/modeling_seq2seq.py
```diff
@@ -397,9 +402,9 @@ def _export(
             **kwargs_shapes,
         )
 
-        if cpu_backend:
+        if get_available_cores() == 0:
```
tengomucho: here too

optimum/neuron/modeling_traced.py
```diff
@@ -395,9 +397,9 @@ def _export(
             )
             config = AutoConfig.from_pretrained(save_dir_path)
 
-        if cpu_backend:
+        if get_available_cores() == 0:
```
tengomucho: here too

optimum/commands/export/neuronx.py
```diff
@@ -333,5 +335,21 @@ def parse_args(parser: ArgumentParser):
...
+    def setup_target_instance(args_string: str):
+        if get_available_cores() == 0:
+            # `--instance_type` is mandary when we compile w/o Neuron devices.
+            compiler_args = shlex.split(args_string)
+            if "--instance_type" not in args_string:
+                raise RuntimeError(
+                    f"You are using an instance without any neuron device, please supply target instance type among {SUPPORTED_INSTANCE_TYPES} for cpu-only compilation"
+                )
+            else:
+                index = compiler_args.index("--instance_type")
```
dacorvo: This can be done also using a custom action: https://docs.python.org/3/library/argparse.html#action

optimum/commands/neuron/subcommands.py
```diff
@@ -53,6 +52,8 @@ def parse_args(parser: "ArgumentParser"):
         )
 
     def run(self):
+        from ...neuron.models.training import consolidate_model_parallel_checkpoints_to_unified_checkpoint
```
dacorvo: Why that change ? This file should be renamed `consolidate.py` by the way ...

optimum/exporters/neuron/convert.py
```diff
@@ -445,13 +444,12 @@ def export(
     model_or_path: "PreTrainedModel | str | Path",
     config: "NeuronDefaultConfig",
     output: Path,
+    instance_type: Literal["trn1", "inf2", "trn1n", "trn2"] | None = None,
```
dacorvo, line 447: You could use SUPPORTED_INSTANCE_TYPES here

optimum/neuron/utils/__init__.py
```diff
@@ -85,6 +85,13 @@
         "patch_within_function",
         "replace_class_in_inheritance_hierarchy",
     ],
+    "instance": [
```
dacorvo: I know it is done similarly for all other "utils", but won't this lead to unnecessary imports ? I prefer importing with the full path myself.

optimum/neuron/utils/instance.py
```diff
@@ -87,3 +87,16 @@ def align_compilation_target(target: str, override: bool):
         logger.info(f"Setting the compilation target to {target}.")
     os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = target
     return target
+
+
+def define_instance_type_with_default_value(instance_type: str | None = None):
```
dacorvo: I would name it `get_default_compilation_target()` and call it only if `instance_type` is None.

optimum/neuron/modeling_traced.py
```diff
@@ -301,9 +302,13 @@ def _export(
             kwargs_shapes["batch_size"] = 1
             disable_fallback = True  # Turn off the fallback for neuron, otherwise dynamic batching will still fail
         auto_cast_type = None if auto_cast is None else auto_cast_type
+        instance_type = define_instance_type_with_default_value(instance_type)
```
dacorvo: I understand that this helper replaces the code I used here:
https://github.com/huggingface/optimum-neuron/blob/87f394bad111d0311134fbc4114d0d3b32e788a9/optimum/neuron/modeling_decoder.py#L166
Just to be picky, this is the compilation target you retrieve here, based on the instance_type.

dacorvo (commented):
Thanks for the pull-request. All my comments revolve around the confusion between instance type (the current platform) and compilation target (the target platform). I think we should avoid mixing the two terminologies. On a CPU platform, the instance type is undefined, so the term can be used to refer to the compilation target. On a neuron platform on the other hand both terms can have different values.

docs/source/guides/export_model.mdx
dacorvo: Because here, it is precisely the compilation target we set, and only `trn1` and `trn2` are allowed.

optimum/commands/neuron/subcommands.py
JingyaHuang: Forgot to remove this sry.

docs/source/guides/export_model.mdx
JingyaHuang: Although inf2/trn1/trn1n are talking about the same thing (so far), we should allow users to type in the instance types, since it's quite counterintuitive for them to use `trn1` when they are targeting to deploy on inf2 (cf. #171), anyway we have themapping to correct this to what could be consumed by neuron. tl;dr I think we should allow users to choose from all available neuron instance types.

dacorvo: I disagree, because `inf2` is just a marketing name, and eventually you export for the `trn1` compilation target. It is the same for CUDA: you generate for a CUDA version / hardware version, and the binary can be used on several different instances that support at least that version.

JingyaHuang: the flag is called `instance_type` and inf2 is a valid instance for neuron although the compiler takes it the same as trn1. 
From the comment of Annapurna, it seems that it's not 100% impossible that one day they distinguish them: https://github.com/huggingface/optimum-neuron/issues/171#issuecomment-1793613824. And as you can see in the issue, it causes confusion for the user that he wanted to deploy on inf2 but the flag was trn1.

I can reduce the choices here, and clarify that inf2 should go to trn1 if you insist.

optimum/neuron/utils/__init__.py
JingyaHuang: will clean this

optimum/exporters/neuron/convert.py
JingyaHuang: How? Feel like it would be illegal: https://typing.python.org/en/latest/spec/literal.html

dacorvo: I think you might be right, my mistake  😜

docs/source/guides/export_model.mdx
dacorvo: Yes, I see we have a different view on how to present this: you think it is better to expose the instance types, assuming the user knows them, where I think it is better to expose the platform types / compilation targets (trn1, trn2). The naming you chose for the parameter is consistent, so let's stick with your list. I might create a list of compilation targets also, to use internally.

dacorvo:
On another subject, I think we should hide and deprecate auto_cast_type, and always use the dtype from the model config. I'll do that at the very least for LLMs, because switching from bfloat16 to float16 for instance can be disastrous for accuracy.

optimum/commands/export/neuronx.py
```diff
@@ -335,21 +336,12 @@ def parse_args(parser: ArgumentParser):
         return parse_args_neuronx(parser)
 
     def run(self):
-        self.setup_target_instance(self.args_string)
+        if get_neuron_major() == -1:
+            self.setup_target_instance()
+        self.setup_target_instance()
```
dacorvo: Did you forget to remove this line ?

JingyaHuang: oups, good catch!

JingyaHuang:
> On another subject, I think we should hide and deprecate auto_cast_type, and always use the dtype from the model config. I'll do that at the very least for LLMs, because switching from bfloat16 to float16 for instance can be disastrous for accuracy.

yeah, true bad to cast from bf16 to fp16, but we might want to keep the possibility of casting fp32 checkpoints to bf16 or fp16 no?

optimum/commands/export/neuronx.py
```diff
@@ -107,8 +110,8 @@ def parse_args_neuronx(parser: "ArgumentParser"):
         "--auto_cast_type",
         type=str,
         default="bf16",
-        choices=["bf16", "fp16", "tf32"],
-        help='The data type to cast FP32 operations to when auto-cast mode is enabled. Can be `"bf16"`, `"fp16"` or `"tf32"`.',
+        choices=["bf16", "fp16", "tf32", "fp8_e4m3"],
```
tengomucho: I think it would be better to remove casting to fp8: I do not think this will work on existing models, quantization should be used in that case, otherwise models will not work as expected.

JingyaHuang: Ok, I can remove it when the CIs are green, don't want to re-run everything for this.

optimum/commands/export/neuronx.py
```diff
@@ -111,7 +111,7 @@ def parse_args_neuronx(parser: "ArgumentParser"):
         type=str,
         default="bf16",
         choices=["bf16", "fp16", "tf32", "fp8_e4m3"],
```
tengomucho:
```suggestion
        choices=["bf16", "fp16", "tf32"],
```

JingyaHuang: oups

tengomucho (approved):

PR #976, opened 2025-09-25
Fix async save

michaelbenayoun:
# What does this PR do?

Async saving was broken when the directories changed between two async saves, and with the previous not being done yet.
This PR solves this by waiting for the previous save to be done and by creating the expected files when the directory is different.


optimum/neuron/models/training/modeling_utils.py
```diff
@@ -1413,6 +1414,19 @@ def save_pretrained(
...
+        # There is this mechanism in neuronx_distributed already, but it breaks when we change the name of the
+        # directory between saves, which we done (checkpoint-10, checkpoint-20, etc).
+        # So we add this extra wait here that uses the same mechanism but prior to calling the save function.
+        if self.trn_config.async_save:
+            g_iostate = neuronx_distributed.trainer.checkpoint.g_iostate
+            if g_iostate is not None:
+                g_iostate.wait_save(async_remove=True)
+                if xr.global_ordinal() == 0:
+                    (save_directory / MODEL_PARALLEL_SHARDS_DIR_NAME).mkdir(parents=True, exist_ok=True)
+                    (save_directory / MODEL_PARALLEL_SHARDS_DIR_NAME / "done").touch(exist_ok=True)
```
tengomucho: You have the same code in modeling_utils.py and peft_model.py. Can you make a function and share it?
Also, what is the "done" for?

michaelbenayoun: Alright. `done` is a file to check that the previous saving is done when doing async saving. It is from `neuronx_distributed`. It's not an issue, but there is a "bug" in NxD because they set the checkpoint dir before calling `g_iostate.wait_save`, so `wait_save` checks the new checkpoint dir for the presence of the `done` file, and fail when we change of checkpoint directory between two saves, which is the case during training.

tengomucho: I see, thanks! Consider adding more info, maybe pointing to the place where the bug is (a permalink to the NxD code maybe) so it will be easier to understand.

michaelbenayoun: Done

PR #980, opened 2025-09-29
Ci install pytorch cpu

tengomucho:
# What does this PR do?

Optimum Neuron requires torch, but it does not need the default CUDA based installation. This dependency takes a lot of disk storage.
Manually installing the CPU version will mitigate failures due to disk being full.

Also, there were some [failures](https://github.com/huggingface/optimum-neuron/actions/runs/18057781790/job/51389972521#step:8:89) due to tests dependencies on vLLM when pre-compiling tests. Restricting those to training tests should sort this out.

### DIFF
--- a/.github/actions/prepare_venv/action.yml
+++ b/.github/actions/prepare_venv/action.yml
@@ -13,3 +13,4 @@ runs:
           python -m pip install -U pip
           python -m pip install --upgrade setuptools==69.5.1
           python -m pip install hf_transfer
+          python -m pip install torch==2.7.1 torchvision~=0.22 --index-url https://download.pytorch.org/whl/cpu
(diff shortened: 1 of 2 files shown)


.github/actions/prepare_venv/action.yml
```diff
@@ -13,3 +13,4 @@ runs:
           python -m pip install -U pip
           python -m pip install --upgrade setuptools==69.5.1
           python -m pip install hf_transfer
+          python -m pip install torch==2.7.1 torchvision~=0.22 --index-url https://download.pytorch.org/whl/cpu
```
michaelbenayoun, line 18: Maybe add a comment explaining why it's necessary

michaelbenayoun (approved):

JingyaHuang (approved):
lgtm, thanks!

PR #983, opened 2025-09-30
Reduce tests dependencies

tengomucho:
The extra tests has been growing, so we now install many packages that we do not need anymore.
The idea is to remove these dependencies so to lean up workflows.


dacorvo:
There is still an issue with audio tests using datasets, but it will be fixed when bumping the SDK version and switching to torch 2.8

PR #985, opened 2025-10-02
ECR Image URI retrieval

tengomucho:
# What does this PR do?

Added a function to retrieve an image from the AWS container registry (ECR). These images can be used, for instance, to deploy an inference on Sagemaker. Example:

```python
from optimum.neuron.utils import ecr
image_uri = ecr.image_uri(region=region, service_name="tgi")
```

This could return the most recently published image URI, that can be newer than the one retrieved when using `get_huggingface_llm_image_uri` from latest Sagemaker Python package.

### DIFF
--- a/tests/sagemaker/test_images_uri.py
+++ b/tests/sagemaker/test_images_uri.py
@@ -1,3 +1,5 @@
+import pytest
+
 from optimum.neuron.utils import ecr
 
 
@@ -15,3 +17,12 @@ def test_ecr_image_uri():
     # only version
     image_uri = ecr.image_uri(version="3.3.4")
     assert image_uri is not None
+    # invalid version
+    image_uri = ecr.image_uri(version="ABCD")
+    assert image_uri is None
+    # not valid service name
+    with pytest.raises(ValueError, match="Invalid service name"):
+        ecr.image_uri(service_name="ABCD")
+    # not valid region
+    with pytest.raises(KeyError, match="ABCD"):
+        ecr.image_uri(region="ABCD")


optimum/neuron/utils/ecr.py
```diff
@@ -0,0 +1,99 @@
...
+        region = ecr_client.meta.region_name
+    repository_id = ACCOUNT_IDS[region]
+    if service_name not in IMAGE_SERVICES:
+        raise ValueError(f"Invalid service name: {service_name}")
+    repository_name = IMAGE_SERVICES[service_name]
+    try:
+        images = ecr_client.list_images(repositoryName=repository_name, registryId=repository_id)["imageIds"]
+    except boto3.exceptions.botocore.exceptions.NoCredentialsError as e:
+        message = (
+            "An error occurred while listing images from the ECR repository. Please check you have exported "
```
dacorvo: There are different ways to set AWS credentials. Exporting the corresponding env vars is not the only one.

optimum/neuron/utils/ecr.py
```diff
@@ -0,0 +1,99 @@
...
+            + " the AWS credentials. Original error: "
+            + str(e)
+        )
+        raise ValueError(message)
+
+    neuronx_images = [image for image in images if "neuronx" in image["imageTag"]]
```
dacorvo: This is perhaps a bit too specific to the current TGI convention. Maybe this should be configurable with a regexp indexed by the target service. Same thing for the version check

dacorvo (approved):

PR #986, opened 2025-10-03
Improve DX when exporting and deploying LLM neuron models

dacorvo:
# What does this PR do?

This modifies the `optimum-cli` `neuron` commands used when exporting and deploying LLM models to neuron to improve the developer experience.

The `optimum-cli neuron cache lookup` command is modified to display a summarized table listing cached configurations.

The `optimum-cli neuron serve` command is modified to automatically select the best export parameters when deploying a standard model, prioritizing the available cached configurations. A parameter is also added to serve models even if no cached configuration is available.


tengomucho (approved):
LGTM

PR #988, opened 2025-10-07
Add support for Trainium 2 for decoder models

dacorvo:
# What does this PR do?

This adds a new `instance_type` parameter to the decoder export that can be set to either `trn1` or `trn2` for Trainium 1 or Trainium 2 compilation targets respectively.

Rule of thumb from my tests:
- we MUST set the NEURON_PLATFORM_TARGET_OVERRIDE if we are exporting from a cpu host, or from a trn1 host to a trn2 target,
- we can change this parameter anytime we want as log as we do not import nxd. Once the import has been done we're stuck with whatever target was guessed by the runtime.

### DIFF
--- a/tools/prune_test_models.py
+++ b/tools/prune_test_models.py
@@ -13,7 +13,7 @@ def main():
     parser.add_argument("--version", type=str, default="")
     args = parser.parse_args()
     api = HfApi()
-    model_prefix = f"{TEST_HUB_ORG}/neuron-testing-{args.version}"
+    model_prefix = f"{TEST_HUB_ORG}/optimum-neuron-testing-{args.version}"
     models = api.list_models(search=model_prefix)
     for model in models:
         if not model.id.startswith(model_prefix):


optimum/neuron/utils/instance.py
```diff
@@ -0,0 +1,89 @@
...
+        current_target = current_instance_type()
+        if target == current_target:
+            # No need to override the compilation target as it matches the current instance type
+            return current_target
+        elif not override:
+            raise ValueError(
+                f"The current platform is {current_target} but we are compiling for {target}."
+                f" Please set the NEURON_PLATFORM_TARGET_OVERRIDE to {target} or use the optimum-cli"
+            )
+        logger.info(f"The current instance type is {current_target}, but we are compiling for {target}.")
```
JingyaHuang, line 85: So this is ok? no matter what instance we are using, as far as we set up the target instance before importing nxd, we are about to compile for that target?

optimum/neuron/__init__.py
```diff
@@ -19,16 +19,17 @@
...
 
 
 if get_neuron_major() == -1:
-    warnings.warn(
-        "It seems this is running on a CPU-only machine,"
-        " so we override the platform target to trn1, otherwise there will"
-        " be errors when importing torch_neuronx."
-    )
-    os.environ["NEURON_PLATFORM_TARGET_OVERRIDE"] = "trn1"
+    if align_compilation_target("trn1", override=False) == "trn1":
```
JingyaHuang, line 27: So now the not working case would be: on cpu-only instance, if we are targeting trn2, and the env var `NEURON_PLATFORM_TARGET_OVERRIDE` is not set, then we won't be able to export models with Neuron model class right?

optimum/neuron/version.py
```diff
@@ -12,6 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.3.1.dev7"
+__version__ = "0.4.0.dev0"
```
JingyaHuang, line 15: we are about to do a release?

tests/decoder/test_cli.py
```diff
@@ -12,8 +12,8 @@
 @is_inferentia_test
 @requires_neuronx
 @pytest.mark.parametrize("batch_size, sequence_length, auto_cast_type", [[1, 512, "fp16"], [2, 128, "bf16"]])
-@pytest.mark.parametrize("tensor_parallel_size", [1, 2])
-def test_export_decoder_cli(batch_size, sequence_length, auto_cast_type, tensor_parallel_size):
+@pytest.mark.parametrize("instance_type", ["trn1", "trn2"])
```
JingyaHuang, line 15: is it necessary to test on cpu instance, but I don't feel it is necessary to test trn2 compilation on inf2 instance no?

JingyaHuang (commented):
minor nits and questions, otherwise lgtm, thanks!

optimum/neuron/utils/instance.py
dacorvo: Yes, this is correct. But it is very difficult to avoid importing nxd when importing neuron symbols, so really this is only feasible in the `optimum-cli` or as the first instruction in a script.

optimum/neuron/__init__.py
dacorvo: If we are using the `optimum-cli` it will be set immediately during the execution of the run command, so it should be ok. Otherwise yes, the neuron runtime will be initialized by default for `trn1` and we won't be able to change it afterwards.

optimum/neuron/version.py
dacorvo: I thought that adding support for trainium 2 deserved an increase in the minor version. And yes, I think we should release as soon as we have validated trn2 support (but we need to be able to test it first).

tests/decoder/test_cli.py
dacorvo: From a development perspective, when adding a new model, it is convenient to be able to test the `trn2` compilation from the platform where you are validating the modeling, and it will primarily be a `trn1` instance in the coming months. That's why I added this test (it is short, and I removed two combinations with tp 1 and 2 to not increase the amount of tests).

JingyaHuang: said that considering the CI workload, if I add that for diffusion and trfrs encoder models, our CIs might explode.

optimum/neuron/models/inference/backend/config.py
```diff
@@ -159,3 +156,7 @@ def weights_to_skip_layout_optimization(self) -> list[str]:
         Can be overridden by subclasses to specify weights that should not be optimized.
         """
         return []
+
+    @property
+    def logical_nc_config(self) -> int:
```
tengomucho, line 161: what is the nc config?

tengomucho (approved):

JingyaHuang (approved):

tests/decoder/test_cli.py
dacorvo: Sure, even for decoders it is overkill, but here I am testing only two configurations of a single tiny model, so it is just two tests for `trn1` and two tests for `trn2`.

optimum/neuron/models/inference/backend/config.py
dacorvo: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/arch/neuron-features/logical-neuroncore-config.html

So basically we set it to the `trn2` default, which is to group physical cores by pairs inside each device: this makes sense since each device has 8 cores but only 4 memory banks, each of 24 GB.
If we set the `nc` config to 1, then we will get more cores, but each with less memory. Also, we will need to export a specific configuration variable to modify the behaviour of the runtime.

dacorvo: With that setting, a `trn2` instance will expose `64` cores with `24 GB` of memory each.

PR #990, opened 2025-10-08
Run the LLM inference cache CI job on cpu

dacorvo:
# What does this PR do?

Since we have validated in #988 that we could export LLM models on CPU, we don't need to use the `inf2` instances anymore.


dacorvo:
Need to check if the generated artifacts are actually indentical as those generated on a `trn1/inf2` platform.

tengomucho (approved):

PR #1004, opened 2025-10-17
add `--served_model_name` to vLLM

tengomucho:
# What does this PR do?

Added "--served_model_name" argument usage when serving LLMs with vLLM. This feature is used in inference endpoints, so the model parameter to vLLM s going to be a local directory containing the model weights, and the served-model-name will include the model_id. This is used by Optimum Neuron's vLLM to determine the neuron config.

### DIFF
--- a/tests/vllm/service/test_vllm_service_generate.py
+++ b/tests/vllm/service/test_vllm_service_generate.py
@@ -69,10 +69,9 @@ async def vllm_service_from_model(request, vllm_launcher, base_neuron_llm_config
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize(
-    "prompt, max_output_tokens", [("What is Deep Learning?", 17), ("What is the colour of the sky ?", 24)]
-)
-async def test_vllm_service_from_model(vllm_service_from_model, prompt, max_output_tokens):
+async def test_vllm_service_from_model(vllm_service_from_model):
+    prompt = "What is Deep Learning?"
+    max_output_tokens = 17
     greedy_tokens, greedy_text = await vllm_service_from_model.client.greedy(
         prompt, max_output_tokens=max_output_tokens
     )
--- a/tests/vllm/docker/test_vllm_docker_service_generate.py
+++ b/tests/vllm/docker/test_vllm_docker_service_generate.py
@@ -61,14 +61,9 @@ async def vllm_docker_service_from_model(request, vllm_docker_launcher, base_neu
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize(
-    "prompt, max_output_tokens",
-    [
-        ("What is Deep Learning?", 17),
-        ("What is the colour of the sky ?", 24),
-    ],
-)
-async def test_vllm_docker_service_from_model(vllm_docker_service_from_model, prompt, max_output_tokens):
+async def test_vllm_docker_service_from_model(vllm_docker_service_from_model):
+    prompt = "What is the colour of the sky ?"
+    max_output_tokens = 24
     greedy_tokens, greedy_text = await vllm_docker_service_from_model.client.greedy(
         prompt, max_output_tokens=max_output_tokens
     )


optimum/commands/neuron/serve.py
```diff
@@ -83,22 +89,26 @@ def parse_args(parser: "ArgumentParser"):
...
+            model_id = model_name_or_path
         instance_type = current_instance_type()
         batch_size = self.args.batch_size
         sequence_length = self.args.sequence_length
         tensor_parallel_size = self.args.tensor_parallel_size
-        config = AutoConfig.from_pretrained(model_id, revision=revision)
+        try:
+            config = AutoConfig.from_pretrained(model_name_or_path)
+        except Exception:
+            config = AutoConfig.from_pretrained(model_id)
```
dacorvo: I think we should not try to fallback: if the config has not been stored locally, then we should refuse to export the model.

optimum/commands/neuron/serve.py
```diff
@@ -83,22 +89,26 @@ def parse_args(parser: "ArgumentParser"):
     @requires_vllm
     @requires_torch_neuronx
     def run(self):
-        model_id = self.args.model
-        revision = None
+        model_name_or_path = self.args.model
+        model_id = self.args.served_model_name
+        if model_id is None:
+            model_id = model_name_or_path
         instance_type = current_instance_type()
```
dacorvo, line 99: Maybe add a log if `model_name_or_path` and `model_id` are different.

optimum/commands/neuron/serve.py
```diff
@@ -124,7 +134,7 @@ def run(self):
                     f"The specified tensor parallel size {tensor_parallel_size} is inconsistent"
                     f"with the one used to export the neuron model ({neuron_config.tp_degree})"
                 )
-            logger.info(f"Loading Neuron model: {self.args.model}")
+            logger.info(f"Loading Neuron model: {model_id}")
```
dacorvo: It should be `model_name_or_path` because it is an already exported model.

tests/fixtures/llm/vllm_docker_service.py
```diff
@@ -134,10 +139,13 @@ def docker_launcher(
                 model_path = os.path.join(context_dir, "model")
                 shutil.copytree(model_name_or_path, model_path)
                 # Create Dockerfile
-                container_model_id = f"/data/{model_name_or_path}"
+                container_model_name_or_path = f"/data{model_name_or_path}"
```
dacorvo: I understand why you did it, but if the caller does not provide an absolute path, it will create a different directory tree

tests/fixtures/llm/vllm_docker_service.py
```diff
@@ -134,10 +139,13 @@ def docker_launcher(
...
+                if served_model_name == model_name_or_path:
```
dacorvo: I don't think you need this, since you always pass `served_model_name`.

tests/fixtures/llm/vllm_service.py
```diff
@@ -36,10 +36,11 @@ class AsyncOpenAI:
 
 
 class TestClient(AsyncOpenAI):
-    def __init__(self, service_name: str, model_name: str, port: int):
+    def __init__(self, service_name: str, model_name: str, served_model_name: str, port: int):
         super().__init__(api_key="EMPTY", base_url=f"http://localhost:{port}/v1")
         self.model_name: str = model_name
         self.service_name: str = service_name
+        self.served_model_name: str = served_model_name
```
dacorvo: You don't need to have both a `model_name` and a `served_model_name` member: just pass either one of them when instantiating the client.

tests/fixtures/llm/vllm_service.py
```diff
@@ -66,8 +67,8 @@ async def greedy(self, prompt: str, max_output_tokens: int, stop: List[str] | No
 
 
 class LauncherHandle:
-    def __init__(self, service_name: str, model_name: str, port: int):
-        self.client = TestClient(service_name, model_name, port)
+    def __init__(self, service_name: str, model_name: str, served_model_name: str, port: int):
+        self.client = TestClient(service_name, model_name, served_model_name, port)
```
dacorvo: Here also, you could simply have passed `model_name=served_model_name`

tests/fixtures/llm/vllm_docker_service.py
```diff
@@ -163,7 +171,14 @@ def docker_launcher(
         )
 
         logger.info(f"Starting {container_name} container")
-        yield ContainerLauncherHandle(service_name, container_model_id, client, container.name, port)
+        yield ContainerLauncherHandle(
+            service_name,
+            container_model_name_or_path,
```
dacorvo: You don't need to pass both values: you could for instance pass something like: `container_model_name_or_path if served_model_name is None else served_model_name`.

tests/vllm/service/test_vllm_service_generate.py
```diff
@@ -15,7 +18,7 @@ async def multi_model_vllm_service(vllm_launcher, neuron_llm_config):
 
 
 @pytest.mark.asyncio
-@pytest.fixture(params=["local_neuron", "hub_neuron", "hub_explicit", "hub_implicit"])
+@pytest.fixture(params=["local_neuron", "hub_neuron", "hub_explicit", "hub_implicit", "local_explicit"])
```
dacorvo: It is actually local_implicit, since you don't specify any export kwargs

tests/vllm/service/test_vllm_service_generate.py
```diff
@@ -34,17 +37,34 @@ async def vllm_service_from_model(request, vllm_launcher, base_neuron_llm_config
...
-        elif request.param == "hub_neuron":
-            model_name_or_path = base_neuron_llm_config["neuron_model_id"]
-        elif request.param == "hub_implicit":
-            model_name_or_path = base_neuron_llm_config["model_id"]
-        else:
-            raise ValueError(f"Unknown request.param: {request.param}")
-        with vllm_launcher(service_name, model_name_or_path) as vllm_service:
-            await vllm_service.health(600)
-            yield vllm_service
+        # Note: the local model path is only used for the local_explicit case
```
dacorvo: Then you should create it only when it is required.

tests/fixtures/llm/vllm_service.py
```diff
@@ -146,6 +147,10 @@ def launcher(
             "--port",
             str(port),
         ]
+        if served_model_name is not None:
+            command += ["--served_model_name", served_model_name]
+        else:
```
dacorvo: I think this else is not required.

optimum/commands/neuron/serve.py
```diff
@@ -83,22 +89,26 @@ def parse_args(parser: "ArgumentParser"):
...
     def run(self):
-        model_id = self.args.model
-        revision = None
+        model_name_or_path = self.args.model
+        model_id = self.args.served_model_name
+        if model_id is None:
+            model_id = model_name_or_path
+        else:
+            if model_name_or_path != model_id:
+                logger.info(f"Serving model {model_id} with at path {model_name_or_path}")
```
dacorvo:
```suggestion
                logger.info(f"Serving model {model_id} at path {model_name_or_path}")
```

tests/vllm/service/test_vllm_service_generate.py
```diff
@@ -34,24 +37,40 @@ async def vllm_service_from_model(request, vllm_launcher, base_neuron_llm_config
...
-        elif request.param == "hub_neuron":
-            model_name_or_path = base_neuron_llm_config["neuron_model_id"]
-        elif request.param == "hub_implicit":
-            model_name_or_path = base_neuron_llm_config["model_id"]
-        else:
-            raise ValueError(f"Unknown request.param: {request.param}")
-        with vllm_launcher(service_name, model_name_or_path) as vllm_service:
-            await vllm_service.health(600)
-            yield vllm_service
+        # Note: the local model path is only used for the local_implicit case
```
dacorvo: Change here also please 🙏

PR #1007, opened 2025-10-22
NxD backend refactoring

dacorvo:
# What does this PR do?

This rewrites the class hierarchy in the NxD backend used for inference custom modeling. The goal is to allow more model architectures to be supported by subclassing the main `NxDPretrainedModel` class.


tengomucho (approved):
LGTM

PR #1010, opened 2025-10-23
ci(vllm): login to docker

tengomucho:
# What does this PR do?

Add login to docker to prevent errors when building the image.

See this failure for example: https://github.com/huggingface/optimum-neuron/actions/runs/18754605681/job/53503150736

It can be avoided using this login.


JingyaHuang (approved):
LGTM.

PR #1012, opened 2025-10-24
feat: Add `encode` and `similarity` of Sentence transformers

### LINKED ISSUE #975: Add support for encode and similarity for sentence transformer models
### Feature request

Am attempting to run the Qwen3 embedding model in a similar way to the code outlined on the Qwen3-embedding model card. When running with ON it is missing the 2 methods `encode` and `similarity`. Can we get them added so the code snippet below can run which is nearly identical to running with normal sentence transformers code ? 

```
from optimum.neuron import NeuronModelForSentenceTransformers

# Load the model
model = NeuronModelForSentenceTransformers("Qwen/Qwen3-Embedding-0.6B")

# The queries and documents to embed
queries = [
    "What is the capital of China?",
    "Explain gravity",
]
documents = [
    "The capital of China is Beijing.",
    "Gravity is a force that attracts two bodies towards each other. It gives weight to physical objects and is responsible for the movement of planets around the sun.",
]

# Encode the queries and documents. Note that queries benefit from using a prompt
# Here we use the prompt called "query" stored under `model.prompts`, but you can
# also pass your own prompt via the `prompt` argument
query_embeddings = model.encode(queries, prompt_name="query")
document_embeddings = model.encode(documents)

# Compute the (cosine) similarity between the query and document embeddings
similarity = model.similarity(query_embeddings, document_embeddings)
print(similarity)
```

### Motivation

Customers want to run embedding models on Trainium and we want to enable them to do so in a similar way to the sentence transformers API

### Your contribution

Can test and give feedback

JingyaHuang:
# What does this PR do?

Fixes #975 

- [x] Add `encode()` and `similarity()`
- [x] tests
- [x] new doc for sentence transformers

## Before submitting

### DIFF
--- a/optimum/commands/neuron/cache.py
+++ b/optimum/commands/neuron/cache.py
@@ -147,7 +147,7 @@ class LookupRepoCommand(BaseOptimumCLICommand):
                     str(entry["batch_size"]),
                     str(entry["sequence_length"]),
                     str(entry.get("tp_degree", entry.get("tensor_parallel_size"))),
-                    str(entry["torch_dtype"]),
+                    str(entry.get("torch_dtype", entry.get("dtype"))),
                     str(entry["target"]),
                 )
             )


optimum/neuron/pipelines/transformers/base.py
```diff
@@ -171,7 +171,7 @@ def load_pipeline(
             model, token=token, revision=revision
         ):
             logger.info("Using Sentence Transformers compatible Feature extraction pipeline")
-            neuronx_model_class = NeuronModelForSentenceTransformers
+            neuronx_model_class = NeuronSentenceTransformers
```
michaelbenayoun, line 174: nit: what about calling this class `NeuronSentenceTransformersModel` to be aligned with the naming.

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+            processor_class=_TOKENIZER_FOR_DOC,
+            model_class="NeuronModelForSentenceTransformers",
+            checkpoint="optimum/bge-base-en-v1.5-neuronx",
+        )
+        + NEURON_SENTENCE_TRANSFORMERS_IMAGE_EXAMPLE.format(
+            processor_class=_GENERIC_PROCESSOR,
+            model_class="NeuronModelForSentenceTransformers",
+            checkpoint="optimum/clip_vit_emb_neuronx",
+        )
+    )
```
michaelbenayoun: Do we need to add all the docstrings?
I personally do not do it but I guess it is a matter of taste.

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        pixel_values: torch.Tensor | None = None,
+        token_type_ids: torch.Tensor | None = None,
+        **kwargs,
```
michaelbenayoun: Why having kwargs? It's not used anywhere in the method.

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+        Args:
+            texts (Union[List[str], List[Dict], List[Tuple[str, str]]]): A list of texts to be tokenized.
+
+        Returns:
+            Dict[str, Tensor]: A dictionary of tensors with the tokenized texts. Common keys are "input_ids",
+                "attention_mask", and "token_type_ids".
+        """
+        return self.preprocessors[0](texts, **kwargs)
+
+    def _get_prompt_length(self, prompt: str, **kwargs) -> int:
```
michaelbenayoun:
```suggestion
    def _get_prompt_length(self, prompt: str, **kwargs) -> int | None:
```

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+                    sentence_embedding=sentence_embedding,
+                    attention_mask=attention_mask,
+                )
+
+    def tokenize(self, texts: list[str] | list[dict] | list[tuple[str, str]], **kwargs) -> dict[str, torch.Tensor]:
+        """
+        Tokenizes the texts.
+
+        Args:
+            texts (Union[List[str], List[Dict], List[Tuple[str, str]]]): A list of texts to be tokenized.
```
michaelbenayoun:
```suggestion
            texts (list[str] | list[dict] | list[tuple[str, str]]]): A list of texts to be tokenized.
```

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+
+        Returns:
+            Dict[str, Tensor]: A dictionary of tensors with the tokenized texts. Common keys are "input_ids",
```
michaelbenayoun: Same comment. 
It's a nit but basically align the doc type annotation with the python type annotation if possible.

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+        return prompt_length
+
+    def _text_length(self, text: list[int] | list[list[int]]) -> int:
+        """
+        Help function to get the length for the input text. Text can be either
+        a list of ints (which means a single text as input), or a tuple of list of ints
+        (representing several text inputs to the model).
+        """
+
+        if isinstance(text, dict):  # {key: value} case
```
michaelbenayoun, line 139: If `text` can be a `dict` update the type annotation in the function signature.

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+            return len(next(iter(text.values())))
+        elif not hasattr(text, "__len__"):  # Object has no len() method
+            return 1
+        elif len(text) == 0 or isinstance(text[0], int):  # Empty string or list of ints
+            return len(text)
+        else:
+            return sum([len(t) for t in text])  # Sum of length of individual strings
+
+    @property
+    def similarity_fn_name(self) -> Literal["cosine", "dot", "euclidean", "manhattan"]:
```
michaelbenayoun, line 149:
```suggestion
    def similarity_fn_name(self) -> SimilarityFunction:
```

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+            >>> model.similarity_fn_name
+            'dot'
+        """
+        if self._similarity_fn_name is None:
+            self.similarity_fn_name = SimilarityFunction.COSINE
+        return self._similarity_fn_name
+
+    @similarity_fn_name.setter
+    def similarity_fn_name(
+        self, value: Literal["cosine", "dot", "euclidean", "manhattan"] | SimilarityFunction
```
michaelbenayoun:
```suggestion
        self, value: Literal["cosine", "dot", "euclidean", "manhattan"] | SimilarityFunction | None
```

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+    ) -> None:
+        if isinstance(value, SimilarityFunction):
+            value = value.value
+        self._similarity_fn_name = value
+
+        if value is not None:
+            self._similarity = SimilarityFunction.to_similarity_fn(value)
+            self._similarity_pairwise = SimilarityFunction.to_similarity_pairwise_fn(value)
```
michaelbenayoun, line 175: We set the function name even when `value` is `None`?

optimum/neuron/modeling_sentence_transformers.py
```diff
@@ -0,0 +1,311 @@
...
+            sentences = [prompt + sentence for sentence in sentences]
+
+            # Some models (e.g. INSTRUCTOR, GRIT) require removing the prompt before pooling
+            # Tracking the prompt length allow us to remove the prompt during pooling
+            length = self._get_prompt_length(prompt, **kwargs)
+            if length is not None:
+                extra_features["prompt_length"] = length
+
+        all_embeddings = []
+        length_sorted_idx = np.argsort([-self._text_length(sen) for sen in sentences])
```
michaelbenayoun: Do we need to use np here?

optimum/neuron/pipelines/transformers/base.py
JingyaHuang: Yeah I hesitated, what you suggested correspond to the naming method in transformers, however in sentence transformers library the class is `SentenceTransformers`, I think it makes more sense to just add the Neuron prefix.

optimum/neuron/modeling_sentence_transformers.py
JingyaHuang: agreed, actually the docstring repeat quite alot what we already have in the sentence transformers doc, will remove it.

optimum/neuron/modeling_sentence_transformers.py
JingyaHuang: why? I think in the setter `self._similarity` will always be `str`.

optimum/neuron/modeling_sentence_transformers.py
JingyaHuang: if value is None then `similarity_fn_name` will be `SimilarityFunction.COSINE` by default. This is the way how it was setup in sentence transformers.
https://github.com/huggingface/optimum-neuron/blob/a34124c08100b9368fd4f892f6d4504bb9da2f1a/optimum/neuron/modeling_sentence_transformers.py#L191

optimum/neuron/modeling_sentence_transformers.py
JingyaHuang: nope, we can use pytorch, it's how it was in sentence transformers, will change it.

optimum/neuron/modeling_sentence_transformers.py
JingyaHuang: ~removed!~

optimum/neuron/modeling_sentence_transformers.py
michaelbenayoun: If `self._similarity_fn_name` is `None`, then you return `SimilarityFunction.COSINE`.
But that's a nit I guess, it's equivalent right?

JingyaHuang: Yeah in that case it returns `SimilarityFunction.COSINE` which is just the str "cosine" which is not a function but its name:
https://github.com/huggingface/sentence-transformers/blob/85ec64559f4414aa536eca4bf53538291e0a333f/sentence_transformers/similarity_functions.py#L31C14-L31C22

PR #1018, opened 2025-11-03
Bump optimum to 2.0

### LINKED ISSUE #1015: Upgrade Optimum version to 2.0.0
### Feature request

Following the recent Optimum 2.0.0 release, Optimum Neuron shall also upgrade the dependent Optimum version (currently pinned to 1.24.0)

### Motivation

* One motivation is that we got reported issues like: https://github.com/huggingface/optimum-neuron/issues/993, unrelated import of other backends caused issue.
* And we want to upgrade the python version since all AWS Neuron PyTorch base DLCs use Python 3.11 now: https://github.com/aws-neuron/deep-learning-containers/blob/main/pytorch/training/2.8.0/Dockerfile.neuronx

### Your contribution

I can work on this

cc. @tengomucho @michaelbenayoun @dacorvo

JingyaHuang:
# What does this PR do?

Fixes #1015 

Optimum released a major 2.0 version which detached some legacy code for backends unrelated to Neuron. The PR aims at supporting Optimum 2.0 and cleaning deprecated code.

### DIFF
--- a/tests/inference/transformers/test_modeling.py
+++ b/tests/inference/transformers/test_modeling.py
@@ -45,9 +45,6 @@ from transformers import (
     PretrainedConfig,
     set_seed,
 )
-from transformers import (
-    __version__ as transformers_version,
-)
 
 from optimum.neuron import (
     NeuronModelForAudioClassification,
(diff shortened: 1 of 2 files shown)


JingyaHuang:
Thanks @dacorvo, will fix the issue in CI. For the init file of neuron, as explained @IlyasMoutawwakil here: https://github.com/huggingface/optimum/pull/2361#issuecomment-3351837184, it seems that we don't need to remove it, the change only need to be done on the top level of Optimum main pkg.

tests/inference/transformers/test_modeling.py
```diff
@@ -1046,11 +1014,6 @@ def test_compare_to_transformers_non_dyn_bs(self, model_arch):
             "model_arch": model_arch,
             "dynamic_batch_size": False,
         }
-        # REMOVEME: convnextv2 contains a bug in the GRN layer, which is used in the convnextv2 model, but the bug has
```
tengomucho, line 1049: thanks for this! I think you can also remove the `skipTest` in the following test, `test_compare_to_transformers_dyn_bs`, as well as in the `_neuronx_export` function in `test_transformers.py`.

dacorvo:
> Thanks @dacorvo, will fix the issue in CI. For the init file of neuron, as explained @IlyasMoutawwakil here: [huggingface/optimum#2361 (comment)](https://github.com/huggingface/optimum/pull/2361#issuecomment-3351837184), it seems that we don't need to remove it, the change only need to be done on the top level of Optimum main pkg.

So I was mistaken 😉

PR #1020, opened 2025-11-04
GRPO Trainer

michaelbenayoun:
# What does this PR do?

This PR adds partial support for GRPO.

It was broken down into smaller PRs:

- #1040 
- #1042 
- #1043 
- #1044 

It adds the `NeuronGRPOTrainer` with a set of optimizations and modifications for the Torch XLA backend used to run things on Trainium instances. There are still core missing features:

- Integration with vLLM: we use a custom CPU vLLM hack for now. The plan is to work on the vLLM part on another PR.
- Weight Synchronization `NeuronGRPOTrainer <-> vLLM`
- No tensor parallelism


optimum/neuron/trainers/grpo_config.py
```diff
@@ -0,0 +1,118 @@
...
+            "generation instead of the default model.generate(). Requires `vllm` to be installed. Required for NeuronGRPOTrainer."
+        },
+    )
+
+    def __post_init__(self):
+        if not self.experimental:
+            raise ValueError(
+                "NeuronGRPOTrainer is experimental and not production-ready. To proceed, set `experimental=True` in "
+                "your NeuronGRPOConfig. This flag exists to ensure users are aware of the current state of the implementation."
+            )
```
michaelbenayoun, line 63: For now we disable the access to the `NeuronGRPOTrainer`

.github/actions/install_optimum_neuron/action.yml
```diff
@@ -7,4 +7,4 @@ runs:
         shell: bash
         run: |
           source aws_neuron_venv_pytorch/bin/activate
-          python -m pip install .[neuronx,tests]
+          python -m pip install .[neuronx,tests,training]
```
dacorvo: You should not install training requirements for all workflows:
- this can create conflicts 
- this will hide any imports errors

dacorvo (changes requested):
The change in the CI workflow aside, this looks good to me, although I did not go into the details of the trainer algorithm.
For the next step, you will need to add a `load_state_dict` method in `NxDPretrainedModel`, as the existing `load_weights` method reads weights from a path. You will need to provide also modified `checkpoint_loader_fn` to load weights from a state_dict directly as for now it also expects to load the state_dict from a path.
Then in vLLM you will need to call the `load_state_dict` method when required.

dacorvo:
@michaelbenayoun can you rebase or merge the main branch, it looks like you are using the non-working sanity changes that @tengomucho fixed in #1077. As a consequence, the training tests are not launched.

optimum/neuron/trainers/transformers.py
```diff
@@ -1226,7 +1238,7 @@ def _train(
                 self.control.should_training_stop = True
 
             self.control = self.callback_handler.on_epoch_end(args, self.state, self.control)
-            xm.mark_step()
+            torch_xla.sync()
```
tengomucho, line 1241: I think `mark_step` is deprecated, so this is a good change, but don't you think it would be better to change all occurrences of this?

tengomucho (commented):
It is hard for me to follow exactly what this allows, so I will have to trust you on this.
I just ask for the sanity checks to be run first, as mentioned by @dacorvo 
Generally speaking, if you add a code feature, it would be better to add a test for it, don't you think?

michaelbenayoun:
> It is hard for me to follow exactly what this allows, so I will have to trust you on this. I just ask for the sanity checks to be run first, as mentioned by @dacorvo Generally speaking, if you add a code feature, it would be better to add a test for it, don't you think?

For the test we will do it once it's fully covered because for now the code runs with a CPU vLLM env.
Also in the sub PRs we have added some tests.

optimum/neuron/trainers/transformers.py
michaelbenayoun: You mean in the whole library?

In tests it is still existing, but in the library, the only occurence is in `optimum/neuron/generation/utils.py, I will update it.

michaelbenayoun:
For now not merging with the recent progress in Torch Native.

PR #1022, opened 2025-11-05
fix: Flux timeout issue + nxd implementation refactoring

JingyaHuang:
# What does this PR do?

* Fix the t5 compilation timeout by switching to ModelBuilder
* Improvement with kernels
* Put flux tests back

- Caveat:
Not optimized for trn2 (some kernels removed due to torchscript interpreter error on inf2)

### DIFF
--- a/.github/workflows/test_inf2_export.yml
+++ b/.github/workflows/test_inf2_export.yml
@@ -66,4 +66,4 @@ jobs:
         run: |
           source aws_neuron_venv_pytorch/bin/activate
           export RANDOM_TASK_PER_MODEL=TRUE
-          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest tests/exporters --durations=0 --timeout=300
+          HF_TOKEN=${{ secrets.HF_TOKEN_OPTIMUM_NEURON_CI }} pytest tests/exporters --ignore=tests/exporters/test_cpu_compilation.py --durations=0 --timeout=300


optimum/exporters/neuron/model_configs.py
```diff
@@ -928,6 +891,23 @@ def generate_dummy_inputs(self, return_tuple: bool = False, **kwargs):
...
 
+    def get_compiler_args(self):
+        compiler_args = "--model-type=transformer -O1"
+        self.context_parallel_enabled = self.world_size != self.tensor_parallel_size
+        if self.context_parallel_enabled and _HARDWARE == hardware.TRN1:
+            compiler_args = "--model-type=transformer -O2"
+        if self.context_parallel_enabled and _HARDWARE == hardware.TRN2:
+            compiler_args += " --tensorizer-options='--enable-ccop-compute-overlap'"
+        else:
+            compiler_args += " --tensorizer-options='--enable-ccop-compute-overlap --cc-pipeline-tiling-factor=4'"
```
tengomucho, line 897: on what hw is this applied?

optimum/neuron/models/inference/flux/flux_transformer_2d/modules/kernels.py
```diff
@@ -0,0 +1,79 @@
...
+        "llama3_transfomer_fwd_<tp|rmsnorm_sp> only supports no specialization or specialization along one axis"
+    )
+
+    global n_prgs, prg_id
+    if grid_ndim != 0 and nl.num_programs(axes=0) > 1:
+        n_prgs = nl.num_programs(axes=0)
+        prg_id = nl.program_id(axis=0)
+
+
+@nki.jit
```
tengomucho, line 48: did you write these kernels or did you import them? If you imported, mention the source, so it will be easier to maintain them in the future.

tests/inference/transformers/test_modeling.py
```diff
@@ -437,6 +435,7 @@ class NeuronModelForMaskedLMIntegrationTest(NeuronModelTestMixin):
         "xlm",
         "xlm-roberta",
     ]
+    ONE_ARCH = [random.choice(SUPPORTED_ARCHITECTURES)]
```
tengomucho: why I understand that testing over all architectures is long, I do not think choosing one randomly like this is a good idea: if something goes wrong it might be difficult to trace back what was tested and the random choice is only done once for all tests. What about having a marker so that tests run always on the same arch, except on nightly when they can run on all architectures?

optimum/exporters/neuron/model_configs.py
JingyaHuang: 1st condition is applied to trainium 2 when context paralleled is enabled. fyi, these trn2 specific features are not tested, I just put it as placeholder.

tests/inference/transformers/test_modeling.py
JingyaHuang: yeah make sense, will add a marker and a nightly workflow.

optimum/exporters/neuron/model_configs.py
```diff
@@ -928,6 +891,23 @@ def generate_dummy_inputs(self, return_tuple: bool = False, **kwargs):
         else:
             return dummy_inputs
 
+    def get_compiler_args(self):
```
dacorvo: The logic to evaluate the hardware is duplicated, and partially redundant with the target that we should have selected beforehand. Could you consider reusing this information ?

optimum/exporters/neuron/convert.py
```diff
@@ -654,6 +659,62 @@ def prepare_compiler_flags(
...
 
 
+def init_custom_process_group_fn(config):
+    if hasattr(config, "fused_spec_config") and config.fused_spec_config is not None:
+        if config.fused_spec_config.draft_config.neuron_config.tp_degree is not None:
+            draft_tp = config.fused_spec_config.draft_config.neuron_config.tp_degree
+            parallel_state.initialize_speculative_draft_group(draft_tp)
+
+
+def get_builder(
```
dacorvo, line 669: This duplicates the code in `models/inference/backend/pretrained_model.py`. I tried to make this class generic enough so that it could be used for any model type: what prevented you from using it here ?

optimum/exporters/neuron/config.py
```diff
@@ -101,3 +105,85 @@ def _create_dummy_input_generator_classes(self, **kwargs) -> list["DummyInputGen
         ]
 
         return dummy_inputs_generators
+
+
+class NxDNeuronConfig:
```
dacorvo: This name overlaps with the `NxDNeuronConfig` already defined in the nxd backend (which is an actual NeuronConfig). Its name is also misleading, because it is not a config class, but an Exporter class (it implements all methods required to export the model in the traced code path).
Can you please rename it (at least to some name that has Export in it).

optimum/neuron/models/inference/backend/utils/layer_boundary_marker.py
```diff
@@ -0,0 +1,78 @@
...
+from torch import nn
+from torch_neuronx.xla_impl.base import xla_hlo_call
+from torch_neuronx.xla_impl.custom_call_targets import (
+    AwsNeuronModuleMarkerEndForward,
+    AwsNeuronModuleMarkerStartForward,
+)
+
+
+# does not support nesting since we do not pass config str
+class ModuleMarkerStart(torch.autograd.Function):
```
dacorvo, line 26: Unless I am mistaken, this is related to training, so we don't need this for inference modeling

optimum/neuron/models/inference/flux/flux_transformer_2d/modeling_flux_transformer_2d.py
```diff
@@ -107,81 +126,76 @@ class NeuronFluxTransformer2DModel(torch.nn.Module):
...
-            guidance_embeds=guidance_embeds,
-        )
+        self.config = config
+
+        self.data_parallel_group = get_data_parallel_group()
+        self.global_rank = SPMDRank(world_size=get_world_group().size())
 
-        self.out_channels = out_channels or in_channels
-        self.inner_dim = self.config.num_attention_heads * self.config.attention_head_dim
+        # TODO: context parallel and out_proj_kernel are not working correctly so far
```
dacorvo: I don't think we should include code paths that do not work from NxDI, as this package will not be updated anymore, so we will never be able to fix this.

optimum/neuron/models/inference/flux/flux_transformer_2d/modeling_flux_transformer_2d.py
```diff
@@ -107,81 +126,76 @@ class NeuronFluxTransformer2DModel(torch.nn.Module):
...
-        num_layers: int = 19,
-        num_single_layers: int = 38,
-        attention_head_dim: int = 128,
-        num_attention_heads: int = 24,
-        joint_attention_dim: int = 4096,
-        pooled_projection_dim: int = 768,
-        guidance_embeds: bool = False,
-        axes_dims_rope: list[int] = [16, 56, 56],
-        reduce_dtype: torch.dtype = torch.bfloat16,
+        config: "NeuronDefaultConfig",
```
dacorvo: Is this change required ? It makes the code harder to read since you need to access these parameters through config._config. Could you drop one level of indirection ?

optimum/neuron/utils/testing_utils.py
```diff
@@ -51,3 +51,13 @@ def is_inferentia_test(test_case):
         return test_case
     else:
         return pytest.mark.is_inferentia_test()(test_case)
+
```
dacorvo, line 54: I don't think we should have any of this in the package itself, but only in the fixtures or pytest config. But that is probably for another pull-request 😉

dacorvo (commented):
Thank you for this pull-request. I understand you had a hard-time already making this work, but maybe there is a missed opportunity of using a least some of the code already present under `models/inference/backend`.
I made some comments about a few things that I think might be this code and the package in general harder to maintain.

tests/inference/transformers/test_modeling.py
```diff
@@ -387,7 +387,7 @@ def test_sentence_transformers_clip(self, model_arch):
 
         texts = ["Two dogs in the snow", "A cat on a table", "A picture of London at night"]
         util.http_get(
-            "https://github.com/UKPLab/sentence-transformers/raw/master/examples/sentence_transformer/applications/image-search/two_dogs_in_snow.jpg",
+            "https://github.com/UKPLab/sentence-transformers/raw/main/examples/sentence_transformer/applications/image-search/two_dogs_in_snow.jpg",
```
tengomucho: even better: `https://raw.githubusercontent.com/huggingface/sentence-transformers/main/examples/sentence_transformer/applications/image-search/two_dogs_in_snow.jpg`

optimum/neuron/models/inference/backend/utils/layer_boundary_marker.py
JingyaHuang: I don't think so, the backward is not implemented so these should target inference.  They are used inside the flux transformer code, seems to be some sorts of marker for subgraphs used during the compilation.

optimum/exporters/neuron/convert.py
JingyaHuang: The `get_builder` in `optimum/neuron/models/inference/backend/pretrained_model.py` requires `neuron_config` and `graph_builders` as inputs which are not used in the exporter logic. I would rather have another `get_builder`(if it's confusing I could rename it) than crafting neuron_config and graph_builders just to reuse the function.

optimum/neuron/utils/testing_utils.py
JingyaHuang: yeah, I’ve been too lazy to tidy up the old inference tests😿, hope will get to it one day.

optimum/exporters/neuron/convert.py
dacorvo: But NxDGraphBuilder is just an abstract API to return model instance and example inputs that you could easily have implemented, since you pass them as parameters. Neuron config is just a dataclass, so you could have used it also, even if you just use tensor_parallel_size. This would have helped also transitioning to trn2 later on, instead of hard-coding lnc for instance.
Anyway, all this code is eventually going to be deprecated in a few months, so not a big deal.

JingyaHuang: I will leave this like this in this PR and will keep it in mind when enabling it for trn2!

PR #1028, opened 2025-11-19
Fix input slots exhaustion in vLLM plugin

dacorvo:
# What does this PR do?

This addresses a bug introduced during the migration to the V1 engine: only the first finished request was removed from the batch, leading to the number of available slots in the static batch to be exhausted.
This pull-request also better compares the cached requests and the scheduled requests to raise an error if some of the latter are missing from the batch.


dacorvo:
Reverting back to draft as the copilot review revealed a bug in the remove_requests that might explain why all requests are not removed  😅

PR #1036, opened 2025-11-28
Device memory usage

dacorvo:
# What does this PR do?

This pull-request adds utilities to read and log neuron device memory usage.
This pull request also refactors and simplifies the sharding logic for attention heads and key/value heads across the codebase, improves logging for cache allocation and head adjustment.
Finally it cleans up unused or redundant code in the Makefile, adding an install target and using `uv` whenever possible.

### DIFF
--- a/optimum/neuron/utils/neuron_device_memory.py
+++ b/optimum/neuron/utils/neuron_device_memory.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # coding=utf-8
-# Copyright 2023 The HuggingFace Inc. team. All rights reserved.
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


Makefile
```diff
@@ -13,6 +13,8 @@
 #  limitations under the License.
 SHELL := /bin/bash
 CURRENT_DIR = $(shell pwd)
+UV = $(shell command -v uv)
+UV_ARGS = $(if $(UV),--index-strategy unsafe-best-match)
```
tengomucho, line 17: We already assume a given system is available (the AMI for example). If that is the case, I think we could install uv there and just use it, without adding additional checks

optimum/neuron/version.py
```diff
@@ -12,6 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.4.3.dev0"
+__version__ = "0.4.3.dev1"
```
tengomucho, line 15: the commit message here speaks about flash decoding... 🙄  While bumping the version might not hurt, is it necessary?

optimum/neuron/utils/neuron_device_memory.py
```diff
@@ -0,0 +1,280 @@
...
+
+    if not device_mem_path.exists():
+        return CoreDeviceMemory(categories={})
+
+    categories: Dict[str, MemoryCounter] = {}
+
+    for category in DEVICE_MEM_CATEGORIES:
+        try:
+            categories[category] = read_memory_counter(device_mem_path, category)
+        except Exception:
```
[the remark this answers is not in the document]
tengomucho: I agree with the comments, plus I would add a constant `EXPECTED_EXCEPTIONS = (FileNotFoundError, PermissionError, ValueError)` to avoid repeating over and over

tengomucho (commented):
Thanks a lot! Few comments, otherwise seems good!

Makefile
dacorvo: `uv` is not installed by default on the AWS DLAMI, which I personally often use when bumping versions.

optimum/neuron/version.py
dacorvo: Changed the commit message

tests/decoder/test_device_memory.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+from optimum.neuron.utils.neuron_device_memory import get_neuron_device_memory
```
tengomucho, line 3: not really a decoder test, but I understand it had to be placed somewhere

tests/decoder/test_device_memory.py
```diff
@@ -9,4 +9,4 @@ def test_neuron_device_memory():
     neuron_device_memory = get_neuron_device_memory()
     assert neuron_device_memory is not None
     assert len(neuron_device_memory.devices) == get_neuron_devices_count()
-    assert neuron_device_memory.get_total_memory() > 0
+    assert neuron_device_memory.get_total_memory() >= 0
```
tengomucho, line 12: when is it going to be 0?

dacorvo: On the CI, it looks like it is zero when no other tests have been run

PR #1040, opened 2025-12-10
Update pyproject.toml for uv and SDK 2.26.1

michaelbenayoun:
# What does this PR do?

- Updates the packages versions for SDK 2.26.1
- Make `uv` usable: we can run `uv sync` without failing now

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -160,13 +160,13 @@ name = "pytorch-cpu"
 url = "https://download.pytorch.org/whl/cpu"
 explicit = true
 
-# [tool.uv.sources]
-# torch = [
-#     { index = "pytorch-cpu" },
-# ]
-# torchvision = [
-#     { index = "pytorch-cpu" },
-# ]
+[tool.uv.sources]
+torch = [
+    { index = "pytorch-cpu" },
+]
+torchvision = [
+    { index = "pytorch-cpu" },
+]
     
 [tool.ruff]
 line-length = 119


pyproject.toml
```diff
@@ -86,7 +84,8 @@ neuron = [
     "neuron-cc[tensorflow]==1.22.0.0",
```
JingyaHuang, line 84: We shall remove the extra which is dedicated to inf1 
cc. @dacorvo @tengomucho

pyproject.toml
```diff
@@ -86,16 +84,19 @@ neuron = [
...
     "torchvision",
-    "numpy==1.22.3",
+    "numpy==1.22.2",
+    "protobuf<=3.20.1",
 ]
 neuronx = [
     "wheel",
-    "neuronx-cc==2.21.18209.0",
-    "torch-neuronx==2.8.0.2.10.13553",
+    "neuronx-cc==2.21.33363.0",
```
dacorvo, line 91: We already merged these changes last week 😉

michaelbenayoun: Awesome, I still had that on my branch, my bad!

michaelbenayoun:
> This seems nice! I think it would be better to add some documentation about how to use uv with optimum-neuron.

I'd say let's test it and use it for a while and change the whole docs once we validated the whole thing. But yes I agree!

dacorvo (commented):
I just tried the branch and `uv sync` reinstalls the torch CUDA version although I have the torch CPU version installed, while `uv pip install` behaves as expected and keeps the torch version.

pyproject.toml
```diff
@@ -133,6 +133,32 @@ include = ["optimum*"]
...
+
+[[tool.uv.index]]
+name = "neuron"
+url = "https://pip.repos.neuron.amazonaws.com/"
+default = true
+    
+[[tool.uv.index]]
+name = "pypi"
+url = "https://pypi.org/simple"
+
```
dacorvo, line 162:
```suggestion
[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[tool.uv.sources]
torch = [
    { index = "pytorch-cpu" },
]
torchvision = [
    { index = "pytorch-cpu" },
]
```

dacorvo: This will make sure we install the CPU version of pytorch

dacorvo (changes requested):
Just one minor change 🙏

pyproject.toml
```diff
@@ -133,6 +133,32 @@ include = ["optimum*"]
...
+  ],
+]
+
+[[tool.uv.index]]
+name = "neuron"
+url = "https://pip.repos.neuron.amazonaws.com/"
+default = true
+    
+[[tool.uv.index]]
+name = "pypi"
```
tengomucho: I do not think you need the `pypi` index, being the default one.

pyproject.toml
michaelbenayoun: Nice!
Adding it.

pyproject.toml
michaelbenayoun: Remove it then!

PR #1043, opened 2025-12-11
Gradient checkpointing fix

michaelbenayoun:
# What does this PR do?

XLA-friendly gradient checkpointing function that accept keyword arguments.

### DIFF
--- a/optimum/neuron/models/training/llama/modeling_llama.py
+++ b/optimum/neuron/models/training/llama/modeling_llama.py
@@ -45,6 +45,7 @@ from ..config import TrainingNeuronConfig
 from ..loss_utils import ForCausalLMLoss
 from ..masking_utils import create_causal_mask
 from ..modeling_utils import NeuronModelMixin
+from ..training_utils import checkpoint_with_kwargs
 from ..transformations_utils import (
     CustomModule,
     FusedLinearsSpec,
@@ -667,8 +668,8 @@ class LlamaModel(NeuronModelMixin, LlamaPreTrainedModel):
         # Decoder layers
         for decoder_layer in self.layers[: self.config.num_hidden_layers]:
             if self.gradient_checkpointing and self.training:
-                hidden_states = checkpoint(
-                    decoder_layer.__call__,
+                hidden_states = checkpoint_with_kwargs(
+                    decoder_layer,
                     hidden_states,
                     causal_mask,
                     position_ids,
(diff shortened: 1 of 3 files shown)


optimum/neuron/models/training/training_utils.py
```diff
@@ -199,3 +200,13 @@ def is_custom_modeling_model(model) -> bool:
...
+
+
+def checkpoint_with_kwargs(fn, *args, **kwargs):
+    """XLA-compatible gradient checkpointing that accepts keyword arguments via functools.partial."""
+    from functools import partial
+
+    from torch_xla.utils.checkpoint import checkpoint
+
+    fn_with_kwargs = partial(fn, **kwargs)
+    return checkpoint(fn_with_kwargs, *args)
```
tengomucho, line 212: I think this should be the other way around:
```suggestion
    return checkpoint(*args, fn_with_kwargs)
```

michaelbenayoun: No, the first argument must be the function we checkpoint, check [here](https://github.com/pytorch/xla/blob/master/torch_xla/utils/checkpoint.py#L219).

michaelbenayoun:
> why do you need this function? Can you provide a test that shows its usage and tests that it works?

It is just a wrapper around the original checkpoint [function](https://github.com/pytorch/xla/blob/master/torch_xla/utils/checkpoint.py#L219). The goal is to make it possible to pass keyword arguments to the function we are wrapping with `checkpoint`.

I can add a test, but you mentioned many times that the CI was too long. For such a small change I was not considering it, please confirm you want a test.

optimum/neuron/models/training/llama/modeling_llama.py
```diff
@@ -667,8 +668,8 @@ def forward(
         # Decoder layers
         for decoder_layer in self.layers[: self.config.num_hidden_layers]:
             if self.gradient_checkpointing and self.training:
-                hidden_states = checkpoint(
-                    decoder_layer.__call__,
+                hidden_states = checkpoint_with_kwargs(
```
tengomucho, line 671: you change this here, only for granite, but not for the other models. I think you should change it everywhere, or nowhere, or explain why you only changed it here.

michaelbenayoun: I dont undertsand? We change it "everywhere".

tengomucho: yep sorry, forgot that qwen3 inherited the forward from llama

PR #1044, opened 2025-12-11
PEFT update for GRPO

michaelbenayoun:
# What does this PR do?

This PR adds methods to merge / unmerge LoRA adapters. This is required because for GRPO, we will need to merge the adapters to the base model before transferring it to vLLM.


michaelbenayoun:
> Seems nice. I think you should publish a complete example at some point, otherwise it is hard to see how to use all these

Yes will do, we will soon have a complete GRPO example, where it is used.

PR #1045, opened 2025-12-16
Add Qwen embedding guide and notebook

pinak-p:
# What does this PR do?

Adds a notebook and guide for Qwen Embedding support on Trainium and Inferentia using Optimum Neuron.

Fixes # (issue)

## Before submitting


notebooks/feature-extraction/qwen_embedding.ipynb
```diff
@@ -0,0 +1,206 @@
...
+    "# If you are using a AWS Inferentia2 instance and use 'tensor_parallel_size=4', you should set the following environment variable as well.\n",
+    "#import os\n",
+    "#os.environ[\"LOCAL_WORLD_SIZE\"] = '4'\n",
+    "\n",
+    "neuron_config = Qwen3NeuronModelForEmbedding.get_neuron_config(\n",
+    "        model_id, batch_size=2, sequence_length=1024, tensor_parallel_size=4\n",
+    "    )\n",
+    "\n",
+    "neuron_model = Qwen3NeuronModelForEmbedding.export(\n",
+    "        model_id=model_id, neuron_config=neuron_config, load_weights=True\n",
```
dacorvo: You must NOT load the weights if you want to save the model to disk, as the torchscript file will contain the weights and fail to load afterwards. AFAIK, this is a limitation of neuronx-distributed: you cannot save the compiled model after having loaded the weights.

docs/source/inference_tutorials/qwen_embedding.mdx
```diff
@@ -0,0 +1,136 @@
+<!---
```
dacorvo: We try to avoid duplicating guides and notebooks: just add the link to your notebook in notebooks.mdx instead.

notebooks/feature-extraction/qwen_embedding.ipynb
pinak-p: I have made the change to the notebook. But I observed that saving the model to disk and subsequently loading works in both scenarios (`load_weights=True` and `load_weights=False`). 

Can you check once if the behavior has changed ?

PR #1046, opened 2025-12-17
vLLM searches local configuration files

tengomucho:
# What does this PR do?

Changed the way vLLM starts, so that it will try to use the local configuration instead of downloading it from the hub.
This allows to launch vLLM on a model whose weights have already been downloaded, so the `HF_TOKEN` is not necessary anymore after the weights have been downloaded.

Also:
- dev version bumped,
- removed a warning related to `torch_dtype` and
- added CLI option aliases for `serve`.


tengomucho:
After discussion, I will re-think and refactor this to simplify and better isolate single and multi model handling, so going back to draft here.

PR #1047, opened 2025-12-19
Fix speculation algorithm

dacorvo:
# What does this PR do?

This pull request refactors and improves the decoder backend to simplify modeling and improve the robustness of the generation mixin.

The main focus for refactoring is on simplifying input management during generation, improving attention mask and position ID handling, and removing unused or redundant code.

The speculative generation algorithm ic completely rewritten in the process.

### DIFF
--- a/optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
+++ b/optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
@@ -273,21 +273,14 @@ class NxDGenerationMixin(GenerationMixin, ABC):
             logger.warning(f"The following kwargs are not supported for neuron model: {list(explicit_kwargs.keys())}")
 
         pad_token_id = generation_config.pad_token_id
-        eos_token_id = generation_config.eos_token_id
 
         if assistant_model.neuron_config.on_device_sampling:
             raise ValueError("Assistant model must not use on-device sampling")
-
-        # Init values
-        if eos_token_id is not None and pad_token_id is None:
-            raise ValueError("If `eos_token_id` is defined, make sure that `pad_token_id` is defined.")
-        if isinstance(eos_token_id, int):
-            eos_token_id = [eos_token_id]
-        eos_token_id_tensor = torch.tensor(eos_token_id).to(input_ids.device) if eos_token_id is not None else None
+        if self.neuron_config.batch_size > 1:
+            raise ValueError("Assisted decoding is only supported for batch size 1")
 
         # Other auxiliary variables
         max_len = stopping_criteria[0].max_length
-        cur_len = input_ids.shape[-1]
         spec_len = self.neuron_config.speculation_length
 
         # Prepare input tensors
@@ -299,8 +292,7 @@ class NxDGenerationMixin(GenerationMixin, ABC):
         assert do_sample is False, "Assisted decoding is only supported for greedy decoding."
         sampling_params = prepare_sampling_params(batch_size=batch_size)
 
-        # Run the target model once and get the first generated token
-
+        # Prefill the target model KV cache and get the first accepted token
         outputs = self.forward(
             input_ids=input_ids,
             attention_mask=attention_mask,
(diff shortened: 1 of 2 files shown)


optimum/neuron/utils/neuron_device_memory.py
```diff
@@ -65,12 +65,40 @@ def read_sysfs_value(file_path: Path) -> str:
...
+    for factor, suffix in [
+        (2**40, "TB"),
+        (2**30, "GB"),
+        (2**20, "MB"),
+        (2**10, "KB"),
+    ]:
+        if memory_bytes >= factor:
+            return f"{memory_bytes / factor:.2f} {suffix}"
+    return f"{memory_bytes} bytes"
+
```
tengomucho, line 87: why did you move this? 🤷

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
```diff
@@ -409,7 +409,7 @@ def _assisted_decoding(
             # 8. If the rest length is smaller than speculation length, we directly run the target model to finish
             if max_len - cur_len < spec_len:
                 # @yihsian: TODO: complete with using target tokengen model
-                break
+                pass
```
tengomucho: nit: I would put "continue" in here, in case someone add code below.

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
```diff
@@ -309,9 +309,9 @@ def _assisted_decoding(
         # Speculation loop
         while True:
             # 1 Token generation using draft model
-            is_for_token_generation = assistant_model.kv_cache_populated
             for _ in range(spec_len):
                 # 1.1 Prepare assistant model inputs
+                is_for_token_generation = assistant_model.kv_cache_populated
```
tengomucho: why do you need to move this?

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
```diff
@@ -12,23 +12,44 @@
...
 
+logger = logging.getLogger("Neuron")
+
+
+def increase_attention_mask(attention_mask: torch.LongTensor, num_new_tokens: int) -> torch.LongTensor:
+    # Prepend num_new_tokens to the attention mask (inputs are right padded)
+    return torch.cat(
+        [attention_mask.new_ones((attention_mask.shape[0], num_new_tokens)), attention_mask],
+        dim=-1,
+    )
```
tengomucho, line 38: can you confirm the resulting tensor has always the same shape? Otherwise this will create a new graph every time, isn't it?
I might be wrong, but I would have thought this had to be something more like this:
```suggestion
    return torch.cat(
        [attention_mask.new_ones((attention_mask.shape[0], num_new_tokens)), attention_mask[...,:-1],
        dim=-1,
    )
```

tests/decoder/test_decoder_generation.py
```diff
@@ -246,14 +247,70 @@ def increase_attention_mask(attention_mask):
...
+    [[17, (5, 5, 5, 1)], [128, (5, 5, 5, 5, 5, 1, 5, 0)]],
+    ids=["partial_generation", "full_generation"],
+)
+def test_generation_assisted_decoding(caplog, speculation, max_new_tokens, speculated_tokens):
     model_path, draft_model_path = speculation
     model = NeuronModelForCausalLM.from_pretrained(model_path)
     tokenizer = AutoTokenizer.from_pretrained(model_path)
     assistant_model = NeuronModelForCausalLM.from_pretrained(draft_model_path)
-    prompt = "What is Deep Learning?"
+    prompt = "One of my fondlest memories is"
```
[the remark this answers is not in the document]
tengomucho: It should be fondest I guess

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
```diff
@@ -273,21 +273,14 @@ def _assisted_decoding(
...
             raise ValueError("Assistant model must not use on-device sampling")
-
-        # Init values
-        if eos_token_id is not None and pad_token_id is None:
-            raise ValueError("If `eos_token_id` is defined, make sure that `pad_token_id` is defined.")
-        if isinstance(eos_token_id, int):
-            eos_token_id = [eos_token_id]
-        eos_token_id_tensor = torch.tensor(eos_token_id).to(input_ids.device) if eos_token_id is not None else None
+        if self.neuron_config.batch_size > 1:
+            raise ValueError("Assisted decoding is only supported for batch size 1")
```
tengomucho, line 280: this makes speculative decoding not very useful for LLM inference, right?

tests/decoder/test_decoder_generation.py
```diff
@@ -246,18 +247,70 @@ def increase_attention_mask(attention_mask):
 
 @is_inferentia_test
 @requires_neuronx
-def test_generation_assisted_decoding(speculation):
+@pytest.mark.parametrize(
+    "max_new_tokens, speculated_tokens",
+    [[17, (5, 5, 5, 1)], [128, (5, 5, 5, 5, 5, 1, 5, 0)]],
```
tengomucho: I think that having lists of `speculated_tokens` tokens might end up hard to maintain. I think at least a comment explaining how to update the list would be good, otherwise you will be the only one capable of updating this if, for example, and SDK produces a different graph producing a different token sequence.

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
dacorvo: This runs on CPU, so there is no compilation involved here. Whn forward is called on the model, the inputs tensors are passed to the model wrappers, properly padded to match the input dimensions specified at compilation time, then passed to the compiled models.

optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
dacorvo: Because in that version of the algorithm, the prefill was also included in the loop, and the flag was never updated for the subsequent decodes.

tengomucho (approved):
LGTM thanks

PR #1055, opened 2026-01-07
Improve vLLM performance when on-device-sampling is disabled

dacorvo:
# What does this PR do?

This pull-request modifies the vLLM runner to use the faster sampling algorithm imported from NxDi when on-device-sampling is disabled.

This also modifies the invocation of the vLLM sampler to correctly apply repetition/frequence penalties.

Finally, this pull request also refactors and simplifies the decoder and sampling logic in the inference backend. The main changes include inlining decoder logic, streamlining the sampling API, and improving top-k parameter handling for sampling.

### DIFF
--- a/optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
+++ b/optimum/neuron/models/inference/backend/modules/generation/generation_utils.py
@@ -17,10 +17,16 @@ from abc import ABC, abstractmethod
 
 import torch
 from transformers import GenerationConfig
-from transformers.generation import GenerationMixin, SampleDecoderOnlyOutput
-from transformers.generation.logits_process import LogitsProcessorList
+from transformers.generation.logits_process import (
+    LogitsProcessorList,
+    TemperatureLogitsWarper,
+    TopKLogitsWarper,
+    TopPLogitsWarper,
+)
 from transformers.generation.stopping_criteria import StoppingCriteriaList
+from transformers.generation.utils import GenerationMixin, SampleDecoderOnlyOutput
 
+from ......generation.logits_process import FusedLogitsWarper
 from .sampling import (
     prepare_sampling_params,
 )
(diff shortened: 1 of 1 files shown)


optimum/neuron/models/inference/backend/modules/generation/sampling.py
```diff
@@ -291,26 +288,24 @@ def _multinomial_sample(self, token_logits, sampling_params, return_values, dim,
         # Apply top_k first
         top_k_logits_values, top_k_logits_indices = self._top_k_masked(token_logits, top_k, dim, rank_id)
 
-        # Apply temperature
-        top_k_logits_values = torch.divide(top_k_logits_values, temperature)
+        if torch.any(temperature != 1.0):
```
tengomucho, line 291: why do you need this if? If I get it right this is executed when compiling, so having a check on the value of the temperature would probably generate two graphs, right?

optimum/neuron/models/inference/backend/modules/generation/sampling.py
```diff
@@ -291,26 +288,24 @@ def _multinomial_sample(self, token_logits, sampling_params, return_values, dim,
...
 
-        # Apply top_p
-        probs_soft_max = self._soft_max(top_k_logits_values, dim)
-        probs_cumsum = cumsum(tensor_in=probs_soft_max, dim=dim, on_cpu=self.on_cpu)
-        top_p = torch.max(torch.min(probs_cumsum), top_p)
-        top_p_mask = torch.greater(probs_cumsum, top_p).index_fill_(
-            dim, torch.tensor([0], device=top_p.device), False
-        )  # need to keep at least one token
-        top_k_logits_values = top_k_logits_values.masked_fill_(top_p_mask, self.IGNORED_LOGITS_VALUE)
+        if torch.any(top_p < 1.0):
```
tengomucho, line 294: same here

tengomucho (approved):
LGTM

PR #1060, opened 2026-01-19
Update LLM deployment documentation

dacorvo:
# What does this PR do?

This pull-request starts by cleaning up the documentation, removing tutorials duplicating notebooks, and simplifying the toctree.

It then updates the deployment tutorials to use vLLM instead of TGI.

In the process, the SageMaker tutorials are also modified to adapt to python SageMaker SDK v3.

### DIFF
--- a/optimum/neuron/utils/ecr.py
+++ b/optimum/neuron/utils/ecr.py
@@ -112,7 +112,7 @@ def image_uri(
             + str(e)
         )
         raise ValueError(message)
-    # Remove images without an image Tag (these are untagged images)
+    # Filter images to retain only images with tags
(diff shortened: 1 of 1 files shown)


optimum/neuron/utils/ecr.py
```diff
@@ -102,6 +115,8 @@ def image_uri(
             + str(e)
         )
         raise ValueError(message)
+    # Remove images without an image Tag (these are untagged images)
```
tengomucho:
```suggestion
    # Filter images to retain only images with tags
```

tengomucho (approved):
Just a nit in a comment wording, otherwise LGTM, thanks!

PR #1066, opened 2026-01-23
Optimize lookup

dacorvo:
# What does this PR do?

This modifies further the cache registry lookup algorithm to use `list_repo_tree` instead of `list_repo_files`. This is faster because `list_repo_tree` accepts a `path_in_repo` parameter that allows us to narrow the search area to include just the target entries directory.

### DIFF
--- a/optimum/neuron/cache/hub_cache.py
+++ b/optimum/neuron/cache/hub_cache.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import functools
 import logging
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


optimum/neuron/cache/hub_cache.py
```diff
@@ -346,6 +346,17 @@ def synchronize_hub_cache(
     hub_cache_proxy.synchronize(non_blocking=non_blocking)
 
 
+@functools.cache
+def hub_registry_path(cache_repo_id):
```
tengomucho, line 351: you could add a docstring to describe what the function does (saying it contains the optimum neuron version)

optimum/neuron/utils/version_utils.py
```diff
@@ -14,65 +14,72 @@
...
 
-_neuronxcc_version: str | None = None
-_torch_xla_version: str | None = None
-_neuronx_distributed_version: str | None = None
-_torch_version: str | None = None
-
-
+def get_pinned_version(package_name: str) -> str:
+    """
+    Get the pinned version of a package from the `optimum-neuron` package metadata.
```
tengomucho, line 28:
```suggestion
    Get the pinned version of a package in the dependency list of the `optimum-neuron` package.
```

PR #1067, opened 2026-01-23
Fix cache registry for embedding models

dacorvo:
# What does this PR do?

Before that pull-request, the embeddings models were registered in the cache registry with the `text-generation` task instead of `feature-extraction`.

This pull-request also introduces an abstract class for neuron embedding models.

### DIFF
--- a/tests/decoder/test_cache.py
+++ b/tests/decoder/test_cache.py
@@ -24,7 +24,7 @@ import pytest
 import torch
 from huggingface_hub import HfApi
 
-from optimum.neuron import NeuronModelForCausalLM
+from optimum.neuron import NeuronModelForCausalLM, NeuronModelForEmbedding
 from optimum.neuron.cache import get_hub_cached_entries, get_hub_cached_models, synchronize_hub_cache
 from optimum.neuron.utils.testing_utils import is_inferentia_test, requires_neuronx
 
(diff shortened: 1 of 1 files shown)


optimum/neuron/models/inference/backend/pretrained_model.py
```diff
@@ -359,7 +359,7 @@ def _export(
         cache_entry = (
             None
             if os.path.exists(model_id)
-            else SingleModelCacheEntry(model_id, task="text-generation", config=config, neuron_config=neuron_config)
+            else SingleModelCacheEntry(model_id, task=cls.task, config=config, neuron_config=neuron_config)
```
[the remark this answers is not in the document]
dacorvo: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/huggingface/optimum-neuron/pull/1067#discussion_r2721483191)

optimum/neuron/utils/input_generators.py
```diff
@@ -54,10 +54,10 @@ class DTYPE_MAPPER:
     REVERSE_MAPPING = {v: k for k, v in MAPPING.items()}
 
     @classmethod
-    def str(cls, dtype):
+    def str(cls, dtype) -> str:
         if not isinstance(dtype, str):
             if dtype in cls.REVERSE_MAPPING:
-                return cls.REVERSE_MAPPING.get(dtype)
+                return cls.REVERSE_MAPPING[dtype]
```
tengomucho, line 60: this commit is unnecessary, but why not

dacorvo: It is not actually. Without that change, pylance thinks the returned value can be None, and it raises further errors down the line whenever this method is called.

PR #1071, opened 2026-01-29
feat(docker): vllm container uses uv to install package

tengomucho:
# What does this PR do?

The vLLM Dockerfile is revisited to be simplified in its installation steps, and to use uv to install the optimum neuron package and dependencies.

### DIFF
--- a/optimum/neuron/version.py
+++ b/optimum/neuron/version.py
@@ -12,6 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.4.5.dev1"
+__version__ = "0.4.5.dev2"
 
 __sdk_version__ = "2.26.1"


docker/vllm/Dockerfile
```diff
@@ -34,23 +33,12 @@ RUN apt-get update -y \
...
-    --index-url https://download.pytorch.org/whl/cpu
-
 # Install optimum-neuron
 RUN mkdir optimum-neuron
 COPY optimum optimum-neuron/optimum
 COPY pyproject.toml optimum-neuron/pyproject.toml
 RUN ls optimum-neuron
-RUN cd optimum-neuron && pip3 install .[neuronx,vllm] \
-    --extra-index-url=https://pip.repos.neuron.amazonaws.com
+RUN cd optimum-neuron && uv pip install --system .[neuronx,vllm]
```
dacorvo, line 41: Did you try using uv sync instead ? It may result in a smaller image.

dacorvo (approved):

docker/vllm/Dockerfile
tengomucho: `uv sync` is more to create a venv, but we just want to install the package here, so no need.

PR #1073, opened 2026-02-03
ci(dependencies): neuronx packages are cached across jobs

tengomucho:
This is to avoid downloading the files several times, so to speed up each job. Now one job will always download the files in a cached location and reuse that.

Note that for setups using the action when the mount point is not shared, the only difference is that it will not benefit from the shared download, but it will work anyway.


tengomucho:
I am going to try another path to improve performance.

PR #1102, opened 2026-04-01
feat(cache): replace Git-based hub cache with HF Storage Buckets

dacorvo:
## Summary

Replace the Git-based `CompileCacheHfProxy` with HF Storage Bucket-based NEFF cache.

- Bucket operations run in isolated subprocess via `uv run --with "huggingface_hub>=1.0"` to avoid transformers version conflict
- Context manager (`hub_neuronx_cache`) handles fetch on enter, sync on exit
- NEFFs stored in both flat (hf-mount compatible) and per-model areas (Xet dedup)
- Export records are flat neuron config dicts (advisory, for lookup only)
- No monkey-patching of `libneuronxla`

## Test plan

- [x] Unit tests: `pytest tests/cache/` (40 tests, CPU)
- [x] Integration: real Neuron compilation against `aws-neuron/optimum-neuron-neff-cache`
- [x] Phase 1: compile + sync to bucket
- [x] Phase 2: clean cache + fetch from bucket (zero recompilation)
- [x] Phase 3-4: warm cache (no fetch, no compilation)
- [ ] vLLM model loader `select_hub_cached_entries` key name normalization (known TODO)

Ref: #1101

🤖 Generated with [Claude Code](https://claude.com/claude-code)


optimum/neuron/cache/bucket_utils.py
```diff
@@ -0,0 +1,143 @@
+# coding=utf-8
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
```
dacorvo: Use 2026 copyright

optimum/neuron/cache/bucket_utils.py
```diff
@@ -0,0 +1,143 @@
...
+# limitations under the License.
+"""Utilities for HF Storage Bucket-based NEFF cache."""
+
+import hashlib
+import json
+import os
+from pathlib import Path
+
+from ...utils import logging
+from ..utils.cache_utils import HF_HOME
```
dacorvo: I think this comes directly from transformers or huggingfqce_hub, so you can import it instead to limit dependencies towards utils

optimum/neuron/cache/bucket_utils.py
```diff
@@ -0,0 +1,143 @@
...
+from ..utils.misc import is_main_worker
+
+
+logger = logging.get_logger()
+
+CACHE_BUCKET_FILENAME = "optimum_neuron_cache_bucket"
+HF_HOME_CACHE_BUCKET_FILE = f"{HF_HOME}/{CACHE_BUCKET_FILENAME}"
+
+DEFAULT_CACHE_BUCKET = "aws-neuron/optimum-neuron-neff-cache"
+if os.environ.get("HUGGINGFACE_CO_STAGING") == "1":
```
dacorvo: I don't think this is relevant anymore

optimum/neuron/cache/bucket_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    if is_main_worker() and existing is not None:
+        logger.warning(
+            f"A custom cache bucket was already registered: {existing}. It will be overwritten to {bucket_id}."
+        )
+
+    with open(hf_home_cache_bucket_file, "w") as fp:
+        fp.write(bucket_id)
+
+
+def delete_cache_bucket_from_hf_home(
```
dacorvo: Is it called by something ?

tests/cache/test_bucket_utils.py
```diff
@@ -0,0 +1,179 @@
...
+    def test_roundtrip(self):
+        original = "org-name/model-name"
+        assert decode_model_id(encode_model_id(original)) == original
+
+    def test_encode_nested_org(self):
+        # Edge case: model_id with multiple slashes (not standard but handle gracefully)
+        assert encode_model_id("a/b/c") == "a--b--c"
+
+
+class TestConfigHash:
```
dacorvo: Don't use test classes, I prefer individual methods.

optimum/neuron/cache/bucket_cli.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Ensure directories exist
+    for module_name in new_module_names:
+        (local_compiler_dir / module_name).mkdir(parents=True, exist_ok=True)
+
+    api.download_bucket_files(bucket_id, files=files_to_download)
+    return {"downloaded": len(files_to_download), "modules": new_module_names}
+
+
+def cmd_sync(args: dict) -> dict:
```
dacorvo: use explicit named args: this is a bad design to use anonymous args

optimum/neuron/cache/bucket_cli.py
```diff
@@ -0,0 +1,212 @@
...
+
+COMMANDS = {
+    "list_modules": cmd_list_modules,
+    "fetch": cmd_fetch,
+    "sync": cmd_sync,
+    "lookup": cmd_lookup,
+}
+
+
+def main():
```
dacorvo: So it seems you wanted to avoid writing a global args parser and parsed the args in the methods instead. 
I understand the intent to avoid duplicate named args, but this is harder to read and maintain. If you want to save defining the parser, I think there are some existing tools for that like click or even simpler typer.

tests/cache/test_bucket_cache.py
```diff
@@ -0,0 +1,386 @@
...
+    hub_neuronx_cache,
+    lookup_cache,
+    sync_cache,
+    sync_tagged_cache,
+    tag_modules,
+)
+from optimum.neuron.cache.bucket_utils import bucket_model_prefix
+
+
+TEST_BUCKET = "dacorvo/neuron-compile-cache"
```
dacorvo: let's use optimum-internal-testing/optimum-neuron-neff-cache instead. How do you prevent collisions when multiple hosts are running these tests ?

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,92 @@
...
     Args:
-        entry (`ModelCacheEntry | None`, defaults to `None`):
-            An optional dataclass containing metadata associated with the model corresponding
-            to the cache session. Will create a dedicated entry in the cache registry.
-        cache_repo_id (`str | None`, defaults to `None`):
-            The id of the cache repo to use to fetch the precompiled files.
-        cache_dir (`str | Path | None`, defaults to `None`):
-            The directory that is used as local cache directory.
+        cache_path: The path of the local cache folder. Defaults to neuron cache path.
+        cache_repo_id: Ignored (kept for backwards compatibility). Bucket is resolved automatically.
```
dacorvo: this was for testing: how do you handle that now ?

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,92 @@
...
+        non_blocking: Ignored (kept for backwards compatibility).
```
dacorvo: why ?

optimum/neuron/utils/cache_utils.py
```diff
@@ -10,7 +10,14 @@
...
 # limitations under the License.
-"""Utilities for caching."""
+"""Utilities for caching.
+
+The repo-based functions (get_hf_hub_cache_repo, create_custom_cache_repo, etc.)
+are deprecated in favor of the bucket-based cache in optimum.neuron.cache.bucket_utils.
+They are kept temporarily for backwards compatibility.
+
+New code should import bucket utilities directly from optimum.neuron.cache.bucket_utils.
+"""
```
dacorvo, line 16: why is that in a separate commit ?

optimum/neuron/cache/hub_cache.py
```diff
@@ -36,6 +40,47 @@
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def hub_neuronx_cache(
+    model_id: str | None = None,
+    export_config: dict | None = None,
+    # Legacy parameters (backwards compat)
+    entry: ModelCacheEntry | None = None,
```
dacorvo: Ok to remove cache entry, but the model cache entry task is still relevant: the keys are the model_id and task, which lead to DIFFERENT graphs, then different NEFFs. We should include task both as a parameter and as a subdir in the cache tree.

optimum/neuron/cache/hub_cache.py
```diff
@@ -36,6 +40,47 @@
...
+    cache_repo_id: str | None = None,
+    cache_dir: str | Path | None = None,
```
dacorvo: what backward compat are you talking about ? this is a purely internal API

optimum/neuron/cache/hub_cache.py
```diff
@@ -36,6 +40,47 @@
...
+    # New parameters
+    sync: str = "immediate",
+    compiler_version: str | None = None,
+    bucket_id: str | None = None,
+):
+    """Context manager for bucket-based NEFF cache.
+
+    Supports both old-style calls (entry=ModelCacheEntry) and new-style (model_id, export_config).
```
dacorvo: No

optimum/neuron/cache/traced.py
```diff
@@ -13,20 +13,30 @@
...
+from ..utils.cache_utils import get_neuron_cache_path
+from ..utils.version_utils import get_neuronxcc_version
+from .entries.cache_entry import ModelCacheEntry
+from .hub_cache import hub_neuronx_cache
 
 
 logger = logging.getLogger(__name__)
 
 
 def cache_traced_neuron_artifacts(neuron_dir: Path, cache_entry: ModelCacheEntry):
```
dacorvo: We should not do that anymore, and use the cintext instead, since now it does not rely on the monkey patching anymore

optimum/neuron/modeling_traced.py
```diff
@@ -336,8 +336,13 @@ def _export(
             )
             cache_entry = SingleModelCacheEntry(model_id=model_id, task=task, config=compilation_config)
             compile_cache = create_hub_compile_cache_proxy()
-            model_cache_dir = compile_cache.default_cache.get_cache_dir_with_cache_key(f"MODULE_{cache_entry.hash}")
-            cache_available = compile_cache.download_folder(model_cache_dir, model_cache_dir)
+            if compile_cache is not None:
```
dacorvo: this has to be completely rewritten, since the logic is different now: we populate the cache instead of fetching the files

optimum/exporters/neuron/convert.py
```diff
@@ -415,21 +429,28 @@ def export_models(
...
 
-    # cache neuronx model
-    if not disable_neuron_cache and is_neuronx_available():
-        model_id = get_model_name_or_path(model_config) if model_name_or_path is None else model_name_or_path
-        if len(compile_configs) == 1:
-            # FIXME: this is overly complicated just to pass the config
-            cache_config = list(compile_configs.values())[0]
-            cache_entry = SingleModelCacheEntry(model_id=model_id, task=task, config=cache_config)
-        else:
+    # Sync new NEFFs to bucket
```
dacorvo: use the same context wrapper

optimum/neuron/cache/bucket_utils.py
```diff
@@ -0,0 +1,135 @@
...
+def bucket_flat_prefix(compiler_version: str) -> str:
+    """Bucket prefix for the flat NEFF area (hf-mount compatible).
+
+    Maps to: neuronxcc-{compiler_version}/
+    This matches the local cache layout exactly.
+    """
+    return f"neuronxcc-{compiler_version}"
+
+
+def bucket_model_prefix(compiler_version: str, model_id: str) -> str:
```
dacorvo, line 95: I think it might be easier to read if we had an org/domain subdir, potentially to do things like: rebuild cache for a given org

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,424 @@
...
+logger = logging.get_logger()
+
+# Path to the standalone bucket CLI script
+_BUCKET_CLI_PATH = str(Path(__file__).parent / "bucket_cli.py")
+
+# Dependencies for the isolated uv run environment
+_UV_DEPS = ["huggingface_hub>=1.0", "typer"]
+
+
+def _run_bucket_cli(command: str, cli_args: list[str]) -> dict:
```
dacorvo: we need to add uv as a dependency, and also typer

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,424 @@
...
+    Raises RuntimeError if the subprocess fails.
+    """
+    uv_path = shutil.which("uv")
+    if uv_path is None:
+        raise RuntimeError(
+            "uv is required for bucket cache operations. "
+            "Install it: https://docs.astral.sh/uv/getting-started/installation/"
+        )
+
+    cmd = [uv_path, "run"]
```
dacorvo: if uv is installed as a python package, then maybe there is an api to get its path ?

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,424 @@
...
+    for dep in _UV_DEPS:
+        cmd.extend(["--with", dep])
+    cmd.extend(["python", _BUCKET_CLI_PATH, command, *cli_args])
+
+    result = subprocess.run(
```
dacorvo: I would rather spawn a server process and query it interactively than these sequential CLI commands

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,424 @@
...
+def tag_modules(
+    new_modules: set[str],
+    model_id: str,
+    task: str | None = None,
+    export_config: dict | None = None,
+    compiler_version: str | None = None,
+    bucket_id: str | None = None,
+    cache_dir: str | Path | None = None,
+):
+    """Write .bucket_meta.json in each new MODULE dir for deferred sync."""
```
dacorvo: It's all about the deferred sync, but this was actually called immediately after each export in practice, so I don't think we should keep this, as it adds a lot of complexity for little benefits. I am worried we would not properly recover from failures also, with pending irrlevant meta

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,424 @@
...
+
+    model_task_prefix = _model_task_prefix(compiler_version, model_id, task)
+    exports_prefix = f"{model_task_prefix}/exports/{on_version}"
+
+    try:
+        result = _run_bucket_cli("lookup", [bucket_id, exports_prefix])
+    except RuntimeError:
+        return []
+
+    return result.get("configs", [])
```
dacorvo: configs seems disconnected here

optimum/neuron/cache/bucket_cli.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python
```
dacorvo: I'd like the bucket cli to expose only low-level apis mirroring the actual huggungface_hub api. It is actually a proxy, and should not embed too much intelligence, so that we can swiftly switch to direct calls when we can update to the real huggingface_hub. All neuron cache specific code should be located in bucket_cache.py

optimum/neuron/cache/bucket_utils.py
```diff
@@ -120,16 +120,3 @@ def local_to_flat_bucket_path(module_dir_name: str, compiler_version: str) -> st
     Returns: e.g. "neuronxcc-2.28.4405.0+abc123/MODULE_X+Y"
     """
     return f"{bucket_flat_prefix(compiler_version)}/{module_dir_name}"
-
```
dacorvo: This looks like a fixup of the previous commit

optimum/neuron/cache/hub_cache.py
```diff
@@ -1,5 +1,5 @@
 # coding=utf-8
-# Copyright 2024 The HuggingFace Team. All rights reserved.
+# Copyright 2026 The HuggingFace Team. All rights reserved.
```
dacorvo: no, this file was pre-existing

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,78 @@
...
-    token = get_token()
-    return CompileCacheHfProxy(cache_repo_id, default_cache, endpoint=endpoint, token=token)
-
-
-REGISTRY_FOLDER = f"0_REGISTRY/{__version__}"
-
-
-@requires_torch_neuronx
 @contextmanager
 def hub_neuronx_cache(
```
dacorvo: I don't want any doppelganger: use a single context

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,78 @@
...
-    entry: ModelCacheEntry | None = None,
+    model_id: str | None = None,
+    task: str | None = None,
+    export_config: dict | None = None,
+    # Legacy parameters
+    entry=None,
```
dacorvo: we don't need to support legacy parameters since we modify the callers

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,78 @@
...
-    compiler_prefix = "neuronxcc-" + neuronxcc_version
-    api = HfApi()
-    root = api.list_repo_tree(cache_repo_id, path_in_repo="", recursive=False)
-    for root_file in root:
-        # Note that the actual path uses the extended compiler version (with commit hash),
-        # that is not available in the package metadata. We thus only check the prefix here.
-        if root_file.path.startswith(compiler_prefix):
-            return root_file.path + "/" + REGISTRY_FOLDER
-    raise ValueError(f"No hub cache registry found in {cache_repo_id} for neuronx-cc-{neuronxcc_version}")
+def synchronize_hub_cache(cache_path: str | None = None, cache_repo_id: str | None = None, non_blocking: bool = False):
```
dacorvo: we don't need this anymore, except maybe to push exports in the hf mount config ?

optimum/neuron/cache/hub_cache.py
```diff
@@ -12,482 +12,78 @@
...
-                sub_paths = registry_file.path.split("/")
-                if len(sub_paths) == len(root_sub_paths) + 4:
-                    # Look at the last four splits, i.e. model_arch/model_org/model_name/SHA.json
-                    model_arch, model_org, model_name = sub_paths[-4:-1]
-                    cached_models.add((model_arch, model_org, model_name))
-    except EntryNotFoundError:
-        pass
-    return cached_models
+    """Get cached neuron configs for a model+task from the bucket."""
+    return lookup_cache(model_id=model_id, task=task)
```
dacorvo: what about the cache_repo_id ? when was it specified ? If the signature is identical, then maybe not worth having two methods

optimum/neuron/utils/cache_utils.py
dacorvo: Also, why does it exist at all ?

optimum/commands/neuron/cache.py
```diff
@@ -12,87 +12,114 @@
...
             type=str,
-            help="The model_id to lookup cached versions for.",
+            help="The model_id to pre-warm cache for.",
         )
         parser.add_argument(
-            "--task",
+            "--cache_dir",
             type=str,
             default=None,
-            help="The optional task to lookup cached versions for models supporting multiple tasks.",
```
dacorvo, line 95: you need task

optimum/neuron/cache/optimum_neuron_cc_wrapper.py
```diff
@@ -15,12 +15,12 @@
...
 from libneuronxla.neuron_cc_wrapper import main as neuron_cc_wrapper_main
 
-from ..utils.cache_utils import get_hf_hub_cache_repos, get_neuron_cache_path
+from ..utils.cache_utils import get_neuron_cache_path
 from .hub_cache import hub_neuronx_cache
 
 
 def main():
-    with hub_neuronx_cache(cache_repo_id=get_hf_hub_cache_repos()[0], cache_dir=get_neuron_cache_path()):
+    with hub_neuronx_cache(cache_dir=get_neuron_cache_path()):
```
dacorvo, line 23: Need to double-check if this wrapper is still used. Otherwise it sjhould be removed as a refactor commit first before even starting the migration to the bucket cache

optimum/neuron/cache/traced.py
dacorvo: I don't think it is used anymore

optimum/exporters/neuron/convert.py
```diff
@@ -27,17 +27,13 @@
 import torch
```
dacorvo, line 27: this should be squashed with the modification of the callers

optimum/exporters/neuron/convert.py
```diff
@@ -344,6 +340,27 @@ def export_models(
             f"Provided {len(output_file_names)} custom names for the export of {len(models_and_neuron_configs)} models. Please provide the same number of names as models to export."
         )
 
+    # Resolve model_id and task for bucket cache context
```
dacorvo: Why can't you use the context here ? This duplicates the logic

optimum/commands/neuron/cache.py
dacorvo: you still need task

optimum/commands/neuron/cache.py
```diff
@@ -121,13 +134,10 @@ def parse_args(parser: "ArgumentParser"):
             type=int,
             help="Only look for cached models supporting at least the specified sequence length.",
         )
-        parser.add_argument("--repo_id", type=str, default=None, help="The name of the repo to use as remote cache.")
 
     def _list_entries(self):
         entries = select_hub_cached_entries(
             self.args.model_id,
-            task=self.args.task,
```
dacorvo, line 129: you still need task here

optimum/commands/neuron/cache.py
```diff
@@ -138,24 +148,21 @@ def _list_entries(self):
...
+                    str(entry.get("tp_degree", entry.get("tensor_parallel_size", "?"))),
+                    str(entry.get("torch_dtype", entry.get("dtype", "?"))),
+                    str(entry.get("target", "?")),
                 )
             )
-        # Remove duplicates (might happen if the same arch was compiled several times with different models and sync'ed afterwards)
         rows = list(set(rows))
-        # Sort by tensor parallel size, then batch size, sequence length, dtype
-        rows = sorted(rows, key=lambda x: (int(x[2]), int(x[0]), int(x[1]), x[3]))
+        rows = sorted(rows, key=lambda x: (x[2], x[0], x[1], x[3]))
```
dacorvo: I am worried this won't sort properly numbers

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,391 @@
...
+
+    cmd = [uv_bin, "run"]
+    for dep in _UV_DEPS:
+        cmd.extend(["--with", dep])
+    cmd.extend(["python", _BUCKET_SERVER_PATH, _server_socket_path])
+
+    _server_process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    try:
+        pid_line = _server_process.stdout.readline().decode().strip()
```
dacorvo: this is overly complicated. Why domn't you just get the pid from the subprocess ?

optimum/neuron/cache/bucket_server.py
```diff
@@ -0,0 +1,191 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""HF Storage Bucket proxy server.
+
+A generic proxy mirroring the huggingface_hub bucket API over a Unix socket.
+Runs in an isolated environment (via uv) with huggingface_hub >= 1.0.
+Contains no neuron-specific logic — all cache semantics live in bucket_cache.py.
+
+Protocol:
+    Request:  4-byte big-endian length prefix + UTF-8 JSON payload
```
dacorvo, line 23: So each request response is limited to 2^32 - 1 char, which is huge, but still, we might want to err if the payload is larger

optimum/exporters/neuron/convert.py
```diff
@@ -344,92 +340,102 @@ def export_models(
...
-    failed_models = []
-    total_compilation_time = 0
-    compile_configs = {}
-    for i, model_name in enumerate(models_and_neuron_configs.keys()):
-        logger.info(f"***** Compiling {model_name} *****")
-        submodel, sub_neuron_config = models_and_neuron_configs[model_name]
-        output_file_name = (
-            output_file_names[model_name] if output_file_names is not None else Path(model_name + ".neuron")
-        )
+    # Use bucket cache context: fetch on enter, sync on exit
```
dacorvo: Still not right: why not wrapping the loop with a with statement ?

optimum/exporters/neuron/convert.py
```diff
@@ -344,92 +340,102 @@ def export_models(
...
+                model_config = DiffusersPretrainedConfig.from_dict(model_config)
+
+            # only register mandatory input shapes
+            input_shapes = sub_neuron_config.input_shapes
+            mandatory_shape = [
+                elem for arg in sub_neuron_config.INPUT_ARGS for elem in ((arg,) if isinstance(arg, str) else arg[1:])
+            ]
+            input_shapes = {k: v for k, v in input_shapes.items() if k in mandatory_shape}
+
+            model_config = store_compilation_config(
```
dacorvo: what's this for ?

optimum/exporters/neuron/convert.py
```diff
@@ -344,92 +340,102 @@ def export_models(
...
-                "The compilation of SDXL's unet with the weights/neff separation is broken since the Neuron SDK 2.18 release. `inline_weights_to_neff` will be set to True and the caching will be disabled. If you still want to separate the neff and weights, please downgrade your Neuron setup to the 2.17.1 release."
+            # TODO: Remove after the weights/neff separation compilation of sdxl is patched by a neuron sdk release: https://github.com/aws-neuron/aws-neuron-sdk/issues/859
+            if not inline_weights_to_neff and getattr(sub_neuron_config, "is_sdxl", False):
+                logger.warning(
+                    "The compilation of SDXL's unet with the weights/neff separation is broken since the Neuron SDK 2.18 release. `inline_weights_to_neff` will be set to True and the caching will be disabled. If you still want to separate the neff and weights, please downgrade your Neuron setup to the 2.17.1 release."
+                )
+                inline_weights_to_neff = True
+
+            start_time = time.time()
+            neuron_inputs, neuron_outputs = export(
```
dacorvo: I am pretty sure that the context should be INSIDE export

optimum/neuron/cache/bucket_cache.py
```diff
@@ -0,0 +1,407 @@
...
+                break
+            except OSError:
+                pass
+        time.sleep(0.1)
+    else:
+        _server_process.kill()
+        _server_process = None
+        raise RuntimeError("Bucket server did not become ready within 5 seconds")
+
+    atexit.register(_stop_server)
```
dacorvo, line 110: what if this process crashes ?

optimum/neuron/trainers/transformers.py
```diff
@@ -733,12 +729,8 @@ def num_tokens(train_dl: DataLoader, max_steps: int | None = None) -> int:
...
 
     def synchronize_hub_cache(self):
-        repo_id = get_hf_hub_cache_repos()[0]
-        if not self.args.skip_cache_push:
-            has_write_access = has_write_access_to_repo(repo_id)
-            if has_write_access:
-                cache_path = get_neuron_cache_path()
-                synchronize_hub_cache(cache_path=cache_path, cache_repo_id=repo_id)
+        # Sync happens automatically in hub_neuronx_cache context on exit.
+        synchronize_hub_cache()
```
dacorvo: ??????

optimum/neuron/utils/runner.py
```diff
@@ -346,22 +346,9 @@ def check_user_logged_in_and_cache_repo_is_set(self):
...
-                "https://huggingface.co/docs/optimum-neuron/guides/cache_system#how-to-use-a-private-trainium-model-cache"
-            )
-
-        main_repo = get_hf_hub_cache_repos()[0]
-        has_write_access = has_write_access_to_repo(main_repo)
-        if not has_write_access:
-            raise RuntimeError(
-                f"You do not have write access to {main_repo}. Please log in and/or use a custom Neuron cache repo."
-            )
+        bucket_id = get_cache_bucket()
```
dacorvo, line 349: hmmmm we are losing the warning about not being authorized to push to the public bucket

PR #1113, opened 2026-08-17
chore(sdk): upgrade to Neuron SDK 2.31

tengomucho:
Draft PR to move optimum-neuron to Neuron SDK 2.31.0 and vLLM 0.16.0.
Note that part of the work in this PR has been created with the help of AI, that was used also to keep the CI working.

## Changes

- **Dependencies**: bump to Neuron SDK 2.31.0 (`7bb4756f`) and align CI Neuron runtime pins (`d2dbef4d`); bump package version to 0.4.7.dev0 with `sdk_version = 2.31.0` (`744e2e3f`); move vLLM to 0.16.0 (`f0105ecb`).
- **vLLM**: port the plugin to the vLLM 0.16.0 module layout (`70066cb7`); force spawn for EngineCore to avoid an NRT fork deadlock (`3e0bac3f`) and make the ModelConfig parallel-config patch picklable for spawn (`a91b7982`).
- **gemma3**: keep flash attention softmax in fp32 for head_dim > 128 (`ff1fb2ff`).
- **Generation**: break greedy ties with argmax over the full logits (`146d3721`).
- **Tests**: isolate cache-test exports to fix Hub cache-hit checks (`fdee2751`); add known-divergence escape for the VLM cross-chunk case (`94d11a9e`).

## Known issue (needs investigation)

The SDK 2.31 tracer segfaults/aborts on several conv-based encoder models (convbert, hubert, wav2vec2, yolos). This is an upstream compiler regression, not branch code — affected exporter tests are skipped for now in `26508ed5`.
This might be related to and fixed by #1119, but I think it is better to work on that after this branch has been merged.


docker/vllm/Dockerfile
```diff
@@ -11,23 +11,37 @@ RUN apt-get update -y \
...
     libexpat1 \
     libpython3-dev \
+    libarchive13 \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
 # Install uv at a specific version on a given path
 RUN curl -LsSf https://astral.sh/uv/0.9.27/install.sh | XDG_BIN_HOME=/usr/local/bin sh
 
+# optimum-neuron requires Python >= 3.11 (SDK 2.31's compiler ships no cp310 wheels),
```
JingyaHuang: Why don't we upgrade to ubuntu24?

tengomucho: you are right, I forgot!

JingyaHuang (approved):
LGTM!

tests/decoder/test_cache.py
```diff
@@ -88,13 +89,38 @@ def get_local_cached_files(cache_path, extension="*"):
...
+    """Runs export_decoder_model + check_decoder_generation in a fresh subprocess.
+
+    torch_neuronx keeps a process-global HLO instance counter
+    (torch_neuronx.experimental.profiler.v2_x.custom_op_name.class_count) that gets baked
+    into the op_name metadata of the HLO, which is itself hashed to compute the Hub cache
+    lookup key. Exporting the same model twice within one process bumps that counter, so
+    the second export's hash never matches the first and the Hub cache lookup always
+    misses. Running each export in its own subprocess keeps the counter fresh so identical
+    models produce identical hashes, matching the deterministic caching behavior this test
+    is meant to verify.
```
dacorvo: This is not only an issue for tests, but also for the cache

tests/decoder/test_decoder_generation.py
```diff
@@ -276,6 +292,7 @@ def test_decoder_generation_long_sequence(neuron_llm_config: dict[str, Any]):
     [17, 30],
     ids=["shorter", "short"],
 )
+@subprocess_test
```
dacorvo, line 283: Hard-coded word size is perhaps not new ? If new, then this is also an issue.

docker/vllm/Dockerfile
dacorvo: This is a workaround. Since we will need to build the AMI correctly anyway, why not installing the correct python, or bump ubuntu ?

dacorvo (commented):
Looks good to me, except for the non-deterministic hash: it is very possible that this makes pre-caching invalid.

tests/decoder/test_cache.py
tengomucho: It can be an issue for the cache is the model is searched within the same process. I think we can probably live with this since most of the time we only use one model at the time. 
I had explored a monky-patching alternative, but in the end it seemed more fragile than the subprocess one. I could raise the issue with the neuron team though.

tengomucho: I force-pushed a better alternative to this change.

tests/decoder/test_decoder_generation.py
tengomucho: Not new...

dacorvo (approved):
Thanks for this, and congrats for finding the cache issue root cause.

PR #1119, opened 2026-08-21
Fix tracing crashes for yolos, wav2vec2, hubert and convbert (aws-neuron-sdk#1265)

jimburtoft:
Fixes the tracing crashes tracked in https://github.com/aws-neuron/aws-neuron-sdk/issues/1265, which are what currently pin this project below Neuron SDK 2.27.

## What was happening

`yolos`, `wav2vec2`, `hubert` and `convbert` don't raise during export — they **kill the Python process**, so there's nothing to catch.

All four are the same bug, and it isn't the individual ops. An `aten` op with no XLA lowering falls back to CPU through `at::native::cpu_fallback`, which calls `XLANativeFunctions::_to_cpu()` on the operands. That transfer only works if the operand has already been materialized into a PjRt buffer:

- operand derived from a **model input** → a buffer exists → fine
- operand derived from a **parameter or buffer** → torch-xla still holds an un-materialized placeholder whose buffer is `nullptr` → `Check failed: pjrt_data->buffer != nullptr` → `SIGABRT` or `SIGSEGV`

So a model only breaks when it applies an unlowered op to a **weight** rather than to an activation:

| Model | Unlowered op | Operand |
|---|---|---|
| `wav2vec2` | `aten::_weight_norm_interface` | `weight_norm` parametrization on `pos_conv_embed.conv` |
| `hubert` | `aten::_weight_norm_interface` | same |
| `yolos` | `aten::upsample_bicubic2d` | `position_embeddings` |
| `convbert` | `aten::im2col` | conv-weight-derived tensor |

The check: the same ops at the same shapes trace fine when fed a model input, and crash when fed a parameter or a registered buffer. Only provenance changes.

Two things worth correcting from the issue as filed:

- **The two reported failure classes are one bug.** The "segfault" and the `RuntimeError: PjRt buffer is null` are the same null-buffer dereference surfacing differently depending on timing. `wav2vec2` was filed under segfault but actually aborts.
- **`cvt` no longer reproduces on 2.31.** It passes untouched, so it needs no patch and could be un-skipped in a separate PR.

## What this PR does

Adds `optimum/exporters/neuron/neuron_trace_patches.py`, applied from `export_neuronx`, which rewrites each call site using ops that *do* have XLA lowerings so no CPU fallback is attempted:

- **`weight_norm`** — fold the parametrization so `w = g * v / ||v||` is evaluated once eagerly instead of on every forward. Applied to **every** model rather than just these two: any architecture using weight normalization (speech encoders, vocoders, some GANs) hits this identically, and folding is a no-op otherwise.
- **`yolos`** — the bicubic interpolation depends only on a parameter and on `img_size`, which is fixed for a traced model, so it's constant with respect to the graph. Snapshot the embeddings before tracing, then precompute and cache. The snapshot matters: reading the live tensor mid-trace synchronizes an XLA tensor whose computation is still in flight and fails with `Check failed: handle->HasValue()`.
- **`convbert`** — reimplement its narrow `unfold` case with pad/slice/stack/reshape. It raises on argument combinations it doesn't cover rather than silently computing something else.

All are monkey-patches applied at runtime, so no vendored `transformers` changes.

**Second commit removes the skip.** #1113 added `SDK_231_TRACE_CRASH_MODEL_TYPES` / `skip_if_sdk_231_trace_crash` to work around this, which silently drops export *and* inference coverage for all four models. Since the crash is fixed, this PR deletes the helper and its three call sites (`tests/conftest.py`, `tests/exporters/test_transformers.py`, `tests/inference/inference_utils.py`). Landing the fix without this would leave the tests that prove it disabled, so the two belong together.

## Verification

`tests/exporters/test_transformers.py` on SDK 2.31 / inf2.8xlarge, **one pytest process per model** so a crash is attributable rather than taking down the run:

| | before | after |
|---|---|---|
| `yolos` | `rc=139` (SIGSEGV) | pass |
| `wav2vec2` | `rc=134` (SIGABRT) | pass |
| `convbert` | `rc=139` (SIGSEGV) | pass |
| `hubert` | `rc=134` (SIGABRT) | pass |
| `cvt` | pass | pass |

Together: **18 passed, 36 skipped** (previously 2 failed / 16 passed, and the run aborted partway when invoked as one process).

Re-validated after rebasing onto `main` with #1113 merged, i.e. against the real SDK 2.31 pins (`neuronx-cc==2.26.6360.0`, `torch-neuronx==2.9.0.2.15.32035`), with the skip removed so these actually execute rather than skipping:

| Suite | Result |
|---|---|
| `tests/exporters/test_transformers.py -k "<the 5 models>"` | **18 passed**, 36 skipped |
| `tests/inference/transformers/test_modeling.py -k "<the 5 models>"` | **11 passed** |
| `tests/exporters/test_neuron_trace_patches.py` (new) | **13 passed** |

The inference tests are coverage that was **not** previously exercised — they were skipped by the same helper, so removing it re-enables the `NeuronModelForCTC`, `AudioClassification`, `AudioFrameClassification` and `XVector` paths for wav2vec2 in addition to the export path.

Note for whoever runs these in CI: the four audio inference tests need **FFmpeg** present, otherwise `torchcodec` cannot load `librispeech_asr_demo` and they fail with `Could not load libtorchcodec` before reaching any Neuron code. `apt-get install ffmpeg` fixes it; unrelated to this change but easy to mistake for a regression.

Numerical equivalence, against CPU, on the tiny-random fixtures *and* the real checkpoints:

| Checkpoint | before | after | cos_sim | patched-vs-unpatched on CPU |
|---|---|---|---|---|
| `facebook/wav2vec2-base-960h` | SIGABRT | pass | 1.0 | **0.0** |
| `facebook/hubert-base-ls960` | SIGABRT | pass | 1.000005 | **0.0** |
| `YituTech/conv-bert-base` | SIGSEGV | pass | 1.0 | **0.0** |
| `hustvl/yolos-tiny` | SIGSEGV | pass | 1.0 / 1.0 / 1.000004 | **0.0** |

The last column is the point: the patches are numerically inert. Residual `cos_sim` deviation is ordinary FP32 Neuron-vs-CPU noise. Also verified with `dynamic_batch_size=True`, since `test_compare_to_transformers_dyn_bs` was one of the failing tests.

New tests in `tests/exporters/test_neuron_trace_patches.py` (13 cases, CPU-only, no Neuron needed, ~13s): `unfold_via_slices` exactness vs `nn.functional.unfold`, one case per clause of its argument guard, weight-norm folding equivalence, and output equivalence for all four models.

## Still worth fixing upstream

These are workarounds; the real issue is in torch-xla. `_to_cpu()` could materialize the placeholder instead of asserting on it — parameters are known constants at trace time. Failing that, replacing the `CHECK` with a thrown exception would at least make it catchable from Python instead of killing the process. I've noted this on the Neuron issue.

## Environment

SDK 2.31 (DLAMI 20260708), inf2.8xlarge, torch-neuronx `2.9.0.2.15.32035`, neuronx-cc `2.26.6360.0`, torch `2.9.1`, transformers `4.57.6`, Python `3.12.3`.

Rebased onto `main` after #1113 merged, so this runs against the SDK 2.31 pins now on `main`. `pyproject.toml` is deliberately untouched — #1113 already relaxed `requires-python` to `>=3.11,<3.13`, which is what the DLAMI's Python 3.12.3 needs.

### DIFF
--- a/optimum/exporters/neuron/convert.py
+++ b/optimum/exporters/neuron/convert.py
@@ -45,6 +45,7 @@ from ...utils import (
     is_sentence_transformers_available,
     logging,
 )
+from .neuron_trace_patches import patch_model_for_neuron_tracing
 
 
 if TYPE_CHECKING:
@@ -536,6 +537,12 @@ def export_neuronx(
 
     dummy_inputs = prepare_dummy_inputs(config, input_shapes, return_dict=True)
     dummy_inputs_tuple = tuple(dummy_inputs.values())
+
+    # Some architectures apply an op that has no XLA lowering to a weight rather than
+    # to an activation, which kills the tracing process outright (SIGABRT / SIGSEGV).
+    # See `neuron_trace_patches` and https://github.com/aws-neuron/aws-neuron-sdk/issues/1265.
+    if isinstance(model_or_path, PreTrainedModel):
+        patch_model_for_neuron_tracing(model_or_path)
 
     # Prepare the model / function(tp) to trace
     if getattr(config, "is_encoder_decoder", False):
--- a/optimum/exporters/neuron/neuron_trace_patches.py
+++ b/optimum/exporters/neuron/neuron_trace_patches.py
@@ -1,0 +1,294 @@
+# coding=utf-8
+# Copyright 2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Compatibility patches that make certain models traceable by `torch_neuronx.trace`.
+
+Background
+----------
+`torch_neuronx.trace` lowers a model by running it once with tensors on an XLA
+device. An ``aten`` op that has no XLA lowering falls back to CPU through
+``at::native::cpu_fallback``, which calls ``XLANativeFunctions::_to_cpu()`` on the
+op's operands.
+
+That transfer only succeeds if the operand has already been materialized into a
+real PjRt buffer:
+
+* operand derived from a **model input** -> a buffer exists -> the fallback works.
+* operand derived from a **parameter or buffer** -> torch-xla still holds an
+  un-materialized placeholder whose buffer is ``nullptr``. The transfer then trips
+  ``Check failed: pjrt_data->buffer != nullptr`` and the process dies, either with
+  ``SIGABRT`` (the ``LOG(FATAL)`` path) or with a bare ``SIGSEGV``. Because the
+  crash happens in C++, it cannot be caught from Python.
+
+So a model breaks only when it applies an unlowered op to a *weight* rather than
+to an activation. This affects, on Neuron SDK >= 2.27:
+
+======================  ==============================  ===========================
+Model                   Unlowered op                    Operand
+======================  ==============================  ===========================
+``wav2vec2``            ``aten::_weight_norm_interface``  ``weight_norm`` parametrization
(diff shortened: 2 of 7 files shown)


tengomucho (approved):
The issue is present in SDK2.31 (`neuronx-cc` version `2.26.6360.0+6f180f47`). So this workaround makes sense, but it would be better to get a fix on the compiler itself.

tengomucho:
Closing this and reopening as #1124

